PR #154, opened 2024-11-04
fix: Include the generator for mount run commands

miguelmig:
Already implicity used by https://github.com/jdx/usage/blob/main/cli/src/cli/generate/fig.rs#L343


miguelmig:
Still not working perfectly need to dig a bit deeper, don't merge yet

jdx:
I'm trying to fix your PR because the job is failing by merging, you can feel free to overwrite. I can only test this using a fork.

miguelmig:
It seems to work now locally.
If, out of nowhere,  it breaks tomorrow I'll investigate if it's something related to caching or the likes.

PR #197, opened 2024-12-14
feat: feature for automatically adding code fences

jdx:


lib/src/docs/markdown/cmd.rs
```diff
@@ -62,6 +66,21 @@ mod tests {
...
+        ```
+        $ echo hello world
+        hello world
+
+        more code
+        ```
+
+        some docs
+
+        ```
```
ccoVeille, line 80: This one is alone, it looks like an issue, no?

lib/src/docs/markdown/cmd.rs
```diff
@@ -62,6 +66,21 @@ mod tests {
...
         flag2 long description
 
+        includes a code block:
+
+        ```
+        $ echo hello world
+        hello world
+
+        more code
+        ```
```
ccoVeille, line 76: Be careful about a pre block containing a fenced block

````
<pre>

foo
```
42
```
bar
</pre>
````

You would have to convert it to

`````
````
foo
```
42
```
bar
````
`````

PR #208, opened 2024-12-21
Better fig generation to avoid linter from complaining

miguelmig:
Generate Fig Spec in a better way to prevent fig's autocomplete linter from complaining

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,0 +1,5 @@
+module.exports = {
+  extends: [
+    "@fig/autocomplete"
+  ],
+};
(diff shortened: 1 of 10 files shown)


miguelmig:
If you(@jdx) ever want to push this to fig's main autocomplete repo, we just need to change the mise task `render:fig` to output as `usage.ts` and copy the action from the mise repo that auto publishes it

jeremiah-snee-openx:
@miguelmig @jdx Trying to follow up on current status? 
- https://github.com/jdx/mise/discussions/4852

PR #216, opened 2025-01-08
feat: added file_path property

jdx:


jdx:
I found this was possible with the source code template

PR #296, opened 2025-07-11
fix: fall back to listing files on unknown completions

jdx:
makes usage behave like shells do when they don't know what to complete


jdx:
bugbot run

jdx:
bugbot run

jdx:
bugbot run

PR #304, opened 2025-07-16
fix(completions): ignore aliases and functions named usage (2nd attempt)

risu729:
Reverts jdx/usage#301. (Reapplies #300.)

Replaces `command -v usage` with `type -p usage` to ignore aliases and functions.
In the 1st attempt, I used `command usage`, but since `usage` (or `usage --help`) exits with an exit code `2`, it didn't work as an existence check.

I apologise for the trouble. Thank you for your help as always.

---

Resolves https://github.com/jdx/mise/discussions/5358.

This PR replaces `usage` with `command usage`, to ignore aliases and functions named `usage`.

Also, for consistency, use `command usage` instead of `command -v usage` to check its existence.
In this way, if aliases or functions named `usage` exist but `usage` is not installed, the completions are skipped as expected.

I don't think this is a case, but something like `alias usage=usage --default-flag` will be ignored by this PR.


lib/src/complete/snapshots/usage__complete__zsh__tests__complete_zsh-2.snap
```diff
@@ -20,7 +19,7 @@ _mycli() {
   typeset -A opt_args
   local curcontext="$curcontext" spec cache_policy
 
-  if ! command -v usage &> /dev/null; then
+  if ! type -P usage &> /dev/null; then
```
jdx, line 22: at least for me I get this in zsh:

```
jdx@air ~ % type -P usage >/dev/null
type: bad option: -P
```

risu729:
I confirmed `type -p` works fine for both bash and zsh.

`bash`
```
$ type -p usage
/home/risu/.local/share/mise/installs/usage/2.2.2/usage
$ type -p beep
```

`zsh`
```
% type -p usage
usage is /home/risu/.local/share/mise/installs/usage/2.2.2/usage
% type -p beep
beep not found
% which beep
beep: aliased to printf '\a'
```

PR #322, opened 2025-09-10
more flexible parsing of the #USAGE lines

NorthIsUp:
allows for whitespace and case insensitivity


jdx:
I specifically want to avoid this so we don't end up with false positives

PR #548, opened 2026-03-18
Support env-backed choices with `choices env=...`

mustafa0x:
## Summary
- add `env` support to `choices` so specs can source allowed values from an environment variable
- honor `Parser::with_env(...)` without falling back to the process environment when a custom env map is provided
- reject explicit values when `choices env=...` resolves to no values instead of silently disabling validation
- preserve `choices env=...` when `SpecArgBuilder::choices()` is called after `choices_env(...)`
- resolve env-backed choices for interactive shell completion at completion time
- preserve `choices env=...` in help/docs instead of snapshotting live env-backed values into generated output
- keep Fig suggestions literal-only so `usage generate fig` does not bake generation-time environment values into a static file

## Example
```kdl
arg "<env>" {
  choices env="DEPLOY_ENVS"
}
```

With `DEPLOY_ENVS="foo,bar baz"`, valid values become `foo`, `bar`, and `baz`.

## Implementation notes
- `choices` now accepts an optional `env` property and preserves it when serializing back to KDL
- env-backed values are split on commas and/or whitespace, with empty entries ignored
- env-backed values are appended after literal choices and deduplicated against existing entries
- parsing uses the caller-provided env map when `Parser::with_env(...)` is used; it only consults the process environment when no custom env map is provided
- env-backed choices that resolve to no values now fail validation for explicit CLI input with a targeted error message instead of accepting any value
- `SpecArgBuilder::choices()` now preserves any previously configured `choices_env(...)`
- help/docs render literal choices as before and also surface the controlling env var for env-backed choices

## Tests
- unit tests for env choice splitting and deduplication
- parser tests covering arg/flag validation with `Parser::with_env(...)`
- parser test covering the empty env-backed choice set case
- builder test covering `.choices_env(...).choices(...)` ordering
- completion test covering env-backed choices
- dump test covering KDL round-tripping

### DIFF
--- a/lib/src/parse.rs
+++ b/lib/src/parse.rs
@@ -740,11 +740,19 @@ fn validate_choices(
     choices: Option<&SpecChoices>,
     custom_env: Option<&HashMap<String, String>>,
 ) -> miette::Result<bool> {
+    if is_help_arg(spec, value)
+        && choices.is_some_and(|choices| {
+            !choices
+                .values_with_env(custom_env)
+                .iter()
+                .any(|choice| choice == value)
+        })
+    {
+        errors.push(render_help_err(spec, cmd, value.len() > 2));
+        return Ok(true);
+    }
+
     if let Some(err) = choice_error(target, value, choices, custom_env) {
-        if is_help_arg(spec, value) {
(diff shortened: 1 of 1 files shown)


jdx:
i think you need to rebase

PR #570, opened 2026-03-31
feat: enable wasm32 compilation by making `xx` crate conditional

andreaTP:
Supersedes #561 this PR contains only the Rust-level changes needed for wasm32 support to avoid having to patch on top downstream.

The `xx` crate doesn't support wasm32. This makes `xx` a conditional dependency (`cfg(not(target_arch = "wasm32"))`) and replaces its usages (`xx::regex!`, `xx::file`, `xx::process`) with std equivalents. `sh` execution is disabled on wasm (returns an error).

CI now includes a minimal `cargo build --target wasm32-wasip1` check to prevent regressions and this is all I need to bundle and ship a `usage4j` library usable and embeddable in Java.

Please @jdx have a look and let me know your thoughts :pray:


andreaTP:
Rebased on latest main.
I understand is not a priority, but I'd like to hear your thoughts on this @jdx .
Is very little changes to enable compatibility

PR #573, opened 2026-04-05
feat(parse): add usage__varargs_idx env var

JanPokorny:
Add special env var for usage scripts containing number of parsed args before trailing varargs.

Intended use -- pass on varargs without needing to use `eval`. Despite being safe, it raises eyebrows and this change is IMO simple and unobtrusive. It can exist alongside, being just another option.

Example of a script that has args and flags followed by varargs:
```bash
shift $usage__varargs_idx  # removes parsed args from $@
eslint "$@"
```

It's not the most elegant solution in the world but it might be less inelegant than `eval`.

Code is by Claude with significant handholding by me.

### DIFF
--- a/cli/src/cli/exec.rs
+++ b/cli/src/cli/exec.rs
@@ -78,6 +78,12 @@ impl Exec {
         for (key, val) in &parsed.as_env() {
             cmd.env(key, val);
         }
+        if let Some(trailing) = parsed.trailing_varargs_count() {
+            cmd.env(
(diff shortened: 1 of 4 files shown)


JanPokorny:
I'd like to hear your thoughts here, @jdx 
(pinging in case this got lost, I get it if you are busy)

JanPokorny:
Closing due to a lack of interest.

PR #641, opened 2026-05-19
feat: enable wasm32 compilation by making `xx` crate conditional

andreaTP:
Supersedes #570 and #561 — this PR contains only the Rust-level changes needed for wasm32 support to avoid having to patch on top downstream.

The `xx` crate doesn't support wasm32. This makes `xx` a conditional dependency (`cfg(not(target_arch = "wasm32"))`) and replaces its usages (`xx::regex!`, `xx::file`, `xx::process`) with std equivalents. `sh` execution is disabled on wasm (returns an error).

CI now includes a minimal `cargo build --target wasm32-wasip1` check to prevent regressions and this is all I need to bundle and ship a `usage4j` library usable and embeddable in Java.

Please @jdx have a look and let me know your thoughts :pray:


jdx:
sorry I think I'm going to pass on this

andreaTP:
This is very poor OSS maintainership: I understand you may not be interested in these changes, but leaving contributions hanging repeatedly and closing them without even minimal feedback is unnecessarily dismissive.

jdx:
I get hundreds of PRs per week and have no obligation to give any feedback or ever respond. Take your entitlement somewhere else.

I also think I was pretty clear this wasn't priority for me: https://github.com/jdx/usage/pull/561#issuecomment-4162839372

andreaTP:
Would you like to open a dialog?

Changes in this PR are straightforward and negligible in this codebase in my opinion. 
This little change fully enables consuming the project from any wasm runtime which seems a valueable outcome.

I fully appreciate that's not a priority, but the impact is very very limited and the potential is interesting from my pov.

jdx:
Your PR is slop. You deleted unrelated comments, it degrades logging functionality. It isn't fixing the issue in the right place. If I can expect more of that please do not bother submitting more PRs to my projects.

I went ahead and did this for you, this is the correct way to fix this issue: https://github.com/jdx/xx/pull/242. This does not necessarily mean that I'm in favor of more PRs for wasm support—I still do not think it is important.

andreaTP:
For the record, the `xx` fix doesn't address the issue.
`usage` still imports `xx::process` unconditionally and fails to compile for wasm, but I wish you well regardless.

PR #649, opened 2026-05-31
fix(parse): keep inherited global flags when a subcommand re-declares them as non-global

JamBalaya56562:
## Summary

Fixes a `parse_partial` bug where a **value-taking global flag placed before a (mounted) subcommand** breaks completion/parsing when the subcommand re-declares that same flag as non-global.

This fixes the parser-side root cause referenced by jdx/mise#10069.

### Reproduction

mise emits a usage spec where `run` mounts tasks dynamically, a task can have a `choices` positional arg, and `run` re-declares the global `-C/--cd` flag as non-global. Completing such a task with the global flag in front:

```
mycli -C /tmp run sample:run <TAB>
→ Error: × Invalid choice for arg profile: -C, expected one of alpha, beta, gamma
```

Expected: `alpha beta gamma`.

### Root cause

During phase 1 (subcommand scan / global-flag gathering), each descent did:

```rust
out.available_flags.retain(|_, f| f.global);
out.available_flags.extend(gather_flags(&subcommand));
```

When a child subcommand re-declares a parent's **global** flag (same `-C/--cd` key) as **non-global**, `extend` overwrites the inherited global entry with the non-global one, and the *next* descent's `retain(|_, f| f.global)` then drops it entirely.

With the global flag no longer in `available_flags`, phase 2 no longer recognizes the leftover `-C` token and feeds it to the task's `choices` positional arg → `validate_choices` bails.

### Fix

Replace `retain` + `extend` with a small `merge_subcommand_flags` helper that merges with **global precedence**: a subcommand's non-global re-declaration no longer shadows an inherited global flag. The global flag stays recognized, so phase 2 consumes it as a flag (and records it in `as_env()`, preserving `usage_*` env vars for normal execution and for mount scripts) instead of mistaking it for a positional.

> Note: the prefix global flag is intentionally left in `input` so phase 2 re-parses it into `out.flags`; that is how its value reaches `as_env()`. The fix is purely about keeping the flag *recognized* across descents.

## Tests

- **Unit** (`lib/src/parse.rs`): `test_prefix_global_flag_does_not_pollute_choices` builds the post-mount structure directly (root global `-C/--cd`, `run` re-declaring it as non-global, mounted `sample:run` with a `choices` arg) and asserts: no false "Invalid choice" bail, the inherited global flag survives descent, the flag value still reaches `as_env()` (`usage_cd`), valid choices still parse, and genuinely invalid choices are still rejected.
- **Integration** (`cli/tests/complete_word.rs` + `examples/mounted-global-flags-choices.sh`): a new fixture whose mounted `choices` vary by `usage_cd`, so the test also confirms the global flag value propagates through the mount despite the non-global re-declaration.

## Downstream note

mise currently has a temporary workaround (promoting the conflicting `run`/`tasks run` flags to `global=true` in its completion spec generation). Once this lands, that workaround becomes a no-op and can be removed; I'll follow up on jdx/mise#10069.

### DIFF
--- a/cli/tests/complete_word.rs
+++ b/cli/tests/complete_word.rs
@@ -162,6 +162,55 @@ fn complete_word_mounted_with_global_flags() {
 }
 
 #[test]
+fn complete_word_mounted_global_flag_choices() {
+    // Regression for the parser-side root cause referenced by jdx/mise#10069:
+    // a value-taking global flag placed before a mounted subcommand must not leak its
+    // consumed tokens into the mounted task's `choices` positional arg.
+    let mut path = env::split_paths(&env::var("PATH").unwrap()).collect::<Vec<_>>();
+    path.insert(
+        0,
+        env::current_dir()
+            .unwrap()
+            .join("..")
+            .join("target")
+            .join("debug"),
+    );
+    path.insert(0, env::current_dir().unwrap().join("..").join("examples"));
+    env::set_var("PATH", env::join_paths(path).unwrap());
+
+    // Baseline: no global flag prefix. The mount sees no `usage_cd`, so it returns the
+    // default choices. This must complete (not error) and not be polluted by stray tokens.
+    assert_cmd(
+        "mounted-global-flags-choices.sh",
+        &["--", "run", "sample:run", ""],
+    )
+    .stdout("one\ntwo\n");
+
+    // Value-taking global flag before the mounted subcommand (the failing case). The choices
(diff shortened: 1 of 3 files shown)


JamBalaya56562:
Thanks for the review! Applied the suggestion — since vailable.retain(|_, f| f.global) runs first and the merged 
ew_flags keys are unique, any present key is guaranteed to be an inherited global flag, so the check is now simply vailable.contains_key(&key) (added a comment noting the invariant). Pushed as a force-update to the branch.

PR #676, opened 2026-06-11
fix: allow for variadic arguments to capture unknown flags as well

rtpg:
When we have a variadic arg meant to be passed down to a wrapped command, generally we'll want unknown flags to also get passed down (for exmaple when the mise command is wrapping at test command)

```
#!/usr/bin/env -S usage bash
#MISE description="print args"
#USAGE arg "[other_args]" var=#true help="Arguments forwarded to wrapped command"

echo "ARGS WERE ${usage_other_args}"
```

^ for example, in this case, we would want to capture _any_ unknown flag along with an argument.

--------

Given the following script (in `some-script.sh`)

```
#!/usr/bin/env -S usage bash
#MISE description="print args"
#USAGE arg "[other_args]" var=#true help="Arguments forwarded to wrapped command"

echo "ARGS WERE ${usage_other_args}"
```

^ currently on `main` if I do:

```
cargo build && ./target/debug/usage bash some-script.sh foo --bar=baz
```

I get 
```
# snip builds
ARGS WERE foo '--bar=baz' baz
```

notice that the `--bar=baz` flag gets also passed along as an arg. This commit fixes this behavior.

Personally I think this is a common enough pattern in shell scripts to warrant its own spec entry... something like `#USAGE restargs "echo_args" help="Arguments to be passed to echo"`. Maybe `restargs` would "just" require `--` (so `usage bash some-script.sh -- foo --bar-baz`), or support `--`. And maybe my wanted behavior is weird because of this arg/flag distinction.

## Bug Fixes

* Fixed argument parsing to correctly handle flag arguments with embedded values (e.g., `--flag=value`) when used with variadic positional arguments. Previously, unknown flags with values could be incorrectly split into separate tokens.

### DIFF
--- a/lib/src/parse.rs
+++ b/lib/src/parse.rs
@@ -556,10 +556,13 @@ fn parse_partial_with_env(
         if enable_flags && w.starts_with("--") {
             grouped_flag = false;
             let (word, val) = w.split_once('=').unwrap_or_else(|| (&w, ""));
-            if !val.is_empty() {
-                input.push_front(val.to_string());
-            }
             if let Some(f) = out.available_flags.get(word) {
+                // Only push the embedded value back when the flag is known so that
+                // unknown --flag=value tokens fall through intact to positional arg
+                // handling without also injecting a stray "value" positional.
+                if !val.is_empty() {
+                    input.push_front(val.to_string());
+                }
                 if f.arg.is_some() {
                     out.flag_awaiting_value.push(Arc::clone(f));
                 } else if f.count {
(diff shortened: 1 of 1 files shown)


rtpg:
Rereading this change (disclaimer: I came up with the test cases, Claude slotted in the bugfix) and the `--` handling just above... feel like there might be more going on

on main right now though:

```
cargo build && ./target/debug/usage bash some-script.sh -- foo --bar=baz

ARGS WERE foo '--bar=baz' baz
```

```
cargo build && ./target/debug/usage bash some-script.sh -- -- foo --bar=baz

ARGS WERE foo '--bar=baz'
```

needed two `--`s to get things working with `usage bash`

PR #727, opened 2026-07-20
fix(completion): write spec cache to private dir instead of world-writable tmp

### LINKED ISSUE #722: Arbitrary file overwrite vulnerability through symlinks
In what appears to be originating from the usage project, for example how mise's bash completion is implemented, the mechanism enables overwriting arbitrary files that the invoking user has write access to using a symlink attack. This is due to use of predictable filenames in a (typically) world writable directory.

Specifically, for example with` mise 2026.7.7:
```shellsession
$ mise completion bash | sed -n 12,14p
    local spec_file="${TMPDIR:-/tmp}/usage__usage_spec_mise_2026_7_7.spec"
    if [[ ! -f "$spec_file" ]]; then
        mise usage >| "$spec_file"
```

`"${TMPDIR:-/tmp}/usage__usage_spec_mise_2026_7_7.spec"` could be a symlink pointing anywhere, planted by anyone/thing as this is very likely a world writable directory, resulting in the symlink target's contents replaced by the `mise usage` output. Exploitation does require quite accurate timing due to the `! -f` test, but that test does not eliminate the issue, it just makes it a TOCTOU one. `-f` returns true for symlinks pointing to regular files.

Additionally, the use of `>|` as opposed to just `>` means users can not protect against this using the `noclobber`  `set` option.

I understand the reason for this appears to be to try and avoid `mise usage` calls, but considering on a not-that-fast-system-by-todays-standards 
```shellsession
$ time mise usage >/dev/null

real	0m0,061s
user	0m0,029s
sys	0m0,037s
```
... I think this is not a good security tradeoff, especially considering this is likely to happen at most once per shell session. I have not researched it further, but perhaps the use of temporary files could be removed altogether, and `usage` made to read the spec from stdin instead, always invoking `mise usage`?

I have not checked the corresponding implementation for any other shells than bash.

I also did not find any particular instructions for filing security related issues, therefore reporting like this.

jdx:
## Summary

Fixes #722 — an arbitrary-file-overwrite vulnerability via symlink attack in the generated shell completion scripts.

The completion scripts cached the usage spec at a **predictable filename in a world-writable directory**:

```sh
local spec_file="${TMPDIR:-/tmp}/usage_..._.spec"
...
mise usage >| "$spec_file"
```

Because the path is guessable and `/tmp` is typically world-writable, another local user (or process) can pre-plant a symlink there. When the victim triggers completion, the spec write follows the symlink and clobbers whatever file it points at — any file the invoking user can write. As the reporter (@scop) notes, the `! -f` guard is only a TOCTOU check (`-f` returns true *through* a symlink), and `>|` / `save -f` deliberately override the user's `noclobber` setting, removing the one defense they had.

This affects all four generators that use a temp-file cache: **bash, zsh, fish, and nu**.

## Fix

**1. Relocate the cache to a private per-user directory (the security fix).** Moved from the shared world-writable temp dir to a per-user cache directory created with mode `700`:

- bash / zsh: `${XDG_CACHE_HOME:-$HOME/.cache}/usage`
- fish: `(XDG_CACHE_HOME or $HOME/.cache)/usage`
- nu: `($env.XDG_CACHE_HOME? | default ($nu.home-path | path join ".cache") | path join "usage")`

Once the directory is owned by and writable only by the invoking user, the symlink-planting precondition disappears — an attacker can no longer create entries in it — so the predictable filename no longer matters.

**2. Reap stale specs by age on regenerate (hygiene follow-up).** The cache filename is version-keyed. Unlike `/tmp` (cleared on reboot), `~/.cache` persists, so old versions would otherwise accumulate forever (a mise spec is ~200KB and mise releases frequently). On a **cache miss**, the script now removes this bin's spec files that haven't been regenerated in the last 30 days:

- bash / zsh / fish: `find "$spec_dir" -maxdepth 1 -name 'usage__usage_spec_<bin>_*.spec' -type f -mtime +30 -delete 2>/dev/null`
- nu: `glob ... | each {|f| if ((ls $f | get 0.modified) < ((date now) - 30day)) { rm --force $f } }`

This is **age-based rather than "delete all other versions"** deliberately. An earlier iteration deleted every other version of the bin on each regenerate, but that thrashes when two versions of the same tool run concurrently (e.g. version-pinned projects, or two shells): each cache miss would delete the other version's spec, so every completion becomes a cache miss + full regeneration, silently defeating the cache. Age-based keeps each live version's (recent) spec and only reclaims genuinely abandoned ones. The reap is scoped to the current bin (other tools untouched) and runs only on the cache-miss path, so the common cache-hit path is unaffected; a continuously-used old version regenerates at most ~once/month.

The `>|` / `save -f` clobber semantics are kept intentionally: the target is now a private cache file that mise owns and is meant to refresh, so honoring `noclobber` against our own cache file would just break completions for `noclobber` users. The security rationale for that complaint is resolved by the directory move.

## Testing

- Updated the unit snapshot tests per shell (`cargo insta`).
- Updated `cli/tests/shell_completions_integration.rs` to sandbox `XDG_CACHE_HOME` instead of `TMPDIR`.
- `cargo test -p usage-lib --lib` (253) and `cargo test -p usage-cli --test shell_completions_integration` (13) pass; `cargo fmt --check` clean.
- Verified end-to-end in real bash and zsh: the `700` cache dir is created; a cache miss regenerates the spec; a 40-day-old same-bin spec is reaped while a recent same-bin spec (a concurrently-live version) survives; another tool's old spec is untouched; and unmatched patterns don't error. fish/nu generators are covered by the unit snapshots (those shells aren't installed in the test env, so their integration tests skip).

---

> [!NOTE]
> **Medium Risk**
> Security-sensitive completion path changes affect all four shell generators and on-disk cache behavior, though scope is limited to where specs are written and cached.
> 
> **Overview**
> Fixes a **local symlink / arbitrary-file overwrite** risk by no longer writing completion spec caches under predictable paths in world-writable temp (`TMPDIR` / `/tmp`).
> 
> **Generated completion scripts** (bash, zsh, fish, nu) and the bundled `usage` completions now use a **per-user cache directory** (`$XDG_CACHE_HOME/usage` or `~/.cache/usage`, created with mode `700`) instead of `/tmp`. Version-keyed caches additionally **prune stale spec files** for the same binary (older than 30 days) only on a cache miss, so abandoned versions do not fill disk while concurrent tool versions keep working.
> 
> Integration tests sandbox **`XDG_CACHE_HOME`** instead of **`TMPDIR`**; unit snapshots reflect the new emitted shell code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b00453036ee1a6944de1ca136729dd7f9821162. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/src/complete/bash.rs
```diff
@@ -65,7 +77,9 @@ fi"#
         r#"
 	local cur prev words cword was_split comp_args
     _comp_initialize -n : -- "$@" || return
-    local spec_file="${{TMPDIR:-/tmp}}/usage_{spec_variable}.spec"
+    local spec_dir="${{XDG_CACHE_HOME:-$HOME/.cache}}/usage"
+    [[ -d "$spec_dir" ]] || mkdir -p -m 700 "$spec_dir"
```
scop, line 81: I think the -d test is pretty much a no-op here, and would be better off removed across all implementations. mkdir -p returns success if the dir exists already, and the -d test won't help with erroring out if the thing already exists as something else than a dir.

Also not sure about the need for 700 perms, but maybe it won't hurt, so leaving it in in this suggestion.

```suggestion
    mkdir -p -m 700 "$spec_dir"
```

scop: Oops, missed merge by a few seconds it seems :|

jdx: Good catch, and you're right on both counts — `mkdir -p` is idempotent so the `[[ -d ]]` guard never changes the common path, and it wouldn't help error out if the path already existed as a non-directory either (it'd just fall through to `mkdir -p` and fail there anyway). The only thing it saved was the odd `mkdir` fork, which is noise next to the `usage complete-word` fork every completion already does.

Since this merged just before your suggestion landed, I opened #730 to drop the guard across bash, zsh, and fish (nu already did an unconditional `mkdir`). Kept the `-m 700` — that's the actual fix for the #722 symlink issue, so it stays intentionally.

jdx: On reflection we're going to keep the `[[ -d ]]` guard intentionally, for perf. You're right that it's a no-op for correctness — `mkdir -p` is idempotent and the test doesn't guard against a non-dir path. But it's not there for correctness; it's there to skip a fork.

This line runs on every Tab (it's before the `[[ ! -f "$spec_file" ]]` cache-hit check), and `[[ -d ]]` is a bash builtin whereas `mkdir` is an external exec. So on the overwhelmingly common path — cache dir already exists — the guard avoids spawning a `mkdir` process on every completion. It's a small win, and yes the completion also forks `usage complete-word` regardless, but shaving an avoidable fork on the interactive hot path with a cheap builtin test felt worth keeping rather than removing.

So I'll leave it as-is across the four generators. Thanks for the careful read — closing the cleanup PR (#730) I'd opened.

PR #730, opened 2026-07-20
refactor(complete): drop redundant -d guard before mkdir in cache dir setup

jdx:
## Summary

Follow-up to #727 addressing @scop's [review comment](https://github.com/jdx/usage/pull/727#pullrequestreview-4738518469).

The completion generators guarded cache-dir creation with an existence test:

```sh
[[ -d "$spec_dir" ]] || mkdir -p -m 700 "$spec_dir"   # bash / zsh
test -d "$spec_dir"; or mkdir -p -m 700 "$spec_dir"   # fish
```

As @scop noted, that guard is effectively a no-op:

- `mkdir -p` already returns success when the directory exists, so the guard changes nothing on the common path.
- It doesn't add safety either — if `$spec_dir` exists as a *non-directory*, `-d` is false, so it falls through to `mkdir -p` (which then fails), and the later `>| "$spec_file"` fails regardless. The guard protects nothing.

The only thing the test saved was an occasional `mkdir` fork, but every completion already forks `usage complete-word`, so that's noise.

This removes the guard across **bash, zsh, and fish** so all four generators create the cache dir the same way (nu already did an unconditional `mkdir`). The `-m 700` mode is kept — that's the actual #722 fix.

## Testing

- `cargo insta test --accept -p usage-lib` — regenerated per-shell snapshots.
- `mise run render` — regenerated the bundled `usage` completion assets (`usage.bash`, `usage.fish`, `_usage`).
- `cargo test -p usage-lib --lib` (257) and `cargo test -p usage-cli --test shell_completions_integration` (13) pass; `cargo fmt --check` clean.

---

> [!NOTE]
> **Low Risk**
> Cosmetic shell-script simplification in completion cache setup; behavior is unchanged because `mkdir -p` is idempotent when the directory already exists.
> 
> **Overview**
> Completion scripts for **bash**, **zsh**, and **fish** no longer test whether the usage spec cache directory exists before creating it. They always run `mkdir -p -m 700` on `$XDG_CACHE_HOME/usage` (or `~/.cache/usage`), matching **nu** and dropping patterns like `[[ -d "$spec_dir" ]] || mkdir …` and fish’s `test -d …; or mkdir …`.
> 
> The change is in the Rust generators (`bash.rs`, `zsh.rs`, `fish.rs`), regenerated **usage** completion assets, and insta snapshots. **`mkdir -p -m 700` is unchanged** — only the redundant existence check is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae17a894c9ab3210be3c0473ccbfe2e2e895b1e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cli/assets/completions/_usage
+++ b/cli/assets/completions/_usage
@@ -25,7 +25,7 @@ _usage() {
   fi
 
   local spec_dir="${XDG_CACHE_HOME:-$HOME/.cache}/usage"
-  [[ -d "$spec_dir" ]] || mkdir -p -m 700 "$spec_dir"
+  mkdir -p -m 700 "$spec_dir"
   local spec_file="$spec_dir/usage__usage_spec_usage.spec"
   usage --usage-spec >| "$spec_file"
   local -a values=() descs=() inserts=()
--- a/cli/assets/completions/usage.bash
+++ b/cli/assets/completions/usage.bash
@@ -10,7 +10,7 @@ _usage() {
 	local cur prev words cword was_split comp_args
     _comp_initialize -n : -- "$@" || return
     local spec_dir="${XDG_CACHE_HOME:-$HOME/.cache}/usage"
-    [[ -d "$spec_dir" ]] || mkdir -p -m 700 "$spec_dir"
+    mkdir -p -m 700 "$spec_dir"
     local spec_file="$spec_dir/usage__usage_spec_usage.spec"
     usage --usage-spec >| "$spec_file"
     # shellcheck disable=SC2207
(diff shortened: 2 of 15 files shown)


jdx:
Closing — decided to keep the `[[ -d ]]` / `test -d` guard intentionally for the small perf win on the completion hot path (see the note on #727). The guard is a shell builtin that avoids a `mkdir` fork on every Tab.

scop:
Thanks!

Just one comment:
> The -m 700 mode is kept — that's the actual https://github.com/jdx/usage/issues/722 fix.

Now that the file is no longer in a presumably world or group writable dir, the only thing the mode prevents is others than the user being able to read files in it, which would have been likely prevented by the user's home dir permissions anyway. So as noted while this may not hurt per se (even though it does for example ignore the user's umask which might intentionally grant e.g. group read access, but getting pretty far fetched/strict/academic here), it's hardly the #722 fix -- moving the base dir to a better location is.

PR #731, opened 2026-07-20
fix(nu): create completion cache dir with mode 700 and fix home-dir lookup

jdx:
## Summary

Follow-up to #727, fixing two problems in the **nu** completion generator's cache-dir setup — both on the same lines that PR touched. Raised (in part) by Bugbot and Greptile on that PR.

### 1. Cache dir was world-readable (the reported gap)

Nushell's builtin `mkdir` has **no mode flag**, so the cache dir was created at the process umask — typically `755`, i.e. world-readable — whereas the bash/zsh/fish generators all use `mkdir -p -m 700`. The symlink-planting attack from #722 stays blocked (an attacker still can't write into `~/.cache`), but other local users could **read** the cached specs.

Fix: create the dir, then `^chmod 0700` it, guarded to non-Windows (where POSIX perms / `chmod` don't apply):

```nu
if not ($spec_dir | path exists) {
    mkdir $spec_dir
    if $nu.os-info.family != "windows" { ^chmod 0700 $spec_dir }
}
```

The `chmod` runs **only when the dir is first created** (behind the `path exists` check), so it stays off the completion hot path — matching the create-time-only semantics of `-m 700` in the other three shells rather than forking `chmod` on every Tab.

### 2. `$nu.home-path` no longer exists → completer errors on every invocation

While fixing the above I found the nu completion is **currently broken on any recent nushell**: `$nu.home-path` was renamed to `$nu.home-dir` in [nushell 0.110.0](https://www.nushell.sh/blog/2026-01-17-nushell_v0_110_0.html) (Jan 2026). Because `default` evaluates its argument **eagerly**, referencing the missing column throws — even when `XDG_CACHE_HOME` is set and the fallback isn't needed.

Fix: optional access with a fallback that works on both new and old nushell:

```nu
let spec_dir = ($env.XDG_CACHE_HOME? | default ($nu.home-dir? | default $nu.home-path? | path join ".cache") | path join "usage")
```

## Testing

Verified end-to-end on **nushell 0.114.1**:
- The generated completion script sources cleanly (no parse error).
- With `XDG_CACHE_HOME` set and via the `~/.cache` fallback, the cache dir is created as `drwx------` (700).
- The spec file is written into the 700 dir (so the dir mode is what enforces confidentiality).
- Second run: dir already exists → no `chmod`, no error.

Also: `cargo test -p usage-lib --lib` (257) passes, `cargo clippy` clean, `cargo fmt --check` clean, nu snapshots regenerated. (nushell isn't installed in CI, so its integration tests skip — hence the manual end-to-end check above.)

---

> [!NOTE]
> **Low Risk**
> Changes only the generated nu completion script strings and test snapshots; no runtime auth or data-path logic in the library itself.
> 
> **Overview**
> Aligns **nushell** completion cache setup with bash/zsh/fish and fixes breakage on recent Nushell.
> 
> Generated completers now resolve the cache path with `$nu.home-dir?` falling back to `$nu.home-path?` (optional chaining avoids errors when `XDG_CACHE_HOME` is set but the old `$nu.home-path` column is gone). They **only** `mkdir` when the usage cache dir is missing, then run `^chmod 0700` on non-Windows so new dirs are not world-readable—matching the other shells’ `mkdir -m 700` without chmod on every Tab.
> 
> Insta snapshots for the nu generator are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f54f4ad6529e1a7991b6d2da88de5732b0fdbb37. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
@coderabbitai review

PR #739, opened 2026-07-25
feat(spec): add effect= to declare what a command does to the world

jdx:
Lets a spec declare what running a command does:

```kdl
cmd "ls"        effect="read"        help="List installed tools"
cmd "use"       effect="write"       help="Install a tool and add it to the config"
cmd "uninstall" effect="destructive" help="Remove a tool"
```

| Effect | Meaning |
| --- | --- |
| `read` | Only inspects state. Running it twice is the same as running it once. |
| `write` | Creates or modifies state, but removes nothing the user cannot recreate. |
| `destructive` | May delete or irreversibly overwrite something. Deserves a confirmation prompt. |

## Why in the spec

Several consumers keep reinventing this same distinction, and each one hand-maintains its own list:

- **docs and `--help`** want to mark destructive commands
- **wrapper scripts** want to confirm before running one
- **AI coding agents** want an allowlist of read-only commands rather than prompting on every invocation — the [Agent Skills spec](https://agentskills.io/specification) has an `allowed-tools` field that is exactly this list, currently written by hand

That is the usual argument for putting something in a spec: one annotation, many renderings, same as `help` or `hide` today. It's structured data, so there's nothing to keep in sync and no prose to go stale.

## Design notes

**Not inherited by subcommands.** `git remote` and `git remote remove` do different things. Inheriting would make the least safe reading of a spec the default one, which is the wrong direction for a field whose whole purpose is flagging danger.

**Unset means unknown, not safe.** Consumers should treat a missing `effect` as "ask". This keeps the field additive — existing specs are unaffected, and a partially annotated spec degrades to today's behavior.

**Coarse on purpose.** Three values, not a capability model. Anything finer becomes a policy language, and that belongs in the consumer.

## Changes

- `SpecCommandEffect` enum in `lib/src/spec/effect.rs`, exported as `usage::SpecCommandEffect`
- `SpecCommand::effect: Option<SpecCommandEffect>`, plus a builder method
- parsed as a prop (`effect="read"`) or a child node (`effect "read"`) — the latter so framework integrations generating KDL don't have to emit props; unknown values are a parse error pointing at the span
- round-trips through KDL serialization
- generated markdown renders an `- **Effect**:` line next to Usage and Aliases

## Not in this PR

- `usage generate skill` consuming it to emit `allowed-tools`. That's the payoff, but it's a separate discussion about output format and shouldn't ride along with the spec primitive.
- manpage rendering.
- framework integrations. clap/cobra/etc. can't infer this — it needs an explicit annotation from the author, same as `hide`.

## Verified

`cargo test -p usage-lib` passes (275 + doctests). Four new tests cover prop parsing, child-node parsing, non-inheritance, KDL round-trip, and the error on an unknown value. Markdown output checked by hand against a sample spec.

`cargo test --workspace` has two pre-existing failures in `usage-cli --test examples` (`test_usage_double_slash_execution`, `..._old`) — confirmed present on a clean checkout of `main`, unrelated to this change.

---

> [!NOTE]
> **Low Risk**
> Additive spec field with optional semantics; no changes to parsing or runtime CLI behavior beyond new metadata and doc generation.
> 
> **Overview**
> Adds an optional **`effect`** on commands so specs can classify side effects as **read**, **write**, or **destructive** (via `effect="…"` or a child `effect` node). Missing `effect` stays **unknown**; values are **not inherited** by subcommands.
> 
> The library introduces **`SpecCommandEffect`**, wires it through **`SpecCommand`** (parse, KDL serialize, **`merge`**), **`SpecCommandBuilder::effect`**, and the public API. Generated markdown adds a **`- **Effect**:`** line when set. Docs in **`docs/spec/index.md`** describe semantics and consumers (docs, wrappers, agents).
> 
> Tests cover parsing, non-inheritance, KDL round-trip, merge behavior, invalid values, and markdown rendering. **`.gitignore`** ignores `*.pending-snap` from insta.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0e519290ba6a240170fb32598b9f2b9a2b892d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@
 #/target
 megalinter-reports/
 tasks/fig/build
+
+# insta writes these while running tests locally
+*.pending-snap
--- a/docs/spec/index.md
+++ b/docs/spec/index.md
@@ -88,11 +88,11 @@ cmd "use"       effect="write"       help="Install a tool and add it to the conf
 cmd "uninstall" effect="destructive" help="Remove a tool"
 ```
 
-| Effect         | Meaning                                                                              |
-| -------------- | ------------------------------------------------------------------------------------ |
-| `read`         | Only inspects state. Running it twice is the same as running it once.                |
-| `write`        | Creates or modifies state, but removes nothing the user cannot recreate.             |
-| `destructive`  | May delete or irreversibly overwrite something. Deserves a confirmation prompt.      |
+| Effect        | Meaning                                                                         |
+| ------------- | ------------------------------------------------------------------------------- |
+| `read`        | Only inspects state. Running it twice is the same as running it once.           |
+| `write`       | Creates or modifies state, but removes nothing the user cannot recreate.        |
+| `destructive` | May delete or irreversibly overwrite something. Deserves a confirmation prompt. |
 
 This is a coarse classification, not a permission model. It exists because
 several consumers keep reinventing the same distinction:
--- a/docs/spec/index.md
+++ b/docs/spec/index.md
@@ -78,6 +78,44 @@ The priority over which is used (CLI flag, env var, config file, default) is the
 are defined,
 so in this example it will be "CLI flag > env var > config file > default".
 
+## Command effects
+
+A command can declare what running it does to the world:
+
+```kdl
+cmd "ls"        effect="read"        help="List installed tools"
+cmd "use"       effect="write"       help="Install a tool and add it to the config"
+cmd "uninstall" effect="destructive" help="Remove a tool"
+```
+
+| Effect         | Meaning                                                                              |
+| -------------- | ------------------------------------------------------------------------------------ |
+| `read`         | Only inspects state. Running it twice is the same as running it once.                |
+| `write`        | Creates or modifies state, but removes nothing the user cannot recreate.             |
+| `destructive`  | May delete or irreversibly overwrite something. Deserves a confirmation prompt.      |
+
+This is a coarse classification, not a permission model. It exists because
+several consumers keep reinventing the same distinction:
+
+- generated documentation and `--help` can mark destructive commands
+- a wrapper script can require confirmation before running one
+- an AI coding agent can be handed an allowlist of read-only commands instead of
+  prompting on every invocation
+
+`effect` is **not inherited by subcommands**. `git remote` and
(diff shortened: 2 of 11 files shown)


jdx:
Both findings were real. Fixed in f41891d.

**effect dropped by `SpecCommand::merge`** — correct, and it matters: `merge` is how included and mounted specs are composed, so an overlay that said nothing about effects erased one already declared. Now handled like every other optional field.

Worth noting what the test turned up while I was writing it: at the *root* level `merge` replaces subcommands wholesale rather than merging them, so an overlay's `cmd "uninstall"` discards the base's entirely — every field, not just `effect`. That's pre-existing behavior and I've left it alone; the test asserts the semantics that actually exist.

**Template branches unexercised** — also correct, the kitchen-sink spec declares no effects. Added a snapshot covering all three labels plus a command with none, so a change to enum serialization or the template comparisons can't silently produce a wrong label.

---

While fixing the merge gap I audited every `SpecCommand` field against the three places that have to know about it. Three fields already fall through, unrelated to this PR:

| field | gap |
| --- | --- |
| `deprecated` | not carried by `merge` — an included spec silently un-deprecates a command |
| `restart_token` | never reaches `docs::models::SpecCommand`, so docs can't see it |
| `examples` | not written by `From<&SpecCommand> for KdlNode` — `spec.to_string()` loses them, so KDL isn't round-trip stable |

Happy to open a follow-up. Making `merge` and both `From` impls destructure their source exhaustively (no `..`) would turn each of these into a compile error instead of silent data loss, at no runtime cost.

jdx:
Both correct, fixed:

**Unused `ParseEffectError`** — dead on arrival. I wrote it, then used `bail_parse!` for the actual error so it'd carry a source span like every other parse failure, and never removed the type. Gone.

**Insta pending snapshot committed** — worse than noise: it isn't on `main`, I added it, and the test run rewrites it, which is what turned the `render produces no diff` CI check red. Untracked, and `*.pending-snap` added to `.gitignore` so it can't recur.

Also ran `prettier -w` on the docs table, which was the other half of the CI failure.

PR #740, opened 2026-07-25
refactor(spec): make missed SpecCommand fields a compile error, and fix the four that were already missed

jdx:
Follow-up to the review on #739, where `effect` was dropped by `SpecCommand::merge` and by the docs model. Neither omission could have been caught by anything but a reviewer noticing.

## The problem

Adding a field to `SpecCommand` means updating seven places — the struct, `Default`, the prop-parse arm, the child-node-parse arm, the KDL serializer, `merge`, the builder, and `docs::models::SpecCommand` + its `From`. Nothing enforces any of them, and every omission fails silently.

Auditing every existing field against the three impls turned up **four fields already falling through**, all on `main`:

| field | gap | effect |
| --- | --- | --- |
| `deprecated` | not carried by `merge` | an included spec silently un-deprecates a command |
| `restart_token` | never reaches the docs model | no template can render it |
| `examples` | not written by the KDL serializer | `spec.to_string()` drops every example |
| `help_md`, `before_help_md`, `after_help_md` | accepted only as props, serialized as child nodes | **any spec with markdown help fails to reparse**: `Error: unsupported cmd key help_md` |

The last one is the worst and I only found it because of the test in the final commit.

## The fix

**`merge` and both `From` impls now destructure their source with no `..`.** Add a field and you get three compile errors pointing at exactly the places that owe it a decision. Runtime-derived fields are bound to `_` with a comment so the reason is visible rather than implied. No macros, no new machinery, no runtime cost.

**A round-trip test covers what the compiler can't.** The parser is a `match` on node names, not a struct pattern, so a field the serializer writes and the parser rejects still compiles — which is exactly how the `help_md` bug survived. The test parses a spec exercising every field, serializes it, reparses, and compares the two serde representations. Comparing serialized values rather than named fields means a field added later is covered without anyone remembering to extend the test.

## Commits

Each is reviewable on its own, and the refactor is behavior-neutral:

1. `refactor(spec):` exhaustive destructuring — no behavior change; the four gaps are bound with `NOTE` comments so they're visible before being fixed
2. `fix(spec):` carry `deprecated` through `merge`
3. `fix(docs):` pass `restart_token` to the docs model
4. `fix(spec):` serialize `examples` to KDL
5. `fix(spec):` accept `help_md` / `before_help_md` / `after_help_md` as child nodes
6. `test(spec):` assert a spec survives a KDL round-trip

## Considered and skipped

A `#[derive(SpecNode)]` macro generating parse/serialize/merge from field attributes would cover all of it, but the irregular fields are a real tax — `deprecated` accepts bool-or-string, `alias` splits into visible and hidden, the collection children each parse differently. Destructuring plus the round-trip test gets nearly the same coverage for a fraction of the machinery. Worth revisiting if the struct keeps growing.

## Notes

- Branched off `main` rather than `#739`, so it's independently mergeable. Whichever lands second gets a trivial conflict in the destructuring lists (one added binding for `effect`).
- `subcommand_lookup` became `pub(crate)` so other modules can name it in a pattern. Still private outside the crate.
- `serde_json` added as a dev-dependency for the round-trip comparison.

## Verified

`cargo test -p usage-lib` passes, 271 + doctests. `cargo test --workspace` has the same two pre-existing failures in `usage-cli --test examples` (`test_usage_double_slash_execution`, `..._old`) present on a clean `main`. `cargo clippy` has one pre-existing error in `choices.rs` under a feature-gated path, also present on `main` and untouched here.

---

> [!NOTE]
> **Medium Risk**
> Changes core spec merge, KDL parse/serialize, and docs mapping used when composing specs (includes/mounts), but fixes are narrow and guarded by new round-trip and merge tests.
> 
> **Overview**
> **`SpecCommand` maintenance is enforced at compile time** by replacing `..` skips with exhaustive destructuring in `merge`, KDL serialization (`From<&SpecCommand> for KdlNode`), and the docs `From` impl—new struct fields must be handled explicitly in each path.
> 
> **Behavior fixes** close four silent gaps: `deprecated` is merged like other optional metadata; `restart_token` is copied into the docs model; `examples` are emitted when serializing to KDL; and `help_md` / `before_help_md` / `after_help_md` are accepted as **child nodes** on reparse (matching how they are written), so round-tripped specs no longer fail with unsupported `help_md` keys.
> 
> **Tests** add `deprecated` merge coverage and a KDL parse → `to_string()` → reparse test that compares full `serde_json` snapshots, plus fixture guards so the round-trip exercise stays meaningful. `subcommand_lookup` is `pub(crate)` only so other modules can name it in destructuring patterns.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53055a09cab9150f1c3ea3c567bd0fc0e992f41c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1760,6 +1760,7 @@ dependencies = [
  "regex",
  "roff",
  "serde",
+ "serde_json",
  "shell-words",
  "strum",
  "tempfile",
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -49,6 +49,7 @@ criterion = "0.8"
 ctor = "1"
 insta = "1"
 pretty_assertions = "1"
+serde_json = "1"
 shell-words = "1"
 tempfile = "3"
 
--- a/lib/src/docs/models.rs
+++ b/lib/src/docs/models.rs
@@ -41,6 +41,7 @@ pub struct SpecCommand {
     pub effect: Option<SpecCommandEffect>,
     pub hide: bool,
     pub subcommand_required: bool,
+    pub restart_token: Option<String>,
     pub help: Option<String>,
     pub help_long: Option<String>,
     pub help_md: Option<String>,
@@ -238,10 +239,6 @@ impl From<&crate::SpecCommand> for SpecCommand {
             subcommand_lookup: _,
         } = cmd;
 
-        // NOTE: `restart_token` is intentionally dropped, preserving existing
-        // behavior; see the follow-up commit.
-        let _ = restart_token;
-
         Self {
             full_cmd: full_cmd.clone(),
             usage: usage.clone(),
@@ -255,6 +252,7 @@ impl From<&crate::SpecCommand> for SpecCommand {
             effect: *effect,
             hide: *hide,
             subcommand_required: *subcommand_required,
+            restart_token: restart_token.clone(),
             help: help.clone(),
             help_long: help_long.clone(),
             help_md: help_md.clone(),
--- a/lib/src/docs/models.rs
+++ b/lib/src/docs/models.rs
@@ -204,35 +204,70 @@ impl From<&crate::SpecCommand> for SpecCommand {
             })
             .collect();
 
+        // Destructured exhaustively (no `..`) so that adding a field to the spec
+        // model fails to compile until this decides whether docs need it.
+        let crate::SpecCommand {
+            full_cmd,
+            usage,
+            subcommands,
+            deprecated,
+            effect,
+            hide,
+            subcommand_required,
+            help,
+            help_long,
+            help_md,
+            name,
+            aliases,
+            hidden_aliases,
+            before_help,
+            before_help_long,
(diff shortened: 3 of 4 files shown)


jdx:
Good catch, and it mattered — `before_help_md` and `after_help_md` are parser paths this branch *adds*, so the fix was shipping untested. Comparing two default values is always equal, which is exactly how a round-trip test degrades into passing for free.

Fixed by populating `before_help_long`, `after_help_long`, `before_help_md`, `after_help_md`, command-level `complete`, and `mounts` in the fixture.

I also added a guard so this can't silently regress: after the equality check, the test asserts each of those fields is actually non-default in the parsed spec, and fails with `fixture does not exercise \`<field>\`` otherwise. Verified it isn't vacuous by adding a bogus field name and watching it fail.

PR #742, opened 2026-07-25
feat(spec): allow effect= on flags and args

jdx:
Follow-up to #739. Rolling `effect=` out across mise, hk, pitchfork and communique surfaced the same gap in all four: **a command's danger often depends on how it is invoked**, and command-level `effect` cannot say that.

Real cases, all found by reviewers on those PRs rather than by me:

| invocation | effect |
| --- | --- |
| `pitchfork logs` | read |
| `pitchfork logs --clear` | deletes stored logs |
| `communique generate` | prints to stdout |
| `communique generate --github-release` | replaces a published release body |
| `mise settings foo` | read |
| `mise settings foo=bar` | writes config |

Without this, the only honest option is to label the whole command with its worst case. That makes `read` rare and costs consumers exactly the precision they wanted the field for — `pitchfork logs` becomes unavailable to a read-only allowlist because of a flag nobody passed.

## The rule

Flags and args may carry `effect=`. The effect of an invocation is the **maximum** of the command's effect and the effect of every flag and argument actually supplied.

```kdl
cmd "logs" effect="read" help="Show daemon logs" {
  flag "--clear" effect="destructive" help="Delete stored logs"
  flag "--follow"
}

cmd "settings" effect="read" {
  arg "[setting]"
  arg "[value]" effect="write"   // `settings foo` reads, `settings foo=bar` writes
}
```

Note the second one: it only works because **args** can carry it too, not just flags. The discriminator there is a positional.

`SpecCommandEffect` now derives `Ord` (read < write < destructive) so the maximum is well defined. Two helpers:

- `SpecCommand::effect_of(flags, args)` — for a consumer that parsed the command line
- `SpecCommand::max_effect()` — the pessimistic bound over every declared flag and arg, for one that didn't

## Raising only, never lowering

`--dry-run` is the obvious thing to want and is deliberately not supported. Lowering is the dangerous direction: a bug in a dry-run path would become a spec that claims a command is safe when it isn't.

Because the rule is monotonic, approximating it is always safe — a consumer with no parser takes `max_effect()` and degrades to exactly today's behavior. That's what makes this additive rather than breaking. Existing specs keep meaning what they meant.

**Most flags should declare nothing.** This is for the handful that change what a command does to the world, not an annotation for every option.

## Changes

- `effect` on `SpecFlag` and `SpecArg`, parsed as a prop or a child node, unknown values a parse error with a span
- `Ord` on `SpecCommandEffect`, plus `effect_of` / `max_effect`
- generated markdown renders a flag's effect beneath it:

  ```
  ### `--clear`

  **Effect**: destructive — may delete or irreversibly overwrite

  Delete stored logs
  ```
- docs in `docs/spec/index.md`

## Verified

`cargo test -p usage-lib` passes (283 + doctests), `clippy --all-features` and `fmt --check` clean. Five new tests cover max-composition, a flag with no effect being a no-op, arg-level raising, the unknown-value error, and the ordering itself.

The KDL round-trip fixture from #740 now includes a flag-level and an arg-level effect, **with assertions that it actually does** — the same guard added there, extended, so this can't quietly stop being covered.

---

> [!NOTE]
> **Medium Risk**
> Semver major and new spec semantics affect downstream consumers and tooling; core parsing/composition logic is well tested but `effect_of` must be wired correctly by integrators.
> 
> **Overview**
> **Flags and arguments can declare `effect=`** so a command’s danger can depend on invocation (e.g. read-only `logs` vs destructive `logs --clear`, or `settings foo` vs `settings foo=bar` via a positional). Effects only **raise** the command’s level; invocation risk is the **max** of command + supplied flags/args. **`SpecCommand::effect_of`** and **`max_effect`** implement that; **`SpecCommandEffect`** now implements **`Ord`**.
> 
> KDL **parse/serialize** and flag/arg **builders** support `effect`; generated markdown can show a flag’s effect. Spec docs describe the rules (including why lowering via `--dry-run` is not supported).
> 
> **Major release 4.0.0** (`usage-lib` / `usage-cli` and generated CLI docs). Additional library surface: **`#[non_exhaustive]`** on several spec types, **`SpecExample`** chaining helpers, and small **`new`/`effect`/`usage`** builder APIs on related types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c2de305d0a159ea1b6bd3cdf0de1bcacbbb7562. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Both review findings were real; fixed.

**`max_effect` ignores flag value args** (cursor) — correct, and it broke the exact guarantee the PR claims. `effect` parses and serializes on every `SpecArg`, including a flag's nested value arg, but composition only looked at the flag itself. A spec declaring danger on `--output <file>`'s value would report a lower tier from `max_effect`. Now a flag contributes both its own effect and its value arg's, with a test covering it.

**`effect_of` inputs vs `ParseOutput`** (greptile) — fair. `ParseOutput` includes values injected from defaults and the environment, and passing those raises the result for options nobody typed. Worth noting the direction: over-reporting is *safe* here by construction — the rule only ever raises — so this is a precision issue rather than a correctness one. Documented on `effect_of`: pass only what the command line supplied; feeding in defaults is safe but pessimistic.

---

## The semver failure is the interesting part

`lint:semver` fails, and it's not incidental:

```
--- failure constructible_struct_adds_field ---
  field SpecArg.effect
  field SpecFlag.effect
Summary semver requires new major version
```

`SpecFlag` and `SpecArg` are all-`pub`, so they're externally constructible with a struct literal and **any** new field is breaking. `SpecCommand` escaped this in #739 purely by accident — it happens to have a private `subcommand_lookup` cache field.

I audited all twelve spec structs. **Eight are externally constructible**, so a field added to any of them forces a major:

| struct | fields | constructible |
| --- | --- | --- |
| `SpecFlag` | 21 | yes |
| `SpecArg` | 16 | yes |
| `SpecConfigProp` | 6 | yes |
| `SpecExample` | 4 | yes |
| `SpecComplete` | 4 | yes |
| `SpecChoices` | 2 | yes |
| `SpecConfig` | 1 | yes |
| `SpecMount` | 1 | yes |
| `SpecCommand` | 28 | no — private field |
| the three builders | — | no — private field |

`Spec` itself is already `#[non_exhaustive]` (`lib/src/spec/mod.rs:32`), so the pattern is established in this codebase; it just was never applied to the rest.

So this PR needs a major either way. My suggestion is to spend it once: add `#[non_exhaustive]` to those eight in the same release. That's breaking on its own, but it's the *last* time a field addition breaks anything — and we've now added a spec field twice in one day.

Left that out of this PR since it's a versioning call. Happy to add it here or as a separate PR onto the same release.

jdx:
Added `#[non_exhaustive]` in this PR, plus the constructors it requires.

## The catch, and why builders had to come with it

`non_exhaustive` is only useful if the types stay constructible, and six of the eight had **no constructor at all** — only `Default`. Worse, I checked the real consumer first: **mise builds `SpecMount`, `SpecChoices`, `SpecArg` and `SpecFlag` with struct literals in seven places**, several using `..Default::default()`, which `non_exhaustive` forbids cross-crate. Shipping the attribute alone would simply have broken mise's build with no migration path.

So this commit also fills the gaps:

- **`SpecFlagBuilder` / `SpecArgBuilder`** gain `usage`, `help_first_line` and `effect` — the three fields they couldn't set. I found these by diffing builder methods against struct fields, and mise sets two of them.
- **New constructors**: `SpecMount::new`, `SpecChoices::new`, `SpecComplete::new` (+ `run`, `descriptions`), `SpecConfig::new`, `SpecConfigProp::new` (+ `env`, `help`, `default_value`)
- **`SpecExample::new`** was `pub(crate)`; now public, with `header`, `help` and `lang`

## Verified against mise, not just in theory

I patched mise to build against this branch and migrated all seven literals. It **compiles clean and emits a byte-identical spec**. That's the evidence the API is actually sufficient rather than merely plausible.

One ergonomic note worth your call: these sites map from `Option`-heavy parsed data, so a pure builder chain is clumsy — you'd need `if let Some(x) = ...` per field. Since every field stays `pub`, the natural migration is build-then-assign:

```rust
let mut arg = usage::SpecArg::builder().name(name.clone()).build();
arg.help = help;
arg.required = required;
```

That's barely worse than the literal and needs no `Option`-taking setters. I'd rather not add `maybe_help()`-style methods unless you want them.

I've kept mise's migration as a patch rather than pushing it — it belongs in mise's PR alongside the `usage-lib` 3 → 4 bump, once this releases.

## Caveat

`cargo semver-checks` isn't installed in my environment, so I couldn't run `lint:semver` locally — CI will confirm. It should now report the major cleanly rather than as an unintended break, but that's the one thing I haven't verified myself.

PR #743, opened 2026-07-25
feat(clap_usage)!: expose spec(), publish against usage-lib 4

jdx:
`clap_usage` on crates.io is **still 2.0.3, requiring `usage-lib ^2.0.3`** — even though the in-repo crate has been building against the workspace version for a while. Its source hasn't changed since 2.0.3, so release-plz never had a reason to bump it, and it silently fell two majors behind.

That pin is why **hk, pitchfork and communique couldn't adopt `effect=`**:

1. `generate()` renders straight to a writer, so the `Spec` never escapes and there's nothing to annotate; and
2. even with an accessor, 2.x has no `effect` field to set.

All three ended up inlining `generate()`'s four lines and depending on `usage-lib` directly (jdx/hk#1121, jdx/pitchfork#666, jdx/communique#240).

## The addition

`spec()` — returns the `Spec` with `bin` already set:

```rust
let mut spec = clap_usage::spec(&mut cmd, "mycli");
spec.cmd.subcommands.get_mut("rm").unwrap().effect = Some(SpecCommandEffect::Destructive);
println!("{spec}");
```

`generate()` is now literally `spec()` plus two `writeln!`s, so its output is byte-identical — there's a test asserting the header and body are unchanged.

## Version

Straight to **4.0.0**, matching `usage-lib` and `usage-cli`. A crate whose whole purpose is producing `usage-lib` types is less confusing tracking their major than carrying an independent one, and it makes the 2.x → current gap unambiguous rather than looking like a small bump.

## What this unblocks

Those three PRs can go back to using `clap_usage` instead of inlining it, which drops a direct `usage-lib` dependency from each and leaves them as just their effects tables plus a two-line change. I'll rework them once this releases — no rush on your side, they're correct either way, just more verbose than they need to be.

## Verified

`cargo test --workspace` passes, `clippy --all-features --all-targets` and `fmt --check` clean. Two new integration tests plus a compiling doctest on `spec()`.

Note the workspace `clap_usage` pin in the root `Cargo.toml` also had to move to 4.0.0 — `usage-cli` depends on it via `workspace = true`, so the build fails without it.

---

> [!NOTE]
> **Low Risk**
> Additive API with preserved `generate()` output; main impact is the intentional semver major bump for dependents upgrading from 2.x.
> 
> **Overview**
> **`clap_usage` is bumped to 4.0.0** (workspace and crate) so it tracks **`usage-lib` 4.x**, closing the gap where crates.io still shipped 2.0.3.
> 
> A new **`spec()`** builds the `usage::Spec` from a `clap::Command` (with `bin` set) and returns it for inspection or post-processing—e.g. setting **`effect=`** on subcommands that clap cannot describe. **`generate()`** now delegates to **`spec()`** and only adds the `@generated` header and write; behavior is unchanged and covered by tests.
> 
> The crate **re-exports `usage`** and documents that consumers can depend on `clap_usage` alone to name types like `SpecCommandEffect`. Integration tests assert annotation works through `clap_usage::usage` and that **`generate` output is unchanged**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d4004f640b81478226e7490eec9ac6b2f7cfb23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Good catch, and it went to the heart of the PR. Fixed.

The `spec()` example set `effect` via `usage::SpecCommandEffect`, but `clap_usage` exported only `generate` and `spec` — so a consumer depending on `clap_usage` alone couldn't name that type. They'd have had to keep the direct `usage-lib` dependency the accessor exists to remove. The documented flow literally wouldn't have compiled for them, which would have made this release pointless for hk, pitchfork and communique.

Now re-exporting the whole crate rather than a curated list, so nothing goes stale as `usage` grows:

```rust
use clap_usage::usage::SpecCommandEffect;
```

The integration test reaches the enum through `clap_usage::usage::` too — the path an external consumer actually has — so it demonstrates self-sufficiency rather than quietly leaning on a second dependency. Both doctests compile; `cargo test --workspace`, `clippy --all-features --all-targets` and `fmt --check` are clean.

Worth noting the test would have kept passing either way: an integration test in this package can reach `usage` regardless, so it was never going to catch the problem a real consumer hits. Routing it through the public path is what makes it meaningful.

PR #745, opened 2026-07-25
fix(ci): stop clap_usage changes from majoring usage-lib and usage-cli

jdx:
#744 proposes **v5.0.0 for `usage-lib` and `usage-cli` — hours after v4.0.0 shipped, with no source change in either crate.** This fixes the cause, so it stops being regenerated.

## Why v5

[`tasks/release-plz`](https://github.com/jdx/usage/blob/main/tasks/release-plz) computes one version for the whole repo:

```bash
version="$(git cliff --bumped-version)"                  # reads every commit
cargo set-version "${version#v}" --exclude clap_usage    # applies to all but clap_usage
```

#743 was `feat(clap_usage)!:`. git-cliff saw the `!` and majored the repo; `set-version` then applied 5.0.0 to every crate *except* the one that actually broke. So the two crates that didn't change get a major, and `clap_usage` gets nothing.

Scoped to the crates the version is actually set on, there is nothing to bump:

```
$ git cliff --bumped-version --include-path 'lib/**' --include-path 'cli/**'
WARN git_cliff > There is nothing to bump
v4.0.0

$ git log --oneline v4.0.0..HEAD -- lib cli
(empty)
```

Which is correct — neither crate has changed since v4.0.0.

## Three changes

**Scope the bump and changelog to `lib/` and `cli/`.** The version now reflects the crates it's set on.

**Exit before opening a release PR when the bumped version equals the current one.** Without this the corrected calculation would keep reopening a PR that bumps nothing — and it's why #744 can't just be closed today: the next run recreates it. With this, closing it sticks.

**Publish `clap_usage` when its manifest is ahead of crates.io.** The publish path only ever ran:

```bash
cargo publish -p usage-lib
cargo publish -p usage-cli
```

`clap_usage` isn't there, which is why it sat unpublished at **2.0.3 from 2025-01 until now** while the workspace built against a far newer version — and why the trusted-publisher config for it has never been exercised. Its version stays hand-managed, so the check is against the crates.io index rather than the usage-lib tag, and it publishes in dependency order (`usage-lib` → `clap_usage` → `usage-cli`).

`breaking_always_bump_major = true` is deliberately left alone. The policy is right; it was being applied to the wrong crates.

## Dry run

```
$ usage_dry_run=1 mise run release-plz
Releasing clap_usage 4.0.0
No usage-lib/usage-cli changes since v4.0.0; nothing to release
```

Exactly the two intended outcomes: `clap_usage` 4.0.0 finally ships, and no spurious v5.

## After merging

The next release-plz run will not recreate #744, so it can be closed. `clap_usage` 4.0.0 publishes through trusted publishing on that run.

---

> [!NOTE]
> **Medium Risk**
> Changes release tagging, changelog generation, and crates.io publish order; mistakes could skip needed releases or publish the wrong crate version.
> 
> **Overview**
> Fixes **release automation** so breaking `clap_usage` commits no longer force a major bump on `usage-lib` / `usage-cli` when those crates did not change.
> 
> **Version and changelog** from `git cliff` are now limited to `lib/**` and `cli/**`, matching where `cargo set-version` applies. If the bumped version equals the current `usage-lib` version, the script **exits without opening a release PR**, so empty v5-style bumps stop recurring.
> 
> Adds **`publish_clap_usage_if_needed`**: compares the manifest version to crates.io and publishes `clap_usage` after `usage-lib` (and again on the tag path when the workspace is already released), since that crate was previously never published by this script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cee8990ef904a384be15e0de46507f49e68ccd32. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tasks/release-plz
+++ b/tasks/release-plz
@@ -2,6 +2,18 @@
 # shellcheck shell=bash
 set -euxo pipefail
 
+# clap_usage is versioned by hand and released on its own cadence, so check it
+# separately: the usage-lib release cycle below would otherwise never publish it.
+# It sat unpublished at 2.0.3 from 2025-01 to 2026-07 for exactly this reason.
+clap_usage_version="$(cargo pkgid clap_usage | cut -d# -f2 | cut -d@ -f2)"
+if ! curl -fsS https://index.crates.io/cl/ap/clap_usage |
+  grep -q "\"vers\":\"$clap_usage_version\""; then
+  echo "Releasing clap_usage $clap_usage_version"
+  if [ "${usage_dry_run:-}" != 1 ]; then
+    cargo publish -p clap_usage
+  fi
+fi
+
 released_versions="$(git tag --list | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)?$')"
 cur_version="$(cargo pkgid usage-lib | cut -d# -f2 | cut -d@ -f2)"
 if ! echo "$released_versions" | grep -q "^v$cur_version$"; then
@@ -15,11 +27,16 @@ if ! echo "$released_versions" | grep -q "^v$cur_version$"; then
   exit 0
 fi
 
-version="$(git cliff --bumped-version)"
+version="$(git cliff --bumped-version --include-path 'lib/**' --include-path 'cli/**')"
+
+if [ "v$cur_version" == "$version" ]; then
+  echo "No usage-lib/usage-cli changes since $version; nothing to release"
+  exit 0
+fi
 
 if [ "${usage_dry_run:-}" == 1 ]; then
   echo "version: $version"
-  changelog="$(git cliff --bump --unreleased --strip all)"
+  changelog="$(git cliff --bump --unreleased --strip all --include-path 'lib/**' --include-path 'cli/**')"
   echo "changelog: $changelog"
   exit 0
 fi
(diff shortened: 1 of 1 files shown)


jdx:
Both findings were real; fixed.

**clap_usage published before usage-lib** (cursor, and greptile independently) — correct, and it would have been worse than a failed clap_usage publish. It depends on usage-lib at the workspace version, so on a release where both advance, cargo rejects the publish for an unpublished dependency and `set -e` aborts the script **before usage-lib goes out at all**. A release that half-fails at the first step.

Now a function called in dependency order — `usage-lib` → `clap_usage` → `usage-cli` — and called once more on the path where usage-lib is already released and only clap_usage is behind. That second call is today's situation, and without it clap_usage would stay stuck until the next lib release.

**Changelog bump not path scoped** (cursor) — also correct. I scoped `--bumped-version` and the dry-run changelog but missed `git cliff --bump -o CHANGELOG.md` and the PR body, so a clap_usage breaking commit could still drive the changelog to a major while `cargo set-version` applied the scoped version — mismatched artifacts in the same release. All four `git cliff` calls are scoped now.

Dry run on the current state is unchanged and still what we want:

```
Releasing clap_usage 4.0.0
No usage-lib/usage-cli changes since v4.0.0; nothing to release
```

`shellcheck tasks/release-plz` is clean.

PR #750, opened 2026-07-27
perf: split measuring from reporting, and drop the dead dispatch trigger

jdx:
Two review findings from #748, which merged before I pushed the fixes.

**The measure job held `pull-requests: write` while running pull-request code.** A same-repo pull request could persist state through the runner environment files and reach a token able to write comments elsewhere in the repository. Measuring now runs with `contents: read` and hands its report over as an artifact; a second job holds the write token, checks out nothing, and runs no project code.

**`workflow_dispatch` was dead.** It was listed as a trigger while the only job required `github.event_name == 'pull_request'`, so a manual run skipped everything and reported success without measuring. Removed rather than made to work — there is no pull request to compare against on a manual run.

The gate survives both. A missing report is written out as an explicit failure rather than left for the reporting job to find nothing and infer a pass, and `always()` on the job and the step keeps a comment failure from skipping it.

zizmor at `--persona=pedantic` and actionlint are clean.

---

---

> [!NOTE]
> **Medium Risk**
> CI-only change with a security win (write token isolated from PR code), but a probable `GH_usage` typo could break PR comments until corrected.
> 
> **Overview**
> **Splits `perf-pr` into `measure` and `report` jobs** so PR code runs only under `contents: read`; the sticky comment and regression gate move to a second job that downloads a `tak-report` artifact, never checks out the repo, and alone holds `pull-requests: write`.
> 
> The measure job **packages** report markdown, gate exit code (default **2** when compare never ran), and short SHAs, then **uploads** them. The report job posts the comment from the artifact and **fails the workflow** via a `case` on status (0 pass, 2 = earlier failure, other = regression).
> 
> **Removes** `workflow_dispatch` and the job guard that required `pull_request` (dispatch could succeed without measuring). Drops the redundant `if: github.event_name == 'pull_request'` on the measure job.
> 
> **Note:** the comment step renames `GH_REPO` to `GH_usage` in env and `gh api` paths — likely unintentional and would break API URLs unless fixed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da7492764bce0be8578b3dffa38a9a7a3307041b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Chores**
  * Updated pull request performance checks to run automatically on pull request events (no manual dispatch).
  * Improved performance reporting by separating measurement and comment/reporting steps.
  * Added more reliable artifact packaging and status handling, even when comparison output is missing.
  * Automatically updates a single pull request comment with instruction-count results.
  * Workflow now fails when instruction counts exceed the configured threshold.

### DIFF
--- a/.github/workflows/perf-pr.yml
+++ b/.github/workflows/perf-pr.yml
@@ -16,7 +16,6 @@ name: perf-pr
 on:
   pull_request:
     types: [opened, synchronize, reopened]
-  workflow_dispatch:
 
 permissions:
   contents: read
@@ -30,9 +29,8 @@ env:
   CARGO_INCREMENTAL: "0"
 
 jobs:
-  compare:
+  measure:
     name: Compare instruction counts
-    if: github.event_name == 'pull_request'
     # Same runner class as perf.yml and perf-backfill.yml. Absolute counts shift
     # between machine types by more than a real regression does, so a comparison
     # across runner classes is not a comparison — tak refuses to make one, and
@@ -40,8 +38,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 40
     permissions:
+      # Read only. This job builds and runs code from the pull request, so it
+      # must not hold a token that can write anything — see the `report` job.
       contents: read
-      pull-requests: write # the sticky comment
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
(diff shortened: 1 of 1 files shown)


jdx:
Fixed, plus two failures of my own that the reviews did not catch but CI did.

**The gate reported a regression that was never measured.** A missing status file was packaged as `1`, so `Fail on a regression` said "an instruction count rose beyond the gate" for a run where nothing was compared. fnox demonstrated it exactly: the build failed for a missing `libudev`, and the check blamed a regression. Missing now packages as `2` and the two get different messages.

**fnox: I broke my own libudev fix.** I added it to both workflows, then overwrote `perf-pr.yml` with a template that predated the fix and only re-committed `perf.yml` — so the dependency was present in the workflow that did not need it yet and absent from the one failing. Restored.

**hk: `taplo` formatting.** hk aligns its TOML and the lint job rejected the unaligned `tak.toml`. Ran `taplo format`.

## Not a bug: "baseline notes are never fetched"

`tak compare` fetches them itself — `notes::read` calls `fetch(remote)` before reading, and `cmd_compare` passes `Some(remote)` for the base. The workflow deliberately adds no notes refspec; hiding that seam is what tak is for.

Confirmed in production rather than by reading. aube's run produced real base numbers and ten-point sparklines, which cannot happen without the baseline.

## Not fixable: "artifact status bypasses the gate"

Correct as stated, and I do not think it can be closed. The measuring job runs code from the pull request, so anything it produces — the status file, the report, the numbers themselves — is under that code's control. Moving the verdict into the trusted job does not help, because the trusted job would still be deriving it from output the untrusted job produced.

This is the same boundary as the backfill validator in aube: **a subject that lies about its own performance cannot be caught, because the subject is the measurement.** What the job split does buy is that a compromised measurement cannot reach a token and write elsewhere in the repository. For a same-repo pull request the author already has push access, so the gate is a guard against mistakes, not against its own author.

Worth stating in the workflow rather than leaving implied — I will add a comment saying so.

---

jdx:
Checked all three against runs rather than reasoning about them.

## Not a bug: artifact jobs lack `actions` scope

`upload-artifact` and `download-artifact` within a single run use the runtime token, not `GITHUB_TOKEN`, so the job-level `permissions` block does not gate them. `actions: read` is needed for *cross-run* artifact access, which this does not do.

Confirmed on a real run of the split workflow (jdx/hk#1125, run 30287239644):

```
Compare instruction counts: success
    [success] Package the report
    [success] Run actions/upload-artifact
Report and gate: success
    [success] Run actions/download-artifact
    [success] Comment on the pull request
    [success] Fail on a regression
```

and the comment landed. Adding the scope would widen the token for no reason.

## Not a bug: baseline notes are never fetched

`tak compare` fetches them itself — `notes::read` calls `fetch(remote)` before reading, and `cmd_compare` passes `Some(remote)` for the base. The workflow deliberately adds no notes refspec; hiding that seam is what tak is for.

Also visible in the run above. The comment says *"Nothing was compared"* and lists both benchmarks as new — which is the correct answer for a repository where `perf.yml` has not yet recorded on main, not a symptom of a missing fetch. aube, where main does have measurements, produced a full table with ten-point sparklines from the same workflow.

## What was real, and is fixed

The gate reported a regression that was never measured: a missing status file packaged as `1`, so a failed build produced *"an instruction count rose beyond the gate"*. fnox demonstrated it exactly. Missing now packages as `2`, and the two get different messages.

---

PR #751, opened 2026-07-27
feat(cli): declare what each usage command does to the world

jdx:
usage added `effect=`, shipped it to mise, hk, pitchfork, aube and communique — and declared nothing for its own 16 commands. Anything pointed at usage's spec sees empty fields on the tool that invented the feature.

| command | effect | flags that raise it |
|---|---|---|
| `complete-word` | read | |
| `generate` | read | |
| `generate completion` | read | |
| `generate completion-init` | read | |
| `generate fig` | read | `--out-file` → write |
| `generate json` | read | |
| `generate manpage` | read | `--out-file` → write |
| `generate markdown` | read | `--out-dir`, `--out-file` → write |
| `generate sdk` | **write** | |
| `lint` | read | |
| `sponsors` | read | |

## Why the shape is interesting

Most of these print to stdout and take an optional flag to write to a file instead, so they sit at `read` with the flag raising them. That's the composition rule earning its place rather than being asserted in a doc — `usage g markdown -f x.kdl` only reads; the same command with `--out-file` writes.

`generate sdk` is the exception. Its `-o --output` is *required*, so there is no read-only way to invoke it and the effect belongs on the command, not a flag.

## Unclassified, deliberately

`bash`, `fish`, `zsh`, `powershell` and `exec` run a script the user supplied, so their effect is whatever that script does. Labeling them would be a lie in whichever direction it was labeled, and `read` in particular would be dangerous. A test asserts every unclassified command has an entry in `UNCLASSIFIED` with its reason, so a new command can't go unlabeled by accident.

## How

clap can't express this, so it's applied to the derived spec through `clap_usage::spec` (#743), the same shape as mise's `command_effects`. Five tests cover stale command paths, stale flag names, the unclassified guard, and both composition cases above.

Also emits `min_usage_version "4.0"` — an older `usage` rejects a spec carrying `effect=` with "unsupported cmd prop effect", so the declaration and the floor move together. mise does the same.

The generated docs pick it up for free: `usage generate sdk` now reads **Effect: modifies state**, and `--out-file` on `markdown` reads **Effect: modifies state** under a read-only command.

---

> [!NOTE]
> **Low Risk**
> Metadata and spec-generation path changes only; no runtime command behavior changes beyond richer `--usage-spec` output.
> 
> **Overview**
> Adds **`command_effects`** so the `usage` CLI’s own `--usage-spec` output carries `effect=read` / `effect=write` on commands and flags (clap can’t express this; it’s patched onto the derived spec like mise/hk).
> 
> Most generators are **read**; **`--out-file`**, **`--out-dir`**, etc. raise matching flags to **write**. **`generate sdk`** is **write** at the command level because output is required. **`bash`**, **`exec`**, and other script runners stay **unclassified**, with tests guarding stale paths and accidental gaps.
> 
> **`usage_spec::generate`** now builds the spec via `clap_usage::spec`, runs `apply`, and emits **`min_usage_version "4.0"`** alongside the KDL. Regenerated **`usage.usage.kdl`**, **`commands.json`**, and CLI reference docs show read-only vs modifies-state in docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ee7a02148313b71e479a2db9706c5c45d4a9ae7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added effect metadata to CLI commands and output flags, distinguishing read-only operations from actions that modify files or state.
  - Added minimum usage specification version metadata.

- **Documentation**
  - Updated CLI reference documentation and command metadata to display read-only and state-modifying effects.
  - Clarified effects for generation commands and file output options.


jdx:
The `markdown` benchmark is +1.09%, over the 1% gate. It isn't a code regression — the benchmark's input is the file this PR is meant to grow.

`tak.toml` points the `markdown` bench at `cli/usage.usage.kdl`, and this PR adds 11 `effect=` props and 4 flag effects to it. Measured with the **unchanged `main` binary** against both spec files:

| | bytes in | bytes of markdown out |
|---|---:|---:|
| `main` spec | 9,418 | 12,208 |
| this PR's spec | 9,627 | 12,589 (+3.1%) |

Same binary, more input, more output. And no code on the `g markdown` path changed at all — the Rust diff is `cli/src/command_effects.rs` (new), `cli/src/lib.rs` (one `mod` line) and `cli/src/usage_spec.rs`, and `command_effects::apply` is reachable only from `--usage-spec`. The renderer is byte-identical to `main`.

So the gate fired on the corpus growing, which means it will fire on any PR that adds a command or a flag — for doing its job. That's a property of pointing the bench at a file that changes with the CLI, not of this change.

Two ways forward, and I'd rather you picked than have me quietly widen your gate:

1. **Merge this as-is.** The +1.09% is real work a user asked for: 15 more lines of rendered docs. Nothing got slower per unit of output.
2. **Freeze the bench input first.** Copy today's spec to a fixture, point `tak.toml` at that, and the bench measures "did the renderer get slower" instead of "did the CLI grow". Refreshed deliberately when someone wants the corpus updated. I'll open that as its own PR if you want it — it stands on its own merits and unblocks every future spec-growing change, but it does drop the signal that a growing spec costs users time, so it's your call.

PR #752, opened 2026-07-27
fix(parse): keep a re-declared global's aliases on one flag

jdx:
## What

`merge_subcommand_flags` in [`lib/src/parse.rs`](https://github.com/jdx/usage/blob/main/lib/src/parse.rs) could leave one logical flag registered under two different `Arc<SpecFlag>`s in `available_flags`. The collision guard now compares flag *origins* instead of `Arc` identity.

## Why

Reproduction: a global declared with both a short and a long (`flag "-y --yes" global=#true`), re-declared non-globally by a subcommand (`flag "-y --yes"`).

The child's flag keys iterate in `BTreeMap` order, so `--yes` comes before `-y`:

1. Key `--yes`: `inherited_global` resolves to the original global `A`. `available["--yes"]` is also `A`, so `Arc::ptr_eq` holds and the skip guard doesn't fire. A merged flag `M` is built and inserted at `--yes`.
2. Key `-y`: `inherited_global` now resolves via `available["--yes"]`, which is `M`. The guard checks `available["-y"]` — still `A` — and `A.global && !Arc::ptr_eq(A, M)` is true, so it `continue`s. The `-y` key keeps pointing at the pre-merge `A`.

The guard's stated intent is "never clobber a *different* inherited global's alias", but `M` isn't a different global — it was derived from `A` in this same loop, so the identity comparison misfires.

Impact is mild but real. Parsing itself still works because lookups are by key, so this only shows up in code that reads the flag *objects* rather than the keys. When the re-declaration adds a third alias (global `-y --yes`, child `-y --yes --assume-yes`), the flag reachable via `-y` is missing `assume-yes` from its `long` list.

## How

Track which inherited global each merged flag was derived from (`merged_origin`), and resolve a flag to its origin — itself, or for a merged flag its source global — before comparing. The guard now skips only when the existing key's origin differs from the incoming global's origin, so a flag this loop just merged is no longer mistaken for a colliding global.

Chained merges are handled too: a new merged flag inherits `global_origin` rather than pointing at its immediate base, so if the global has further aliases still to be processed they all resolve back to the same origin.

## Reviewer notes

- New test `test_redeclared_global_aliases_share_one_flag` covers the three-alias case and asserts all three keys resolve to a flag whose `long` contains every alias, that `unique_flags` sees exactly one object, and that the global's `effect="write"` survives the merge. Confirmed it fails on the pre-fix code with exactly the reported symptom (`-y` → `long: ["yes"]`) and passes after.
- The genuine-collision behavior is unchanged and still covered by the existing `test_orphan_short_does_not_clobber_unrelated_global`: an unrelated global (`-r --restrict`) keeps its alias, as global precedence dictates.
- 296 `usage-lib` tests pass; `cargo clippy --all --all-features -- -D warnings` and `cargo fmt --all --check` are clean. Two failures in `usage-cli --test examples` (`test_usage_double_slash_execution{,_old}`) are pre-existing and environmental — they fail identically on clean `main`.
- Follow-up, not in this PR: `usage::available_flags` on `feat/usage-mcp` (#746) works around this by collapsing duplicates by name after the pointer dedup. That branch isn't an ancestor of `main`, so I left it alone; its workaround can be simplified when it next rebases onto this fix.

---

> [!NOTE]
> **Medium Risk**
> Touches subcommand descent flag-merging in the parser (`available_flags`), which affects completions and metadata on flag objects; behavior is narrow and covered by new tests, with existing collision tests unchanged.
> 
> **Overview**
> Fixes **`merge_subcommand_flags`** so a subcommand that re-declares an inherited global no longer leaves different alias keys (`-y`, `--yes`, etc.) pointing at different `Arc<SpecFlag>` objects with inconsistent merged metadata.
> 
> The collision guard now compares **flag origins** (via `merged_origin` / `origin_of`) instead of `Arc` pointer identity, so a merged flag from an earlier loop iteration is not treated as a competing global. On the first merge for a child re-declaration, **all existing map entries** for that global’s aliases are rebound to the single merged flag—covering cases where the child only re-declares a subset of the global’s aliases.
> 
> Adds regression tests for full and partial re-declaration; unrelated global short collisions are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae7be7f1c8708279c93a26446174a053a231ed6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed alias handling for inherited global options when subcommands redefine those options.
  * Ensured all aliases consistently resolve to one option and preserve the complete alias set.
  * Corrected behavior when a subcommand adds a new alias or omits an existing alias during redefinition.

### DIFF
--- a/lib/src/parse.rs
+++ b/lib/src/parse.rs
@@ -104,9 +104,9 @@ fn merge_subcommand_flags(
             }) {
                 continue;
             }
-            let merged = merged_cache
-                .entry(Arc::as_ptr(&flag) as usize)
-                .or_insert_with(|| {
+            let merged = match merged_cache.get(&(Arc::as_ptr(&flag) as usize)) {
+                Some(merged) => merged.clone(),
+                None => {
                     let mut merged = (*global_flag).clone();
                     for s in &flag.short {
                         if !merged.short.contains(s) {
@@ -118,10 +118,22 @@ fn merge_subcommand_flags(
                             merged.long.push(l.clone());
                         }
                     }
-                    Arc::new(merged)
-                })
-                .clone();
-            merged_origin.insert(Arc::as_ptr(&merged) as usize, global_origin);
+                    let merged = Arc::new(merged);
+                    merged_cache.insert(Arc::as_ptr(&flag) as usize, Arc::clone(&merged));
+                    merged_origin.insert(Arc::as_ptr(&merged) as usize, global_origin);
+                    // Rebind the global's *other* aliases onto the merged flag. The loop only
+                    // visits keys the child declared, so an alias the child left out (the `-y` of
+                    // a `-y --yes` global re-declared as just `--yes`) would otherwise keep
+                    // pointing at the pre-merge flag and miss the aliases just unioned in. One
+                    // logical flag must be one object under every key it answers to.
+                    for existing in available.values_mut() {
+                        if origin_of(&merged_origin, existing) == global_origin {
+                            *existing = Arc::clone(&merged);
+                        }
+                    }
+                    merged
+                }
+            };
             available.insert(key, merged);
             continue;
         }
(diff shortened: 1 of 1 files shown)


jdx:
Rebased onto `main` and addressed the review feedback.

### CodeRabbit's `lib/src/parse.rs` finding — valid, fixed in ae7be7f

> Update the alias-merging logic around `merged_origin` so every inherited alias for the merged flag is rebound to the merged `Arc`, including aliases omitted by the child declaration; do not redirect only the currently iterated child key.

Correct, and it's the same defect this PR set out to fix reached by a different route. The merge loop only visits the alias keys the *child* declared, so any alias of the global the child left out kept pointing at the pre-merge flag. A `-y --yes --confirm` global re-declared as just `--yes --assume-yes` left `-y` and `--confirm` on a flag whose `long` was missing `assume-yes`.

Fixed by rebinding every key that shares the merged flag's origin global when the merged flag is first built (once per re-declared child, not per alias key). Added `test_partially_redeclared_global_keeps_all_aliases_on_one_flag` covering exactly the requested case — a parent flag with multiple aliases where the child re-declares only one — asserting both shared identity (`unique_flags(...).count() == 1`) and that every alias key carries the full `long` list. Verified it fails before the fix with `-y` → `long: ["yes", "confirm"]` and passes after.

The genuine-collision behavior is unchanged: the rebind only touches keys whose value already resolves to the *same* origin global, so an unrelated global keeps its alias, as `test_orphan_short_does_not_clobber_unrelated_global` still verifies.

### The `.github/workflows/perf*.yml` comments — no longer part of this PR

The notes-baseline finding (Cursor Bugbot, Greptile P1, CodeRabbit) and the runner-pinning and `workflow_dispatch` comments were all against `perf-pr.yml` / `perf.yml` / `tak.toml` / `mise.toml`. Those files were only in this PR's diff because the branch carried an unmerged copy of the perf commit; that work landed on `main` separately via #748. The rebase dropped the duplicate, so this PR is now a single-file change to `lib/src/parse.rs` and none of that feedback applies here. Worth re-filing against `main` if the notes-fetch gap is still real there — it looked like a legitimate point, just not one this PR can fix.

### Verification

297 `usage-lib` tests pass; `cargo clippy --all --all-features -- -D warnings` and `cargo fmt --all --check` clean. The two `usage-cli --test examples` failures (`test_usage_double_slash_execution{,_old}`) are pre-existing and environmental in my sandbox — they fail identically on unmodified `main`.

PR #801, opened 2026-08-11
feat(argv): emit a usage spec from static metadata

jdx:
Next step toward the derive: the cold half of what it will emit. Ticks two boxes in `PLAN.md`.

## The split

A CLI knows more about itself than a parse needs. `usage-argv`'s tables hold what a token is matched against; everything else — help text, choices, defaults, effects, mounts — now lives in a parallel tree behind a new `spec` feature.

Two reasons, and the second is the one that matters:

- A successful parse never reads any of it, so keeping it out of the hot tables keeps them dense.
- **The metadata borrows the table entry it describes.** `FlagMeta.flag: &Flag` means a flag's long and short forms have exactly one definition, so what the docs say and what the parser matches cannot drift apart. Nothing is duplicated between the two trees.

A `debug_assert` also checks the two are in step, since they are written in the same order by construction.

## Emission

`Spec::to_kdl` writes the spec by hand rather than through the `kdl` crate — this crate has no dependencies and I would rather keep it that way for a cold path that only has to be *correct*.

Correct is verified rather than eyeballed: **13 tests parse the output back with usage-lib and check the resulting `Spec` field by field**, over a fixture chosen to be awkward — help text containing a quote, multi-line long help, a hidden alias next to a visible one, a negated flag, choices on both a flag and an argument, every `double_dash` mode, both effect levels, a mount, a restart token, and two levels of nesting. Then it renders through usage-lib's markdown and manpage generators, which is the same code an adopter's `usage g` build runs, so a spec that parses but renders to nothing fails here instead of in someone's docs.

There is also an insta snapshot of the emitted text, so a change to the writer has to be looked at rather than inferred from assertions still passing.

Writing the test found a real bug immediately: I had invented a `help_heading` child node that the spec does not have, and the whole document failed to parse.

## One thing deliberately missing

`help_heading` groups flags under a heading in help output. clap has it, mise uses it (in `watch`, for its vendored watchexec arguments), and **the spec has no node for it**. Rather than carry the field and silently drop it on the way out — which would make the emitted KDL a summary rather than a definition — it is absent, and `PLAN.md` gains a "spec gaps found on the way" section proposing it. Per the canonicality rule, the spec gets extended first.

## Next

The derive itself, which generates both trees from a Rust type. This PR is what it will emit; the macro is the part that writes it for you.

---

> [!NOTE]
> **Low Risk**
> Large additive surface behind an off-by-default feature; usage-lib serialization changes are targeted at round-trip correctness for existing spec shapes.
> 
> **Overview**
> Adds an optional **`spec`** feature on `usage-argv` with a **cold metadata tree** (`Spec`, `CommandMeta`, `FlagMeta`, `ArgMeta`) that **borrows** the existing parse tables (no duplicated flag names) and **`Spec::to_kdl()`** — dependency-free, hand-written KDL for docs/completions downstream.
> 
> **Conformance** turns on `spec` and adds round-trip tests: parse emitted KDL with **usage-lib**, assert fields on an awkward fixture, **insta** snapshot stability, and markdown/manpage rendering. **`PLAN.md`** marks metadata + KDL emission done and documents deliberate omissions (`help_heading`, root `mount`).
> 
> **usage-lib** fixes KDL **re-serialization** so node arguments starting with `-` and help text with control characters (e.g. ANSI) stay quoted/escaped and **reparse** correctly — aligned with the argv writer’s quoting rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce12211c9c55a83adb64babefc960a5e0c4d103e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## New Features
- Added an optional specification feature for exporting CLI metadata.
- Added KDL output for commands, flags, arguments, examples, defaults, choices, effects, mounts, and restart tokens.
- Added validation and round-trip support for generated specifications.
- Verified metadata through Markdown and manpage rendering.

## Documentation
- Documented the optional specification feature.
- Added notes on remaining specification gaps and prerequisites for future derive support.

## Bug Fixes
- Improved handling of quoted, escaped, and dash-prefixed values in generated specifications.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2014,6 +2014,7 @@ dependencies = [
 name = "usage-conformance"
 version = "0.0.0"
 dependencies = [
+ "insta",
  "serde",
  "serde_json",
  "usage-argv",
--- a/PLAN.md
+++ b/PLAN.md
@@ -129,6 +129,10 @@ carrying them rather than being worked around.
       it and mise uses it (in `watch`, for its vendored watchexec arguments), and
       the metadata tree deliberately omits it rather than dropping it silently on
       the way out.
+- [ ] **A mount on the root command** — the spec accepts `mount` only inside a
+      `cmd` block, so a CLI whose _top-level_ subcommands are discovered by running
+      something cannot say so. Worth deciding whether that is a gap or a deliberate
+      restriction.
 
 ### Then: what a CLI framework has to have
 
--- a/PLAN.md
+++ b/PLAN.md
@@ -97,14 +97,16 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
 
 ### Next: the derive
 
-- [ ] **Static metadata tables** — a second, cold table holding what the hot one
+- [x] **Static metadata tables** — a second, cold tree holding what the hot one
       deliberately omits: help and long help, about text, hidden-ness, visible and
-      hidden aliases, value names and hints, `choices`, defaults, `env`, effects,
-      mounts, examples. Separate from the parse tables so the hot path keeps its
-      cache locality.
-- [ ] **KDL emission** — write the spec from those tables without depending on
-      usage-lib, which pulls kdl, miette, tera, and regex. Verified by
-      round-tripping the output through `Spec::from_str` and comparing.
+      hidden aliases, value names, `choices`, defaults, `env`, effects, mounts,
+      restart tokens, examples. Behind the `spec` feature, and each entry borrows
+      the parse-table entry it describes, so names have one definition and cannot
+      drift.
+- [x] **KDL emission** — `Spec::to_kdl`, written by hand so the crate keeps having
+      no dependencies. Verified by parsing the output back with usage-lib and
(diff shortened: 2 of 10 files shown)


jdx:
All four bot findings on this PR are fixed in the two commits since — the comments above are re-posts against the cumulative diff rather than the current state.

| finding | fix | test |
| --- | --- | --- |
| flag `effect` never written | emitted, like commands' | `a_flag_can_carry_an_effect` |
| several `default`s collapse | child block, since KDL properties are unique per node | `several_defaults_all_survive` |
| root examples dropped | root nodes written at the top level, where they belong | `root_level_examples_survive` |
| root `mount` emits unparsable KDL | not emitted — the spec accepts `mount` only inside `cmd`, so it is a debug assertion plus a `PLAN.md` entry asking whether it should be expressible | n/a |

16 round-trip tests, and the whole workspace is green.

PR #802, opened 2026-08-11
feat(spec): add help_heading, and render it

jdx:
Stacked on #801 — review that first; this PR's diff will include it until it merges.

Closes the gap #801 turned up: writing the spec emitter, I reached for `help_heading` and found the spec has no such field.

## Why it matters more than it looks

A CLI with dozens of flags is unreadable without sections, and clap has had `help_heading` for years. Because the spec could not record one, **`From<&clap::Arg>` was silently dropping it** — so every CLI in the fleet that groups its flags has been losing the grouping on the way into its spec. mise groups its entire `watch` passthrough set that way (31 uses, all in one file).

That is the part worth noting: this was not a missing feature so much as a leak that was invisible because nothing downstream could have shown it.

## The spec side

- `help_heading` on `SpecFlag` and `SpecArg`, as a property (`help_heading="Filtering"`) or a child node for longer text, written back out.
- Mapped from clap's `get_help_heading()` for both flags and positionals.
- Builder setters, docs in the flag and arg references, tests for the round trip and the clap conversion.
- `usage-argv` carries it in both `FlagMeta` and `ArgMeta`, which is what prompted this: the derive cannot emit what the spec cannot express, so per the canonicality rule the spec went first.

## The rendering side

Help output and generated markdown both group by heading now — a field nothing displays is half a feature.

Grouping happens in the docs models, not in a template: Tera can filter on an attribute's *value* but cannot partition on one, and "everything without a heading" is not expressible as a filter. Behaviour:

- Unheaded entries keep the default section title (`Flags:` / `Arguments:`) and come **first**.
- Each heading gets its own section, in the order the headings first appear.
- A heading with nothing *visible* in it produces no section, and a CLI that heads every flag gets no empty `Flags:`.
- Positionals group too, since the spec field is on both.

One thing worth knowing if you touch this: the groups hold clones, and `render_md` mutates the flag and arg lists *after* the model is built — so grouping at construction published copies without their rendered markdown. That cost me a debugging round; groups are rebuilt at the end of `render_md` and the method that does it says why.

**Every existing snapshot is unchanged** — output is byte-identical when nothing has a heading — and four new ones cover the grouped case across both renderers, including hidden entries. `mise run render` produces no diff.

---

> [!NOTE]
> **Low Risk**
> Presentational metadata and docs/help rendering only; parsing behavior is unchanged and output without headings remains identical.
> 
> **Overview**
> Adds **`help_heading`** to flags and positionals in the usage spec (KDL parse/serialize, builders, reference docs) and threads it through **`usage-argv`** metadata and KDL emission so derive output can record section titles losslessly.
> 
> **Clap bridge fix:** `From<&clap::Arg>` now maps `get_help_heading()`, so grouped flags are no longer dropped when converting to a spec.
> 
> **Rendering:** CLI help (short/long templates) and generated markdown partition flags and args by heading via `flag_groups` / `arg_groups` in the docs models. Unheaded items stay under default **Arguments** / **Flags** and appear first; custom headings follow in first-seen order; sections with only hidden entries are omitted. Markdown regroups after `render_md` so grouped clones keep rendered help text.
> 
> Conformance roundtrip tests and new snapshot tests cover grouped help and global flags; existing output stays byte-identical when no headings are set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da412c73921f96714b8cdcf0cfd492228d086eef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added configurable help headings for flags and arguments.
  * Help and Markdown output now groups entries under their headings while preserving declaration order.
  * Empty sections and hidden entries are omitted from generated documentation.
  * Global and local flags are displayed in separate grouped sections.
* **Documentation**
  * Added specification examples and reference documentation for configuring help headings.
* **Bug Fixes**
  * Improved consistency between generated help output and Markdown documentation.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -128,9 +128,9 @@ carrying them rather than being worked around.
 - [x] **`help_heading`** — grouping flags under a heading in help output. Now a
       spec field on both flags and arguments, and the clap bridge no longer drops
       it, which it had been doing for every CLI that groups its flags.
-- [ ] **Rendering headings** — the spec can express a heading and nothing displays
-      one yet. Grouping flags in help and markdown output needs the docs models and
-      the templates to change, which is its own PR.
+- [x] **Rendering headings** — help output and generated markdown both group by
+      heading now. Unheaded entries keep the default section and come first, and a
+      heading with nothing visible in it produces no section.
 - [ ] **A mount on the root command** — the spec accepts `mount` only inside a
       `cmd` block, so a CLI whose _top-level_ subcommands are discovered by running
       something cannot say so. Worth deciding whether that is a gap or a deliberate
--- a/PLAN.md
+++ b/PLAN.md
@@ -125,10 +125,12 @@ Each of these is a thing a CLI wants to say that the spec has no way to record.
 Per the canonicality rule the spec gets extended first, so these block the derive
 carrying them rather than being worked around.
 
-- [ ] **`help_heading`** — grouping flags under a heading in help output. clap has
-      it and mise uses it (in `watch`, for its vendored watchexec arguments), and
-      the metadata tree deliberately omits it rather than dropping it silently on
-      the way out.
+- [x] **`help_heading`** — grouping flags under a heading in help output. Now a
+      spec field on both flags and arguments, and the clap bridge no longer drops
+      it, which it had been doing for every CLI that groups its flags.
+- [ ] **Rendering headings** — the spec can express a heading and nothing displays
+      one yet. Grouping flags in help and markdown output needs the docs models and
+      the templates to change, which is its own PR.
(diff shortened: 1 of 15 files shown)


jdx:
Both findings — `ArgMeta` and `write_arg` missing `help_heading` — are fixed in the current commit. The spec field is on flags *and* arguments, so the metadata had to be too; the fixture now puts a heading on the root's positional and `a_help_heading_survives` asserts both survive the round trip.

Rebased on the latest #801, which picked up its four fixes as well.

PR #803, opened 2026-08-11
feat(derive): compile a struct into parse tables and a spec

jdx:
Stacked on #802 (which is stacked on #801) — review those first; this diff includes them until they merge.

The piece the last three PRs were building toward: one Rust type in, a parser and a spec out.

```rust
/// A tool that does things
#[derive(usage::Cli)]
#[usage(bin = "ex", version = "1.0")]
struct Cli {
    /// How many jobs to run at once
    #[usage(short = 'j', long, env = "EX_JOBS", default = "4")]
    jobs: Option<String>,

    /// Colorize output
    #[usage(long, negate = "--no-color", default = "true")]
    color: bool,

    /// Files to process
    files: Vec<String>,
}
```

That gives you `Cli::parse_from(argv)`, `Cli::command()`, `Cli::spec()`, and `Cli::to_kdl()` — so the same declaration feeds `usage g markdown|manpage`, the completion generators, and grouped help output from #802.

## What's generated

Three things, and the split is the design:

- **`static` parse tables** — all a successful parse reads.
- **`static` metadata** — what spec emission and help need, which a parse never touches.
- **a parse function** — a `match` on table keys assigning straight into the struct's fields. No map to build and read back, nothing allocated that does not end up in the result.

`command()` returns a `&'static`, so there is no command tree to construct before parsing starts. A test asserts the pointer is the same every call, which is the property the whole project exists for.

A field with `long` or `short` is a flag; anything else is positional. Help comes from the doc comment — first paragraph short, whole comment long.

## Scope, stated plainly

**One command per struct.** Subcommands need an enum of variants, cross-type table references, and a nested path through the parse function; that is its own PR and a box in `PLAN.md`.

**Values are text** — `bool`, `String`, `Option<String>`, `Vec<String>`, or an unsigned integer with `count`. Converting to other types is also where `env`, required-ness, and `choices` get enforced, and that layer does not exist yet. So `Option<u32>` is a *compile error* explaining exactly that, rather than something that silently half-works.

## The error messages got real attention

They are the surface an author actually interacts with, so:

- `short = "j"` → *a short flag is a character: write `short = 'j'`*
- a duplicate `--flag` → points at both declarations, second first
- an argument after a variadic one → *can never be filled, because the variadic takes every remaining word*
- an unknown option → lists the ones that exist
- `count` on a `String` → says it has to be an unsigned integer

## Tests

Twelve, over a deliberately awkward CLI: attached and bundled shorts, `--flag=value`, repeated flags, a negation turning off a default, a hidden flag that still parses, `--` passthrough, and a typo reported rather than bound. Then the same declaration is checked to emit a spec usage-lib accepts field by field, render as markdown and a manpage, and group by heading. Two more CLIs cover the empty cases — no flags, no positionals — which is where generated code tends to break on an unused variable or an empty `match`.

Not published: a CLI framework that cannot express subcommands is not one to depend on by accident. The version tracks the workspace so it is ready the moment it can.

---

> [!NOTE]
> **Medium Risk**
> Large new proc-macro surface that defines CLI parsing behavior for adopters; mitigated by extensive conformance tests but still pre-v1 (no subcommands, limited types, env not enforced at parse time).
> 
> **Overview**
> Adds the **`usage-derive`** workspace crate and **`#[derive(Cli)]`**, so a single struct with `#[usage(...)]` attributes becomes a **`usage-argv` parser**, **static spec metadata**, and **KDL** for docs/completions.
> 
> Generated code exposes **`parse_from` / `parse`**, **`command()`** and **`spec()`** as `&'static` tables, and **`to_kdl()`**. Parsing is a direct `match` on flag/arg keys into prefixed locals (avoids field-name clashes). The model layer rejects invalid declarations at compile time (duplicate flags, `var` vs `variadic`, unsupported types, dashed long/name normalization, etc.) with targeted errors.
> 
> **v0 scope:** one command per struct; text-ish field types only (`bool`, `String`, `Option<String>`, `Vec<String>`, counting integers). Subcommands and typed value conversion are explicitly deferred in **PLAN.md**.
> 
> **Conformance** gains **`conformance/tests/derive.rs`** end-to-end tests (parsing, spec round-trip, markdown/manpage/help). Release tooling includes **`usage-derive`** in publish and git-cliff paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 973a60aed5ffb7395de5e43bc90eec71b8b40ba9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added `usage-derive`, enabling CLI definitions through a `#[derive(Cli)]` macro.
  * Supports flags, positional arguments, defaults, aliases, repeatable values, negation, help text, environment settings, and generated command specifications.
  * Added parsing, help/documentation rendering, and KDL serialization for derived CLI types.
  * Added compile-time diagnostics for unsupported or invalid declarations.

* **Documentation**
  * Documented supported attributes, value types, limitations, and the current roadmap.

* **Tests**
  * Added comprehensive end-to-end coverage for parsing, help output, specifications, errors, and CLI configurations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2018,7 +2018,18 @@ dependencies = [
  "serde",
  "serde_json",
  "usage-argv",
+ "usage-derive",
  "usage-lib",
+]
+
+[[package]]
+name = "usage-derive"
+version = "5.1.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
+ "usage-argv",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
     "argv",
+    "derive",
     "clap_usage",
     "cli",
     "conformance",
@@ -19,6 +20,7 @@ license = "MIT"
 clap_usage = { path = "./clap_usage", version = "5.0.0" }
 usage-cli = { path = "./cli" }
 usage-argv = { path = "./argv", version = "5.1.0" }
+usage-derive = { path = "./derive" }
 usage-lib = { path = "./lib", version = "5.1.0", features = ["clap"] }
 
 [workspace.metadata.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ license = "MIT"
 clap_usage = { path = "./clap_usage", version = "5.0.0" }
 usage-cli = { path = "./cli" }
 usage-argv = { path = "./argv", version = "5.1.0" }
-usage-derive = { path = "./derive" }
+usage-derive = { path = "./derive", version = "5.1.0" }
 usage-lib = { path = "./lib", version = "5.1.0", features = ["clap"] }
 
 [workspace.metadata.release]
--- a/PLAN.md
+++ b/PLAN.md
@@ -107,9 +107,17 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       no dependencies. Verified by parsing the output back with usage-lib and
       checking the resulting spec field by field, then rendering it through the
       markdown and manpage generators an adopter's docs build actually uses.
-- [ ] **`usage-derive` v0** — flags, positionals, subcommands, doc-comment help
-      (first paragraph short, whole block long), spec emission. Usage-native
-      attributes mirroring the KDL vocabulary rather than a clap dialect.
+- [x] **`usage-derive` v0** — flags, positionals, doc-comment help (first
+      paragraph short, whole block long), and spec emission, from usage-native
+      attributes. Unsupported field types are a compile error rather than a
+      surprise, and the messages point at the offending field.
+- [ ] **Subcommands in the derive** — one command per struct today. Needs an enum
+      of variants, cross-type table references, and a nested command path through
+      the parse function.
+- [ ] **Typed values** — fields are text today (`String`, `Option<String>`,
+      `Vec<String>`, `bool`, counting integers). Parsing into other types needs an
+      error type for value conversion, which is also where `env`, required-ness,
+      and `choices` get enforced — see the post-binding layer below.
 - [ ] **`usage-derive` v1** — everything mise needs: constraints
       (`requires`/`conflicts`/`overrides`/`required_unless`), `var`, `count`,
       `env`, defaults, delimiters, the `double_dash` modes, global flags, flatten,
--- a/PLAN.md
+++ b/PLAN.md
@@ -89,8 +89,9 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       allocations on success _and_ failure, proved by a counting allocator rather
       than asserted. Answers the binding vectors, including every one where
       usage-lib diverges from the grammar.
-- [x] **Released on the shared version** (#798) — an ordinary member of the
-      workspace rather than a 0.0.0 curiosity outside the release cycle.
+- [x] **Released on the shared version** (#798, #803) — `usage-argv` and
+      `usage-derive` are ordinary members of the workspace at the shared version
+      rather than 0.0.0 curiosities outside the release cycle.
 - [x] **Repeatable vs. variadic flags, and `double_dash_seen`** (#799) — a
       repeatable flag was greedy enough to eat a positional, and `automatic` mode
       reported a separator nobody typed.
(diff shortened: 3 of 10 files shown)


jdx:
All five findings fixed. Every one was real, and the first is the interesting one.

| finding | fix |
| --- | --- |
| `var` enabled greedy collection | `var` and `variadic` are now separate, spelled as a spec spells them; a `Vec` flag is repeatable without saying so |
| negation aliases bypassed collision checks | a negation is another long form, so it collides like one |
| generated impl dropped struct generics | refused, with a reason — the tables are `static` and fields must be concrete |
| non-ASCII shorts truncated | refused: a short is matched as one byte, so `-é` could never be recognized |
| `u8` count overflowed | saturates, with the accumulator typed as the field rather than inferred |

**The first one deserves a note**: that is the same repeatable-vs-greedy conflation #799 fixed in the conformance harness, recurring in the derive's attribute vocabulary a day later. Twice in one week is a design smell, not bad luck — the spec spells greedy as a variadic *argument* (`--include <pattern>...`) and repeatable as `var` on the flag, and any layer that collapses them will eat a positional. The derive now uses both words, and three tests pin the difference.

Worth knowing that my original test missed it by always writing `--include` twice, which is exactly the case where the two behave identically.

17 tests, and the three new compile errors were checked by hand for message and span.

jdx:
Two more findings from the latest review, both fixed.

**`long = "--no-color"` was unreachable.** Stored with its dashes, while a token has its `--` taken off before matching — so nothing could ever reach it, and the duplicate check could not see that it collided with a plain `no-color`. Explicit long names are now stripped the way negations already were, and a test covers the round trip: the spec records `color` on the long form and `--no-color` on the negation, which is how a spec spells each.

**Some shorts can never be delivered.** `-5` reads as a negative number so that `--offset -1` works, `--` ends flag parsing, and `=` separates a short from its value in `-j=8`. Each is refused with that reason instead of compiling into a flag nobody can type.

Both are the same shape of bug as the earlier ones: the derive accepted a declaration the grammar cannot honor. Worth noting that is the failure mode to look for in the rest of this crate — the grammar is written down, so the derive's job is to refuse anything it does not permit, and every finding so far has been a place where it did not.

18 tests. On the check status: Greptile reports `failure` because the review scored 3/5 against this repo's 4/5 threshold, not because a finding is outstanding — its summary says as much.

jdx:
Three more findings fixed, and two of them are the same distinction biting a third time.

| finding | fix |
| --- | --- |
| `variadic` on a non-`Vec` field kept only the last value | has to be a `Vec` — the only field that can hold what the flag collects |
| every `Vec` flag marked repeatable, including `variadic`-only ones | `variadic` no longer implies `var`; emitting both claimed two ways of collecting at once |
| a bare `short` took the field's first letter at read time | resolved after all attributes are read, as a bare `long` already was |

The middle one is worth dwelling on. That is now **three** places the repeatable-vs-variadic distinction has gone wrong: the conformance harness (#799), the derive's attribute vocabulary, and now its spec emission. Each time the symptom differed — a stolen positional, then a flag that claims both `var=#true` and a `...` argument — and each time the cause was a layer treating "collects several values" as one idea.

Two tests now assert the emitted spec for both shapes side by side (`--include` repeatable with a single-value argument, `--exclude` with a variadic argument and no `var`), which is the check that would have caught all three.

The bare-short fix needed a test where the field name and the renamed name start with different letters — `verbose` renamed to `quiet` — since anything else passes by coincidence.

21 tests. Also rebased on the updated #802 and #801.

jdx:
One new finding fixed; the other three in that batch are re-posts of what the previous two commits already fixed — the bots re-review the cumulative diff, so those threads read worse than the code is. Verified in the current tree: `variadic` requires a `Vec`, `variadic` no longer implies `var`, and a bare `short` resolves after all attributes are read.

**New:** a short of `' '` or `'\t'` was written straight into the flag's forms string, which a spec spells space-delimited — so the declaration could not be read back. Refused with the reason.

That is the fifth character class a short cannot be, and they now sit together in one place: non-ASCII, digits, `-`, `=`, and whitespace or control characters. Each one is the same underlying shape — **the derive accepting a declaration the grammar or the spec format cannot carry** — which is the thing to look for in the rest of this crate. Worth saying plainly: seven of the eleven findings on this PR have been that, so the derive's job is less "translate attributes" than "refuse what cannot round-trip", and I'd treat any new attribute as guilty until it has a rejection test.

PR #806, opened 2026-08-11
feat(spec): allow a mount at the top level

jdx:
You asked for this in passing — it makes sense that users should be able to mount at the top level — and it turned out not to be a big issue.

## What it does

`mount` was accepted only inside a `cmd` block, so a CLI whose *own* commands are discovered rather than declared could not say so:

```kdl
name "mycli"
bin "mycli"
cmd "install"
mount run="mycli plugin-commands"
```

The root is a command like any other, and now parses, serializes, and resolves a mount like one.

## Resolved lazily, which matters here

A subcommand's mounts run when the parser descends into it. Nothing ever descends into the root, so an eager root mount would spawn a process on **every invocation** — including `mycli install`, where the answer is already known.

So the root's mount runs only when a word matches nothing declared. Declaring the commands you know about keeps the common path free, and only `mycli something-from-a-plugin` pays for discovery.

The test for that is my favourite part: it mounts `exit 1` and then parses a declared command successfully. If resolution were eager, the parse would fail — no filesystem, no timing, no subprocess to observe.

## A latent bug this would have tripped over

`find_subcommand` memoizes its lookup into a `OnceLock`, and `merge` adds subcommands **without clearing it**. So a name that arrived by mounting was invisible if anything had already looked one up on that command.

That was latent for subcommand mounts too — they work today only because mounting happens to precede the first lookup on the command being mounted into. Fixed in `merge`, where the subcommands actually change, so it cannot depend on call order again.

Three tests: a root mount discovering a subcommand, the laziness proof, and a write-and-read-back round trip.

---

> [!NOTE]
> **Medium Risk**
> Changes core argument parsing and when mount subprocesses run; behavior is covered by new tests but affects every invocation path that hits an unknown subcommand word at the root.
> 
> **Overview**
> Top-level **`mount`** is now valid in usage specs (parse, round-trip serialize, and docs), so CLIs can declare static commands and discover the rest—e.g. plugin subcommands via `mount run="mycli plugin-commands"`.
> 
> **Lazy vs eager discovery:** Full parsing uses **`MountTiming::WhenAWordIsUnknown`**, so the root mount runs only when the next token is a non-flag word that is not already a declared subcommand—avoiding a discovery subprocess on every invocation (`mycli install`, `mycli --help`). **`parse_partial`** stays **`Eager`** so completions like `mycli <tab>` see mounted commands with nothing typed yet.
> 
> **Bug fix:** **`SpecCommand::merge`** now resets **`subcommand_lookup`** so subcommands added by mounting remain findable after any prior lookup memoization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9eb998b370315553bcf57675c4295a3640ff9859. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added support for top-level command mounts, including plugin discovery and lazy execution.
  - Root-mounted commands and flags are discovered when needed and preserved in formatted specifications.
  - Added configurable mount precedence for overriding default commands.
  - Improved completion by resolving mounts early when necessary.

- **Bug Fixes**
  - Fixed command and alias lookups after merging additional subcommands.
  - Prevented duplicate or unnecessary mount discovery when commands, flags, or defaults already match.

### DIFF
--- a/docs/spec/reference/cmd.md
+++ b/docs/spec/reference/cmd.md
@@ -83,6 +83,23 @@ task commands as if they were statically defined in the usage spec.
 a shebang script therefore needs a POSIX shell to be available; one that invokes a program
 directly, like `mycli mount-usage-tasks` above, works either way.
 
+### Mounting at the top level
+
+A `mount` also works as a top-level node, for a CLI whose _own_ commands are
+discovered rather than declared — one whose subcommands come from plugins, say:
+
+```kdl
+name "mycli"
+bin "mycli"
+cmd "install"
+mount run="mycli plugin-commands"
+```
+
+The root's mount runs only when a word matches nothing already declared, so
+`mycli install` costs nothing extra and only `mycli something-from-a-plugin` pays
+for discovery. Declaring the commands you know about therefore keeps the common
+path free.
+
 ### Global flags and mounted commands
 
 A mounted command describes a different program, so the flags of the commands it is mounted under
--- a/docs/spec/reference/cmd.md
+++ b/docs/spec/reference/cmd.md
@@ -95,10 +95,10 @@ cmd "install"
 mount run="mycli plugin-commands"
 ```
 
-The root's mount runs only when a word matches nothing already declared, so
+The root's mount runs only when a _word_ matches nothing already declared, so
 `mycli install` costs nothing extra and only `mycli something-from-a-plugin` pays
-for discovery. Declaring the commands you know about therefore keeps the common
-path free.
+for discovery. Flags never trigger it — `mycli --help` does not run your mount
+command — so declaring the commands you know about keeps the common path free.
 
 ### Global flags and mounted commands
 
--- a/lib/src/parse.rs
+++ b/lib/src/parse.rs
@@ -635,8 +635,15 @@ fn parse_partial_with_env(
         // does not spawn a process for every invocation, and a flag — `--help`, or
         // anything unrecognized — never triggers discovery at all, which it would
         // otherwise do simply by not being a subcommand.
+        // A declared `default_subcommand` already says what an unmatched word means,
+        // and it costs nothing — so discovery waits behind it unless a mount asks to
+        // outrank it. Without this, a task runner would spawn its discovery process
+        // once per task invocation.
+        let default_catches_it = spec.default_subcommand.is_some()
+            && !out.cmd.mounts.iter().any(|m| m.overrides_default);
         if !mounts_resolved
             && !out.cmd.mounts.is_empty()
+            && !default_catches_it
             && !input[idx].starts_with('-')
             && out.cmd.find_subcommand(&input[idx]).is_none()
         {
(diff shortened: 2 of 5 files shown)


jdx:
Both findings fixed — and the Major one was a good catch that my own docs contradicted.

**A flag was triggering discovery.** Lazy resolution keyed off "this word matches no subcommand", and `--help` matches no subcommand either — so it spawned the mount command. That is a side effect nobody asked for, and the opposite of what the page I wrote in this PR promises. Only a word that could name a command triggers discovery now, with a test that mounts `exit 1` and parses `ex --verbose` successfully.

**The default-subcommand path** descended without recording that it had already run the new command's mounts, so a later unmatched word could run them twice.

Both are the same underlying mistake on my part: I keyed the decision on *what didn't match* rather than on *what the token is*.

jdx:
Another round, and the best finding yet: **my laziness broke the case this feature exists for.**

A completion asks about `mycli <tab>` — no word typed, so nothing to trigger discovery with, so the mounted commands were never offered. The whole point of a root mount is that completions can see discovered commands, and lazy resolution silently defeated it.

Timing now depends on who is asking, which is the same split as the missing-value check in #807:

| caller | when the root mount runs | why |
| --- | --- | --- |
| `parse_partial` (completions) | up front | has to offer every command, and has no word to go on |
| `parse` (execution) | when a word matches nothing declared | knows the word it was given, so it only pays when the answer might come from there |

Flags never trigger it either way. Test: `parse_partial(&spec, ["ex"])` must see the mounted command with nothing typed.

## One ordering question I'd rather you decide

A spec with **both** a root mount and a `default_subcommand` now tries discovery before falling back to the default. So for a mise-shaped CLI (`default_subcommand = "run"`), every task name would spawn the mount command before reaching `run`.

Nothing has both today, so it is theoretical — but the two orderings have real consequences:

- **discovery first** (current): a discovered command can shadow the default subcommand, at the cost of a subprocess per unmatched word.
- **default first**: the declared fallback always wins and never pays for discovery, but a mounted command is unreachable in any spec that declares a default.

I lean toward default-first, since a `default_subcommand` is an explicit statement about what unmatched words mean — but it makes root mounts and default subcommands mutually exclusive in practice, which is worth choosing deliberately rather than by accident of code order.

jdx:
Made the ordering configurable, defaulting the way you leaned.

You are right that mounting has a real cost — it runs a process. So a `default_subcommand` now **outranks discovery**, because it already says what an unmatched word means and costs nothing to consult. Without that, a CLI shaped like a task runner would spawn its discovery process once per task invocation.

A mount that should shadow the fallback opts in:

```kdl
mount run="mycli plugin-commands" overrides_default=#true
```

Two tests, one per direction, both proving it by side effect rather than by timing: the default-wins test mounts `exit 1` and parses successfully, and the opt-in test mounts a real `echo` and checks the discovered command wins while an unknown word still reaches the default.

So the full timing story for a root mount is now:

| asking | when discovery runs |
| --- | --- |
| completion | up front — has to offer every command, and has no word to go on |
| parse, no `default_subcommand` | when a word matches nothing declared |
| parse, with `default_subcommand` | never, unless the mount says `overrides_default` |
| any flag token | never |

jdx:
Two more findings fixed, one of which was a real UX bug I had not seen.

**Completion and execution disagreed.** With a `default_subcommand` and a root mount that does not say `overrides_default`, a parse skipped discovery while a completion still ran it — so the shell would offer a discovered command, and running it would hand the word to the default subcommand instead. Offering what execution will not route is worse than offering nothing, so the gate now applies to both paths.

That settles what the setting means, which is cleaner than what I had: **a root mount under a default subcommand contributes nothing anywhere unless it asks to outrank it.** One test checks both halves at once by mounting `exit 1` — the completion must not offer the command, and the parse must reach the default.

**I also had to correct this PR's own documentation.** Rendering help goes through `parse_partial`, so `--help` *does* resolve a root mount — and should, since help lists every command. My claim that "flags never trigger it" was true of a parse and false of help. The page now says which is which:

| asking | when discovery runs |
| --- | --- |
| completion, or rendering help | up front — both need the whole command list |
| parse, no `default_subcommand` | when a word matches nothing declared |
| parse or completion, with `default_subcommand` | never, unless the mount says `overrides_default` |
| an unrecognized flag during a parse | never |

The remaining threads above are re-posts from before these commits — the bots re-review the cumulative diff, so line numbers shift and fixed findings reappear.

PR #807, opened 2026-08-11
fix(parse): stop losing a flag that is missing its value

jdx:
First of the parser fixes, taking you at your word that these are bugs that went unnoticed because the parser is mostly used for completions.

Three divergences the corpus recorded, all of them the parser quietly discarding something:

**`ex --jobs` parsed fine and bound nothing.** A forgotten value looked like a working command, and `ex --jobs --force` bound `force` while dropping `jobs` silently. Now an error.

Worth noting where the check went: `parse`, not `parse_partial`. A half-typed `--jobs ` is exactly what a completion is asked about, so the pending flag stays in the partial output and only a full parse — which has nothing left to wait for — treats it as a failure. That distinction is why `ParseOutput::flag_awaiting_value` exists, so the fix is two lines and a comment rather than a new code path.

**`-j=8` bound `=8`.** The value was taken as everything after the letter. One `=` between letter and value is a separator, matching the long form; `-j==8` still means `=8`.

**`--jobs=` bound nothing**, collapsing "given empty" into "not given". The two are now told apart by whether an `=` was written rather than by whether anything followed it.

## The corpus did its job

Fixing these made the suite report **five labels as stale** — "you claimed usage-lib diverges here, it now agrees, delete the label" — which is the workflow the corpus was built for. Deleting them takes the recorded divergences from 16 to 11.

One test needed rethinking rather than updating: it asserted usage-argv resolves "at least a dozen" divergences, which was a floor that shrinks every time usage-lib gets fixed. An empty set would mean the reference caught up entirely, which is the goal, so the floor is gone and the per-vector assertion carries the weight.

## Two I did not touch, because they need your call

**Unrecognized flags fall through to positionals** (`ex --wat` binds `--wat` to an argument). This is the root of most remaining divergences, but mise parses *task* arguments with this parser at run time — `usage::Parser::new(&spec).parse(&args)` in `src/task/mod.rs` — so rejecting an undeclared flag changes what a task accepts, not just what a completion offers. That could break tasks that pass arbitrary flags through today.

**A repeated `--` is eaten.** The grammar says only the first is a separator. But `test_double_dashes_without_preserve` asserts the current behavior explicitly, and `double_dash="preserve"` exists precisely to keep separators — which makes "the default consumes them all" look like a deliberate design rather than an oversight. I reverted my fix for this one when that test failed. If your reading is that the spec's design wins, the corpus vector changes instead of the parser.

Both are marked **Needs a decision** in `PLAN.md`.

---

> [!NOTE]
> **Medium Risk**
> Changes real argv binding and errors for full parses (including mise task argument parsing), though completions stay on parse_partial; behavior is intentional and corpus-backed.
> 
> **Overview**
> Aligns **usage-lib** parsing with the argv grammar on three flag-binding bugs that previously showed up only in the conformance corpus.
> 
> **Missing flag values** now fail in full **`parse`** when a value-taking flag ends without a value (e.g. `ex --jobs`, `ex --jobs --force`), using **`flag_awaiting_value`** after partial parsing. **`parse_partial`** is unchanged so half-typed flags still work for completions.
> 
> **Attached `=` semantics** are fixed: long **`--jobs=`** binds an empty string (distinct from omitting the flag); boolean **`--force=`** no longer leaves a stray empty positional; short **`-j=8`** binds **`8`** by treating a single leading **`=`** as a separator (long-form parity).
> 
> The corpus drops stale **usage-lib** divergence labels on the affected vectors, adds coverage for **`--force=`**, bumps in-scope vector count, and relaxes the conformance test that required resolving “at least a dozen” divergences. **PLAN.md** and **docs/spec/argv.md** record remaining open decisions (unknown flags → positionals, repeated **`--`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22703a2971b236f8d780147e8c32815f7f083a0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -208,13 +208,18 @@ how a fix gets verified — including telling you to delete the label afterwards
 
 - [ ] Unrecognized flags fall through to positionals, so `ex --wat` binds `--wat`
       to an argument, or reports `unexpected_arg` when there is none. This is the
-      root of most of the recorded divergences.
-- [ ] A flag missing its value is dropped silently: `ex --jobs` parses fine and
-      binds nothing, so a forgotten value looks like a working command.
-- [ ] `=` is kept in attached short values, so `-j=8` binds `=8`.
+      root of most of the recorded divergences. **Needs a decision**: mise parses
+      task arguments with this parser at run time, so rejecting an undeclared flag
+      would change what a task accepts, not just what a completion offers.
+- [x] A flag missing its value is dropped silently — now an error, in `parse` but
+      not `parse_partial`, since a half-typed flag is exactly what a completion is
+      asked about.
+- [x] `=` is kept in attached short values, so `-j=8` binds `=8`.
 - [ ] A repeated `--` is eaten, so a forwarded command line containing its own
-      separator is altered in transit.
-- [ ] `--jobs=` binds nothing rather than the empty string.
+      separator is altered in transit. **Needs a decision**: an existing test
+      asserts this, and `double_dash="preserve"` is the declared way to keep
+      separators, which may make it intentional.
+- [x] `--jobs=` binds nothing rather than the empty string.
 - [ ] A flag with a variadic argument rejects its second value, though
       [the flag reference](https://usage.jdx.dev/spec/reference/flag) documents the
       form.
--- a/conformance/tests/argv.rs
+++ b/conformance/tests/argv.rs
@@ -77,11 +77,11 @@ fn agrees_where_usage_lib_does_not() {
         resolved += 1;
     }
 
-    assert!(
-        resolved >= 12,
-        "expected usage-argv to resolve at least a dozen recorded divergences, \
-         resolved {resolved}"
-    );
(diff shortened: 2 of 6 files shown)


jdx:
Fixed. Telling `--jobs=` apart from `--jobs` by whether an `=` was written also started queueing the empty text after `--force=` — a flag that holds no value and has nothing to do with it — which left a stray empty word to be read as a positional.

An empty value now reaches only a flag that takes one, with a corpus vector pinning `--force=` on a boolean. A *non-empty* value on such a flag (`--force=8`) behaves as it did before, which is a separate question the corpus does not answer yet — arguably it should be an error, but that is a behavior change beyond this PR.

jdx:
Both remaining threads here are re-posts of the empty-suffix finding, which is fixed in `0f2e0ff`: an empty value now reaches only a flag that takes one, so `--force=` no longer queues a stray empty word for the positionals. A corpus vector pins it.

Green and ready.

PR #810, opened 2026-08-11
feat(spec): make unknown flags configurable, and keep them as values

jdx:
Your call, implemented. An unrecognized flag-like token stays what it has always been here — a word, offered to the positionals — and that is now a documented decision rather than an accident. The grammar said one thing and the parser did another, which was the only indefensible state.

## Why the outlier position is the right one

Every comparable parser refuses the token, and they are right for what they do: parse **their own** argv, where a dash-word is a flag or a typo. A usage spec also parses command lines whose flags it does not own:

- a shell script run through `usage exec`, forwarding options to a tool it wraps
- a task's arguments, where the task script is the authority on what it accepts
- a completion, asked about a line that is still half-typed

In those, an unrecognized token is data in transit far more often than a mistake, and refusing it breaks the wrapper for anyone who did not enumerate the flags of the program behind it. The grammar page now says this, with the cost stated plainly: a misspelled `--hekp` becomes an argument, and whether it does depends on whether a positional happens to be free.

## Configurable at both levels

```kdl
unknown_flags "error"               // for the whole CLI
cmd "exec" unknown_flags="value"    // except here, which forwards a command line
```

Nearest command that states a preference wins, then the spec, then `value`. Unlike `effect` this **is** inherited — it describes how a command line is read rather than what a command does, and a CLI that forwards options tends to forward them everywhere.

From Rust: `#[usage(unknown_flags = "error")]`, which is the case that usually wants it — a compiled binary generally owns every flag it accepts.

## The carve-out oclif had to learn

Even when refusing, a lone `-` and a negative number stay values. [oclif/core#600](https://github.com/oclif/core/issues/600) is that exact regression: they switched to "anything starting with a hyphen is a flag", broke negative-number arguments, and had to special-case them back. The check here skips non-flag-like tokens before it ever consults the setting, and a test pins `-1` under `unknown_flags "error"`.

One subtlety worth flagging: the rejection had to go **inside** the flag branches, where the lookup has just failed and nothing from the token has been applied. Placed downstream, `-az` set `-a` before discovering that `z` names nothing — a partly-applied bundle from a rejected token.

## Scope

- `lib`: the setting on `Spec` and `SpecCommand`, KDL parse and emit, inherited resolution, refusal in both flag branches.
- `argv`: lenient by default, with the effective mode stored per command — inheritance is resolved by whoever builds the tables, keeping it out of the parse.
- `derive`: the struct attribute, emitted into both the tables and the spec.
- corpus: six vectors move from expecting an error to expecting a value; five new ones cover strict mode, the per-command override, whole-bundle rejection, and the negative number.
- docs: `/spec/argv#unrecognized-flags` explains the reasoning; `/spec/reference/cmd` documents the knob.

This closes the last of the big recorded divergences. 91 vectors, 10 remaining — and those are now narrow.

---

> [!NOTE]
> **Medium Risk**
> Changes default argv parsing for unrecognized flags (reject → treat as values) across both parsers, which can alter existing CLI behavior. Mitigated by an explicit strict mode and broad corpus coverage.
> 
> **Overview**
> **Unrecognized flag-like tokens are now values by default**, offered to positionals like any other word. Specs that wrap or forward argv (scripts, tasks, completions) no longer reject tokens they do not own. CLIs that want typo detection opt in with `unknown_flags "error"`.
> 
> The setting lives on both the root spec and individual commands, **inherits** (nearest enclosing preference wins), and is overridable per command — e.g. strict overall except `cmd "exec" unknown_flags="value"`. Derive exposes the same knob via `#[usage(unknown_flags = "error")]`.
> 
> Also tightens number detection so `-1e5` / `-1.5e-3` stay values while `-inf` and `-1x` remain flag-shaped, and ensures unknown short bundles like `-az` apply **none** of their letters before falling through or erroring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dd75908c41c18ffe2aaa1b60f912849297e8457. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for unrecognized flags, with lenient value handling by default and strict rejection as an option.
  * Supports command-specific overrides, inherited settings, and specification serialization.
  * Unknown short-flag bundles are validated as a whole.

* **Bug Fixes**
  * Preserved negative numbers, numeric exponent forms, and lone hyphens as positional values.
  * Prevented partial application of short-flag bundles when validation fails.

* **Documentation**
  * Documented unknown-flag behavior, inheritance, precedence, and parsing rules.


jdx:
Both fixed, and the first one is the best catch on this PR.

**The two parsers disagreed about `-1e5`.** usage-lib asked `f64::from_str`, which takes an exponent; usage-argv used a hand-written byte scanner that stopped at `e` and called the token a flag. So `--offset -1e5` was a value under one parser and a refused unknown flag under the other — precisely the drift the corpus exists to prevent, in the one place the corpus had no vector.

One rule now, spelled out identically on both sides: **digits, at most one `.`, and an optional signed exponent.** Narrower than a float parse deliberately — `f64` also accepts `inf` and `NaN`, and `-inf` is far likelier to be a misspelled flag than a number somebody meant to pass. Four vectors pin the edges so they cannot drift again:

| token | reading |
| --- | --- |
| `-1e5`, `-1.5e-3` | value |
| `-inf`, `-1e`, `-1x` | flag-shaped, names nothing |

The grammar page states the rule rather than leaving "a negative number" to interpretation.

**`unknown_flags` was not emitted for subcommands.** The tables carry the effective value per command, so repeating an inherited answer says nothing — but a command that *differs* has to say so, or the setting never reaches the spec. Now written only where it changes, with a test asserting both halves: a matching subcommand stays quiet, a differing one declares it.

One note on CI: the remaining `IN_SCOPE` bump in this PR disappears when #813 lands, which replaces that count with a snapshot of the exempt vectors. #813 also unbreaks `main`, so it is worth taking first.

jdx:
Two fixes, and the first is embarrassing in a useful way.

**The property was written twice.** An overriding subcommand emitted `unknown_flags="value" unknown_flags="value"` — a KDL node carrying the same property twice, which keeps only the last. That is *exactly* the bug this PR fixes for `overrides` and `required_unless`, reintroduced two commits later when I relocated the block and left a copy behind.

The test did not catch it because it asked whether the output **contained** `unknown_flags="value"`, and a doubled write contains that too. It now finds the line and counts occurrences — and I put the duplicate back to watch it fail before removing it again:

```
assertion `left == right` failed: a differing subcommand declares it exactly once:
cmd "exec" unknown_flags="value" unknown_flags="value" {
```

Worth naming the pattern, since it is the third time in this stack: `contains` is a weak assertion for generated output, because the wrong output usually contains the right output.

**The numeric grammar is now stated exactly** — digits, at most one `.`, then optionally `e`/`E`, an optional sign, and at least one digit — rather than leaving "an optional exponent" to interpretation. Plus the American spelling the locale check asked for.

The `IN_SCOPE` thread above is stale: it reviewed this branch before the rebase, and recommends #813, which is merged. That constant no longer exists here.

PR #813, opened 2026-08-11
fix(test): snapshot the exempt vectors instead of counting them

jdx:
**`main` is red**, and it is a tripwire I wrote that did it. #806 and #809 each added one ordinary corpus vector. Each was green alone — each saw only its own increment against a hardcoded count — and together they made 65 where the constant said 64.

## Counting was the mistake, not the number

Bumping it to 65 would have fixed today and guaranteed tomorrow. A count asserts something **nobody can verify by reading it**, and it collides whenever two changes touch the data it counts.

What the assertion was for — noticing when a vector changes sides — is better served by a snapshot of *which* vectors are exempt and why:

```
arg-required-missing: required args are checked after binding
arg-variadic-max: var_min and var_max are checked after binding
argv-beats-env: env fallback is applied after binding
choice-invalid: choices are validated after binding
...
```

- Fails as a **reviewable diff naming the vector that moved**, instead of a number to be guessed at.
- Adding an ordinary vector **does not touch it at all**, so concurrent PRs cannot collide.
- The list documents the post-binding boundary that the count obscured — which was the interesting information all along.

## The file had two magic numbers, not one

`error_expectations_are_reachable` asserted at least six error classes were exercised. Same problem, same fix: deleted, since the per-vector checks already cover it.

While there, `tests/argv.rs` is now only about the parser. `no_vector_has_an_unloadable_spec` was a duplicate of `specs_are_valid` in `reference.rs`, and `out_of_scope_vectors_say_why` is redundant now that the reasons are in the snapshot. Corpus well-formedness is checked in one place. 135 lines to 92, and no numbers.

Worth merging ahead of my other open PRs, since it unbreaks `main`.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to conformance harness; no production parser or runtime behavior.
> 
> **Overview**
> Fixes flaky **`main`** caused by a hardcoded **`IN_SCOPE`** count on binding vectors: concurrent corpus PRs each bumped the number locally and collided when merged.
> 
> **`the_exempt_vectors`** now **`insta`**-snapshots the sorted list of out-of-scope vector IDs with their skip reasons, so failures show *which* vector moved rather than a magic number. Ordinary new in-scope vectors no longer touch that snapshot.
> 
> **`argv.rs`** is trimmed to parser-focused checks: removed the in-scope count, the “≥6 error classes” floor, duplicate **`no_vector_has_an_unloadable_spec`**, and **`out_of_scope_vectors_say_why`** (reasons live in the snapshot). Module docs point exempt tracking at the new test and corpus validation at **`reference.rs`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3acd140bc8768f4a7d6067781051bb9e856398ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/conformance/tests/argv.rs
+++ b/conformance/tests/argv.rs
@@ -1,19 +1,18 @@
 //! Checks usage-argv against the corpus.
 //!
-//! This is the point of writing the grammar down. usage-argv is expected to match
-//! every binding vector — including the fifteen where usage-lib does not, since
-//! those record the grammar's intent rather than the reference's behavior.
+//! This is the point of writing the grammar down: usage-argv is expected to match
+//! every vector it answers, including the ones where usage-lib does not, since those
+//! record the grammar's intent rather than the reference's behavior.
 //!
-//! Vectors that turn on something decided after the last token (`required`,
-//! `choices`, `env`, defaults, `var_min`/`var_max`, `overrides`) are out of scope
-//! for this crate by design. Their count is asserted below so the exempt set
-//! cannot quietly grow.
+//! Vectors that turn on something decided after the last token — `required`,
+//! `choices`, `env`, defaults, `var_min`/`var_max`, `overrides` — are out of scope
+//! for this crate by design, and [`the_exempt_vectors`] is the record of which.
+//!
+//! Corpus well-formedness is checked once, in `reference.rs`, rather than again
+//! here.
 
 use usage_conformance::argv::{run, Outcome};
-use usage_conformance::{load, Expect, Reference, Vector};
-
-/// Vectors that exercise binding, which is what usage-argv implements.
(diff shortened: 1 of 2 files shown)


jdx:
The thread above reviewed an earlier push. The count is gone entirely — `exempt.len()` was replaced by an insta snapshot of which vectors are exempt and why, so identity *is* what is guarded now:

```
arg-required-missing: required args are checked after binding
argv-beats-env: env fallback is applied after binding
choice-invalid: choices are validated after binding
...
```

A vector changing sides fails as a diff naming it, rather than as a number off by one.

PR #814, opened 2026-08-11
test(parse): let each vector declare which layer answers it

jdx:
Top of the stack, and the follow-up to a question you asked: are the exempt vectors a to-do list or expected? Expected — but my exemption rule was inflating them.

## The problem

The harness decided whether usage-argv should answer a vector by looking for a post-binding feature **anywhere in its spec**. That is coarser than the principle:

| vector | what it asserts | why it was exempt |
| --- | --- | --- |
| `long-negation` | `--no-color` → `color=false` | the flag also declares `default=#true` |
| `choice-valid` | `--shell zsh` → `shell="zsh"` | the flag declares `choices` |
| `argv-beats-env` | `--jobs 8` → `jobs="8"` | the flag declares `env` |
| `required-flag-present` | `--file a.txt` → `file="a.txt"` | the flag declares `required` |

`--no-color` binds false whatever else that flag declares. `long-negation` is the embarrassing one: negation is a **binding** rule usage-argv implements, and it was not being tested on it.

## The fix

Each vector says which layer it is a question for — `binding` (the default) or `post-binding`. usage-argv now answers **81 of 99** rather than 75, and the 18 it skips are the ones that genuinely need to know a value's type.

Declaring it also serves what the corpus is *for*: an implementation in another language should be **told** which vectors apply to the layer it implements, rather than inferring it from KDL it may not fully parse yet. That is a fact about the vector, so it belongs in the vector.

## Two things removed

The snapshot of exempt vectors goes away. It existed to notice a vector silently leaving usage-argv's coverage, and that is now a one-line diff in the corpus itself — the same information, at the source.

The grammar page loses its vector counts ("answers 63 of the 87"), which had already gone stale twice. Same anti-pattern as the constant: a number in prose that nobody can check and everybody has to update.

---

> [!NOTE]
> **Low Risk**
> Test-harness and corpus metadata only; no production parser or runtime behavior changes.
> 
> **Overview**
> Corpus vectors now declare which parser layer they exercise — **`binding`** (default) or **`post-binding`** — instead of the harness inferring that from post-binding features anywhere in the spec.
> 
> That inference was too coarse: cases like `long-negation`, `choice-valid`, and `argv-beats-env` were skipped even though they only assert ordinary binding. usage-argv now answers those, and only vectors that truly need defaults, env, choices, required checks, overrides, or var bounds are marked out of scope.
> 
> The exempt-vector snapshot and stale coverage counts in the docs are removed; which vectors are post-binding is now a one-line fact in the corpus itself.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccc784eda4f14d4b67c9844a202ab07cf7665c59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Good catch — a `.snap.new` for a test I had just deleted was committed alongside the deletion.

Removed, and the reason it was possible is now fixed: `.gitignore` ignored `*.pending-snap`, which is what insta writes for *inline* snapshots, but file snapshots get `.snap.new` and only the first spelling was covered. Both are ignored now, so the next pending snapshot cannot be staged by accident.

PR #816, opened 2026-08-11
feat(derive): compile subcommands from an enum

jdx:
First of the next stack, and the largest gap in the derive: it could describe **one** command, so nothing with subcommands could use it — which is every CLI it is meant for.

## Shape

```rust
#[derive(Cli)]
#[usage(bin = "ex")]
struct Ex {
    #[usage(short = 'v', long, global)]
    verbose: bool,
    #[usage(subcommand)]
    command: Option<Commands>,
}

#[derive(Subcommands)]
enum Commands {
    /// Install a tool
    Install(Install),
    /// Run a task
    #[usage(name = "run")]
    RunTask(Run),
}

/// Install a tool
#[derive(Args)]
struct Install {
    #[usage(short = 'f', long)]
    force: bool,
    tools: Vec<String>,
}
```

## Two problems worth explaining, because they shaped the design

**A macro sees one item.** The three derives cannot share a counter or read each other's fields, so keys carry a **hash of the type they came from** in the high half, the index in the low half, and a tag for command/flag/argument. Independently expanded macros therefore cannot hand two fields the same key — which matters because a parse dispatches on it. Two type names would have to collide in 32 bits, and `Spec::to_kdl` now asserts the tree holds no duplicates, so a collision fails a test instead of binding the wrong field. A test checks that invariant directly.

**A `static` cannot call a method.** The parent splices its children's tables into its own `static`, so the traits expose them as **associated consts**, not functions. That keeps the tables static all the way down — nothing is assembled at run time, which was the whole point.

## Restructuring that fell out

Every derived struct now collects into a generated `Partial` rather than into locals, since a subcommand's values cannot live in the root's parse function. That replaced the prefixed-locals machinery from #803 entirely, and made `apply` shared between a root and a subcommand.

A doc comment on a **variant** overrides the struct's, because that is where a reader of the enum expects to describe the command — ignoring it would lose the description silently. Overriding one field of the struct's metadata is possible in a const, so this costs nothing at run time.

## Verified

Eight tests: routing by name and by a renamed variant, per-command fields with defaults, a global flag on either side of the command word, one command not answering to another's flag, an absent subcommand, key uniqueness across separately expanded types, and a snapshot of the emitted spec:

```kdl
name "ex"
bin "ex"
flag "-v --verbose" help="Say more" global=#true
cmd "install" help="Install a tool" {
    flag "-f --force" help="Overwrite an existing install"
    flag "-j --jobs" help="How many at once" default="4" {
        arg "<jobs>"
    }
    arg "[tools]..." help="What to install"
}
cmd "run" help="Run a task" {
    arg "<task>" help="The task to run"
}
```

Nested `cmd` blocks that read like something a person would have written, which is what `usage g markdown|manpage` needs.

## Stated limits

- A `subcommand` field must be `Option<T>`: reporting that one was *required* is a required-ness question, and that layer lands in the next PR of this stack.
- Nesting stops at one level — a subcommand's own subcommands need `Args` to carry a `subcommand` field too.

---

> [!NOTE]
> **Medium Risk**
> Large derive/codegen refactor plus a public key-type change (`u32` → `u64`) that affects parse dispatch. Mitigated by conformance tests and debug duplicate-key checks, but wrong keys would bind the wrong fields.
> 
> **Overview**
> Enables declaring **subcommands** with the derive: a `#[usage(subcommand)]` field holds an enum (`Subcommands`), and each variant wraps an `Args` struct for that command's flags and arguments.
> 
> Independently expanded macros cannot share a counter, so keys widen from `u32` to **`u64`** and carry a type-name hash. New `CommandArgs` / `Subcommands` traits expose associated consts so parents splice children's tables into `static` parse tables with no runtime assembly. `Spec::to_kdl` debug-asserts the tree has no duplicate keys.
> 
> Codegen now accumulates into a generated `Partial` (shared `apply`) instead of parse-local variables, which is what lets a root route events into subcommands it cannot see. Variant doc comments override the wrapped struct's help; nesting and required subcommands remain out of scope for now.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5373abbdc400e8da7dabaa3b86769bc782648466. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for defining and parsing CLI subcommands, including nested structures, command-specific options, global flags, defaults, and optional subcommands.
  * Added `Args` and `Subcommands` derive macros for generating command-line parsers and metadata.
  * Expanded command keys to support larger values and improved uniqueness checks.

* **Bug Fixes**
  * Added validation and diagnostics for duplicate command keys and subcommand names.
  * Improved error reporting for invalid subcommand definitions.

### DIFF
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -105,7 +105,14 @@ pub struct Command<'a> {
     /// [`UnknownFlags`].
     pub unknown_flags: UnknownFlags,
     /// Caller-assigned identifier, echoed back in [`Event::Command`].
-    pub key: u32,
+    ///
+    /// Wide enough for a derive to make these unique without coordination: two
+    /// macro expansions cannot see each other, so the generated keys carry a hash
+    /// of the type they came from in the high half and a per-type index in the low
+    /// half. A parse dispatches on this, so a collision would bind the wrong field
+    /// — [`Spec::to_kdl`](crate::spec::Spec::to_kdl) checks the tree for duplicates
+    /// in debug builds.
+    pub key: u64,
 }
 
 impl Command<'_> {
@@ -126,7 +133,8 @@ impl Command<'_> {
 pub struct Flag<'a> {
     /// Caller-assigned identifier, echoed back in [`Event::Flag`]. This is how
     /// generated code knows which field to assign without any string comparison.
-    pub key: u32,
+    /// See [`Command::key`] on why it is this wide.
+    pub key: u64,
     /// Unused by binding, kept so a table entry can carry its own name for
     /// diagnostics.
     pub name: &'a str,
@@ -175,8 +183,9 @@ impl Flag<'_> {
 /// A positional argument.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Arg<'a> {
-    /// Caller-assigned identifier, echoed back in [`Event::Arg`].
-    pub key: u32,
+    /// Caller-assigned identifier, echoed back in [`Event::Arg`]. See
+    /// [`Command::key`] on why it is this wide.
+    pub key: u64,
     /// Whether this argument keeps taking values once it has one.
     pub var: bool,
     /// This argument's relationship to the `--` separator.
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -26,6 +26,30 @@ use core::fmt::Write as _;
 use crate::UnknownFlags;
 use crate::{Arg, Command, DoubleDash, Flag};
 
+/// The first key that appears twice anywhere in a command tree, if any.
+///
+/// Keys are what a parse dispatches on, and a derive assigns them without being able
+/// to see other expansions — it hashes the type name to keep them apart. That makes
+/// a collision astronomically unlikely rather than impossible, so it is checked
+/// where a CLI is written out, which every adopter does in a test.
+fn duplicate_key(cmd: &Command<'_>) -> Option<u64> {
+    let mut keys = std::vec::Vec::new();
+    collect_keys(cmd, &mut keys);
+    keys.sort_unstable();
+    keys.windows(2)
+        .find(|pair| pair[0] == pair[1])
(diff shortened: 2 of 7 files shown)


jdx:
Four findings fixed, and two of them my tests had passed **by coincidence** — worth spelling out, because it is the same trap twice.

**The variant's name never reached the tables.** `Subcommands` read it, validated it against its siblings, documented it as the command name — and then spliced the wrapped struct's command unchanged, so the *struct's* name selected the command. My test wrote `RunTask(Run)` with `name = "run"`, where the two spellings agree, so it passed while the feature did not work. Each variant now gets its own command and metadata overriding the name, and the new test uses `Install(AddArgs)` named `add`, plus an assertion that the struct's name selects nothing.

**Two variants wrapping one struct cannot work**, and writing the test for it is what showed why: a command's values collect into the struct that declares them, and its fields' keys come from that struct — so two commands sharing one collect into whichever was reached first, and `select` picks between them by a key they share. It is now a compile error naming both sites:

```
error: two variants both wrap `S`, and a command collects into the struct that
       declares it — so give each command its own struct, even if the fields
       are identical
```

**`in_module` prepended `super::` to any path**, turning `crate::cmds::X` into nonsense and quietly redirecting `self::X` and `super::X` to the wrong module. Each prefix now shifts by the right amount, and an absolute path is left alone.

**The duplicate-key check counted one table met twice as a collision.** It now pairs each key with the address it came from: two *different* tables sharing a key is the fault; meeting the same table twice is not.

Plus two smaller ones: "no subcommand given" is an `Option` rather than a reserved key, and `apply` arms evaluate to whether they claimed the event instead of returning — a command with no flags of its own had every arm diverge, which clippy caught as an unreachable tail.

10 tests on this file now. The coverage failure earlier was a runner TLS error in `actions/checkout`, not the code; re-run and green.

PR #817, opened 2026-08-11
feat(derive): check what a parse cannot decide on its own

jdx:
Second of the stack (#816 merged while this was in flight, so it targets `main`). The parser binds tokens and stops there, which left required-ness, `choices`, variadic bounds, and `env` fallback unimplemented — the 18 corpus vectors the harness reports as **post-binding**, and the reason a `subcommand` field had to be `Option<T>`.

## Order matters, so it is stated

1. **The environment** fills what argv left out.
2. **Required-ness** — which the type declares: a `String` has nowhere to put "absent". After the environment, so a flag with `env` set is not reported missing.
3. **`choices` and bounds**, which judge a value however it arrived, including from the environment or a default.

Only the command that actually ran is judged. A flag that `install` requires says nothing about an invocation of `run` — which is why the check hangs off the `CommandArgs` trait and the enum forwards it to the selected variant only. A bare `subcommand` field now means one is required and says so when absent, removing #816's stated limit.

## Two decisions worth arguing with

**A field remembers whether a token supplied it**, rather than inferring that from an empty value. `--out=` is a value somebody typed, and treating `""` as unset would make it unsayable — the same distinction #810 fixed in the parser, carried through to the checks.

**Bounds constrain the values a field was given.** An unused optional flag is absent, not a violation. Otherwise `var_min = 1` becomes a second spelling of required-ness, and there is no way left to say "at least two, *if* you use it at all". My own test caught this: I put `var_min = 1` on an optional flag and every unrelated test in the file started failing with `VarTooFew`.

## Contradictions are compile errors

```
error: a `bool` or counting field has no value to check against `choices`
error: `var_min = 3` is more than `var_max = 2`, so nothing could satisfy both
error: `var_min` and `var_max` count values, so the field has to be a `Vec`
error: the default `z` is not one of this field's choices, so it could never be valid
```

Consistent with the rest of this crate: the derive's job is refusing what cannot round-trip or cannot mean anything.

## The declarations reach the spec

A test pins that, because a choice list the parser enforces but the spec omits would be invisible to docs and completions — exactly the kind of half-implemented feature this project keeps turning up.

10 tests: each check, the empty-value case, per-command judging, and a required subcommand.

## Still missing, and now stated in the crate docs

`overrides` and `required_unless` (both need the order flags arrived in, which nothing records yet), nesting past one level, and typed values.

---

> [!NOTE]
> **Medium Risk**
> Touches public `Error` and trait APIs and adds substantial generated validation/env logic that changes how CLIs accept or reject input. Well-tested, but incorrect checks would mis-parse real command lines.
> 
> **Overview**
> Implements the **post-binding** layer the parser deliberately left out: after tokens are bound, generated code now checks required-ness, `choices`, `var_min`/`var_max`, and fills from `env` when argv left a field unset.
> 
> **Validation order** is fixed: env fallback → required checks → choices/bounds. Only the selected subcommand is judged. A bare `subcommand: T` field is now allowed and reports `MissingSubcommand` when absent (previously had to be `Option<T>`).
> 
> Extends `Error` with `MissingRequired`, `InvalidChoice`, `VarTooFew`, `VarTooMany`, and `MissingSubcommand`, and marks it `non_exhaustive`. Adds `check` to `CommandArgs`/`Subcommands`. Declarations also flow into the emitted spec so docs and completions see the same constraints.
> 
> Contradictions (`choices` on bool, `var_min > var_max`, default outside choices) are compile errors. New conformance tests cover each check, empty values counting as given, and env precedence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bc52c746c9f9799d4f3e7079b3c64070162fd53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added post-parse validation for required values, allowed choices, variable-length arguments, and required subcommands.
  * Added environment-value fallback and clear precedence when command-line values are provided.
  * Added support for validating only the selected subcommand.
  * Expanded generated command specifications with choices and argument bounds.

* **Bug Fixes**
  * Improved handling of boolean environment values and invalid defaults.

* **Documentation**
  * Documented validation order, scope, supported constraints, and current limitations.

### DIFF
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -268,7 +268,11 @@ pub enum Event<'t, 'v> {
 /// Carries the offending token so a caller can render a good message, but no
 /// message of its own: rendering belongs to a cold path, and building a string
 /// here would allocate on the way to reporting that nothing was allocated.
+///
+/// `non_exhaustive`, because an error enum grows: a caller matching on it needs a
+/// fallback arm so that recognizing a new failure is never a breaking change.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum Error<'t, 'v> {
     /// A flag-like token matched no flag in scope. `token` is the whole token as
     /// typed, so a bundle containing an unrecognized letter reports `-fz` rather
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -284,6 +284,39 @@ pub enum Error<'t, 'v> {
     ArgRequiresDoubleDash { arg: &'t Arg<'t> },
     /// The command tree is deeper than [`MAX_DEPTH`].
     TooDeep,
+
+    // The rest are raised *after* the parse, by whoever owns the target type: they
+    // need to know a value's declared type, which the parser deliberately does not.
+    // They share this enum so that a caller has one error to handle rather than two.
+    /// Something the command requires was never given.
+    MissingRequired {
+        /// The flag or argument's name, as the spec calls it.
+        name: &'t str,
+    },
+    /// A value was given that is not among the declared choices.
+    ///
+    /// Carries the choices rather than the offending value: rendering the value means
+    /// owning it, and an error that allocates on a path this crate promises not to
+    /// allocate on would be a poor trade for a better message. Diagnostics are a
+    /// separate layer.
+    InvalidChoice {
+        name: &'t str,
+        choices: &'t [&'t str],
+    },
+    /// Fewer values than `var_min`.
+    VarTooFew {
+        name: &'t str,
+        min: usize,
+        got: usize,
+    },
+    /// More values than `var_max`.
+    VarTooMany {
+        name: &'t str,
+        max: usize,
+        got: usize,
+    },
+    /// A subcommand was required, and none was given.
+    MissingSubcommand,
 }
 
 /// Interpret a value as UTF-8.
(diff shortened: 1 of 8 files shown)


jdx:
Four findings fixed, and two of them are about not breaking people.

**A counting field with `env` discarded the value and marked the field filled anyway.** It now reads a number — `EX_VERBOSE=3` is how you say `-vvv`, since the environment cannot repeat a flag — and something unparseable leaves the field alone rather than counting as given. Both cases are pinned.

**A required *flag* said so at run time but not in the spec**, so docs and completions described a different CLI from the one that runs. Same rule an argument already followed. That is the second time this PR has caught a declaration enforced in one place and missing from the other, which is why the test asserting "the declarations reach the spec" earns its keep.

**On semver, you were right to raise it.** `Error` is now `non_exhaustive` — an error enum grows, and recognizing a new failure should never be a breaking change. This is the last time adding a variant breaks anyone, and it immediately proved itself by breaking the conformance harness's exhaustive match, which now has a wildcard. Both `check` methods default to `Ok(())`, so a hand-written implementation with nothing to check is not forced to say so.

**The environment test race is the most interesting one.** My comment claimed consolidating the cases into one `#[test]` handled it, and that was wrong in a way worth spelling out: it removed the race between *those cases* and not the race with their **neighbours** — including tests that read the environment only *implicitly*, by parsing a CLI that has an `env` field. Those tests would have been flaky depending on scheduling.

The fix is a separate test binary, since each integration test file is its own process. The file says so at the top, because the next person will be tempted to merge it back.

PR #818, opened 2026-08-11
feat(derive): nest commands to any depth

jdx:
Third of the stack. mise reaches four levels — `mise bootstrap macos launchd-agents apply` — so #816's one-level limit had to go.

## A nested command is not a special case

An `Args` struct carries a `subcommand` field exactly as the root does:

```rust
#[derive(Args)]
struct Settings {
    #[usage(long)]
    file: Option<String>,
    #[usage(subcommand)]
    command: SettingsCommands,   // and these can nest again
}
```

What made that cheap was pulling the wiring out of the root's emitter into one place both use — the tables to splice, the state to carry, how an event is routed, how the field is built. **The root now differs from a nested command only in how it is entered.**

## Two consequences worth flagging

**`build` and `select` are fallible.** A command can require a subcommand of its own, and "none was given" is only knowable where the value has to exist.

**Every generated reference to a user type now sits at one scope.** The root's post-binding checks were emitted beside the parse rather than inside the generated module — harmless until a *nested* command's check referred to the user's enum from there and `super::` escaped the crate root:

```
error[E0433]: too many leading `super` keywords
```

That was worth more than the fix: two emitters had drifted into putting the same code at different scopes, and the bug only appeared once the two met. Both now put the checks in the module and call them, so there is one answer to "where does this code live". A unit test pins `in_module`'s behaviour for plain, `crate::`, `::absolute`, `self::`, and `super::` paths, since I had reasoned about it twice and been wrong once.

## Verified

Eight tests on a three-level CLI: routing to the deepest command, each level keeping its own flags, a global reaching any depth *and* working after the deepest command, a middle command requiring one of its own, a deep command's requirements being its own rather than its parent's (`settings set jobs` → `MissingRequired { name: "value" }` while `settings ls` is fine), and the spec:

```kdl
cmd "settings" help="Manage settings" {
    flag "--file" help="Which settings file" {
        arg "<file>"
    }
    cmd "set" help="Set a value" {
        arg "<key>" help="Which setting"
        arg "<value>" help="The value"
    }
    cmd "ls" help="Show every value" {
        flag "--json" help="As JSON"
    }
}
```

## What is left before a mise-shaped CLI is expressible

`flatten`, and the `conflicts`/`requires`/`overrides` family — which need the order flags arrived in, so they want a small ordering record. Then the bench harness and the gate.

---

> [!NOTE]
> **Medium Risk**
> Touches core derive codegen and trait APIs (`CommandArgs`/`Subcommands`), including fallible `build`/`select` and selection-by-position instead of key. Well covered by new nesting conformance tests, but hand-written trait impls would break.
> 
> **Overview**
> **Enables arbitrarily nested subcommands** — an `Args` struct can carry a `subcommand` field exactly as the root does, to any depth (needed for mise's four-level trees).
> 
> Root and nested commands now share one `subcommand_parts` wiring path for tables, routing, checks, and builds. `CommandArgs::build` and `Subcommands::select` become fallible so a middle command can require its own subcommand. Selection uses table position (via pointer identity) instead of command keys, so key collisions cannot pick the wrong variant.
> 
> Also hardens key assignment: fingerprints hash the whole declaration (not just the type name), and flag/arg match arms verify table identity so same-named structs in different modules cannot misbind. Post-binding checks move into the generated module so root and nested code share one scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7822d16cd85d2e28721db4eb24896327794ef66e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for arbitrarily nested subcommands, including command-specific and global options.
  * Added support for required nested subcommands and arguments.
  * Improved routing, validation, and error reporting during nested command construction.

* **Bug Fixes**
  * Resolved naming collisions for same-named arguments in different modules, ensuring commands route correctly.

* **Documentation**
  * Updated documentation to reflect deep subcommand nesting support.

### DIFF
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -747,7 +747,11 @@ fn quoted(value: &str) -> String {
 pub trait CommandArgs: Sized {
     /// Values collected so far. Partly-filled by construction, since a parse can
     /// stop early.
-    type Partial: Default;
+    ///
+    /// No `Default` bound: [`CommandArgs::start`] is what produces a fresh one,
+    /// because a command with subcommands of its own has nested state that a derived
+    /// `Default` cannot set up.
+    type Partial;
 
     /// The parse tables for this command.
     ///
@@ -786,7 +790,10 @@ pub trait CommandArgs: Sized {
     }
 
     /// Build the struct from what was collected.
-    fn build(partial: Self::Partial) -> Self;
+    ///
+    /// Fallible because a command can require a subcommand of its own, and "none was
+    /// given" is only knowable here — at the point where the value has to exist.
+    fn build<'t, 'v>(partial: Self::Partial) -> Result<Self, crate::Error<'t, 'v>>;
 }
 
 /// An enum whose variants are a command's subcommands.
@@ -815,8 +822,11 @@ pub trait Subcommands: Sized {
     /// Build the variant whose command was selected, by its [`Command::key`].
     ///
     /// `None` when no variant was selected, which a caller reads as "no subcommand
-    /// was given".
-    fn select(partial: Self::Partial, key: u64) -> Option<Self>;
+    /// was given". An `Err` comes from building the variant that *was* selected.
+    fn select<'t, 'v>(
+        partial: Self::Partial,
+        key: u64,
+    ) -> Result<Option<Self>, crate::Error<'t, 'v>>;
 }
 
 #[cfg(test)]
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -817,15 +817,22 @@ pub trait Subcommands: Sized {
     ///
     /// A flag that `install` requires says nothing about an invocation that ran
     /// `run`, so only the command that was actually reached is judged.
-    fn check<'t, 'v>(partial: &mut Self::Partial, key: u64) -> Result<(), crate::Error<'t, 'v>>;
+    ///
+    /// Identified by its position in [`Subcommands::COMMANDS`] rather than by its
+    /// key: the position is found from the table's own address, so two commands whose
+    /// keys happen to collide still cannot be confused for one another.
+    fn check<'t, 'v>(
+        partial: &mut Self::Partial,
+        selected: usize,
+    ) -> Result<(), crate::Error<'t, 'v>>;
 
-    /// Build the variant whose command was selected, by its [`Command::key`].
+    /// Build the variant at `selected`, a position in [`Subcommands::COMMANDS`].
     ///
     /// `None` when no variant was selected, which a caller reads as "no subcommand
     /// was given". An `Err` comes from building the variant that *was* selected.
     fn select<'t, 'v>(
         partial: Self::Partial,
-        key: u64,
+        selected: usize,
     ) -> Result<Option<Self>, crate::Error<'t, 'v>>;
 }
 
--- a/conformance/tests/nesting.rs
+++ b/conformance/tests/nesting.rs
@@ -1,0 +1,169 @@
+//! Commands inside commands, to any depth.
+//!
+//! mise is four levels deep in places (`mise bootstrap macos launchd-agents apply`),
+//! so one level was never going to be enough. A nested command is not a special case
+//! here: an `Args` struct carries a `subcommand` field exactly as the root does, and
+//! generates the same code for it.
+
+use std::ffi::OsStr;
+
+use usage::Spec as LibSpec;
+use usage_argv::Error;
+use usage_derive::{Args, Cli, Subcommands};
+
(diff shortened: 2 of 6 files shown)


jdx:
Three findings fixed, and writing the test for the first exposed a second bug of mine — which is becoming the pattern worth noting.

**Keys hashed the type's *name***, so two same-named structs in different modules got the same key and the wrong command was selected. They now hash the **whole declaration**: a macro cannot see a module path, but it can see the item, so two types have to be *identical* to collide. The duplicate-key assertion in `to_kdl` stays as the backstop.

**Then the test for that refused to compile** — and the message was mine:

```
error: two variants both wrap `Op`, and a command collects into the struct that declares it
```

`add::Op` and `remove::Op` are different types. The check I added in #816 compared rendered type *names*, and `type_name` renders only the last segment, so it was refusing two perfectly good commands. It compares whole paths now. That check has now been wrong in both directions — too permissive in #816, too strict here — which is what you get for comparing types by a string built for a different purpose.

Two smaller ones: `check` reads its parameter unconditionally, since a command that declares nothing to check would otherwise leave it unused in the user's crate where nobody can silence it; and a comment about deriving `Default` no longer described the code.

9 tests on this file. The new one asserts both commands route correctly *and* that `to_kdl`'s duplicate-key assertion stays quiet, so a regression shows up as either a wrong variant or a panic.

jdx:
Good push on this one — hashing the whole declaration *narrowed* the collision to two byte-identical structs in different modules, but narrowing is not fixing, and the failure mode was still "one command's flag fills another's field".

There is a better answer than hashing harder: **`static` items have distinct addresses**, so each arm now checks that the event came from its own table.

```rust
#key if ::core::ptr::eq(*flag, &FLAG_0) => { … }
```

The key still gets to the right arm in one jump; the check costs one comparison on the arm that was already selected. A collision now means an event goes **unclaimed** rather than misbound — and `to_kdl`'s duplicate-key assertion still turns it into a failed test rather than a puzzle. That closes the class properly instead of making it less likely.

(The other threads in this batch are re-posts from before the previous push: the unread `partial`, the stale `Default` comment, and the name-based keys are all fixed.)

jdx:
Right on both counts, and the second one mattered.

**The identity guard covered half the dispatch.** Flags and arguments verified which table an event came from; commands still compared keys — so which subcommand was reached, which one to check, and which variant to build could all still be confused by a collision. Guarding half of it is not guarding it.

A command is now identified by its **position in `COMMANDS`**, found by matching the table's own address:

```rust
COMMANDS.iter().position(|c| ptr::eq(*c, *cmd))
```

Key comparison is gone from command selection entirely, and `check`/`select` became a jump on a small integer instead of a chain of comparisons — so this is slightly faster as well as exact.

The duplicated `Spec::to_kdl` sentence in the crate docs is also gone; my previous edit left the old tail behind.

jdx:
Went back through all eight comments on this PR to make sure nothing was left. Where each landed:

| finding | status |
| --- | --- |
| cursor: command select ignores pointer identity | fixed here — selection is by position found via `ptr::eq` |
| greptile: colliding keys select wrong command | fixed by the same change: address identity cannot be fooled by a key collision |
| CodeRabbit: `check` can leave `partial` unread | fixed here (`let _ = &partial;`) |
| CodeRabbit: stale comment about `Default` | fixed here |
| CodeRabbit: duplicated `Spec::to_kdl` fragment | fixed here |
| CodeRabbit (Major): unqualified declaration as command identity | fixed in #823 |
| greptile ×2: identical declarations still collide | same fix, #823 |

The Major one is worth a note, because I had waved it off in a comment in this PR — I'd argued the residue was harmless since a key only chooses which `match` arm to jump to and each arm then checks the event came from its own table. That reasoning was right as far as it went, and it is why nothing ever misbound: `add::Op` and `remove::Op` really do bind their own flags. What it missed is the safety net above it. `Spec::to_kdl` asserts that no two things in a CLI share a key, so a perfectly good CLI failed that assertion in the debug build tests run in. I reproduced it, then fixed the identity: the declaration's hash is now mixed with `module_path!()` in a `const fn`, so the two halves of a key are decided in the two places that can each see half.

The regression test needed a second pass too — cursor caught that my two commands were not actually identical (their doc comments differed, and the fingerprint hashes the whole item), so it passed with or without the fix. They are byte-identical now and I checked it fails when the module is stubbed out.

The identity fix sits in #823 rather than here because that is where I was working when it reproduced; the stack merges together, so it lands with this. Say the word if you would rather each PR stood alone and I will move it down.

PR #819, opened 2026-08-12
feat(spec): add `conflicts` to flags

jdx:
The spec can say `overrides`, `required_if` and `required_unless`, but not that two flags must not be given together. mise declares forty `conflicts_with` relationships in clap, and the `From<&clap::Command>` bridge was dropping every one — so a spec generated from mise's CLI silently lost them.

Per the canonicality law (code authors, the spec defines), the spec gains the property before the derive can carry it.

## What's here

- **`SpecFlag.conflicts`**, parsed as a property for one target (`conflicts="--stdin"`) or a child node for several, and serialized back the same way — mirroring `overrides`.
- **The clap bridge** now reads `Command::get_arg_conflicts_with`. clap keeps conflicts on the command rather than on the argument, so the conversion in `cmd.rs` is the only place both are in view.
- **Enforcement in usage-lib's parser**, beside the requirement checks: a conflict is a question about the invocation as a whole, not about any one token, since the flag it conflicts with may still be ahead of it. It gets its own loop because the requirement loop skips exactly the flags a conflict check needs — the ones that *were* given.
- **Four corpus vectors** and a `conflicting_flags` error code.

`overrides` vs `conflicts` is the distinction worth keeping straight, and it's now documented: `overrides` resolves a collision by letting the last flag win (`--color`/`--no-color`, where a later flag is a correction); `conflicts` reports it, for combinations with no sensible meaning, where silently honouring one hides the mistake.

## Notes for review

- clap exposes only the direction the conflict was declared in, so the spec records it once. That's sufficient because the check looks at every flag that was given — there's a test asserting both orders are refused.
- `UsageErr::InvalidFlag { reason: "conflicts with …" }` is reused rather than adding a variant: `UsageErr` isn't `#[non_exhaustive]`, so a new variant would be a breaking change for a diagnostic that reads fine as an invalid-flag reason.
- Conflicts fire on flags the user gave (including via `env`, consistent with how `required_unless` treats a selector), never on defaults.
- Stacked on #818.

---

> [!NOTE]
> **Medium Risk**
> Touches core argv post-binding validation and the clap→spec bridge, so incorrect conflict handling could accept or reject flag combinations wrongly. Changes are additive and covered by corpus/unit tests.
> 
> **Overview**
> Adds **`conflicts`** on flags so two flags can be declared mutually exclusive and rejected when both are present — unlike `overrides`, which lets the last one win.
> 
> The clap bridge now carries `conflicts_with` into the spec (it was previously dropped because clap stores conflicts on the command). The parser enforces conflicts after binding, in either order, including values supplied via env.
> 
> Also wires `conflicts=--out-file` onto the CLI's `--multi` markdown flag, and documents the new `conflicting_flags` error in the corpus and spec docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5de5a532e62fe95596e95dc0c6ca5f9fa66415ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cli/usage.usage.kdl
+++ b/cli/usage.usage.kdl
@@ -161,7 +161,7 @@ Common sections: - 1: User commands - 5: File formats - 7: Miscellaneous - 8: Sy
         flag "-f --file" help="A usage spec taken in as a file, use \"-\" to read from stdin" required=#true {
             arg <FILE>
         }
-        flag "-m --multi" help="Render each subcommand as a separate markdown file"
+        flag "-m --multi" help="Render each subcommand as a separate markdown file" conflicts=--out-file
         flag --html-encode help="Escape HTML in markdown"
         flag --out-dir help="Output markdown files to this directory (required when using --multi)" effect=write {
             arg <OUT_DIR>
--- a/conformance/src/lib.rs
+++ b/conformance/src/lib.rs
@@ -157,6 +157,8 @@ pub enum ErrorCode {
     VarTooFew,
     /// A variadic got more values than `var_max`.
     VarTooMany,
+    /// Two flags declared to conflict were both given.
+    ConflictingFlags,
 }
 
 /// Whether the reference implementation matches a vector's expectation.
--- a/conformance/src/reference.rs
+++ b/conformance/src/reference.rs
@@ -120,7 +120,9 @@ fn classify(msg: &str) -> Observed {
     } else if msg.contains("Invalid flag") {
         // usage-lib funnels several distinct situations through InvalidFlag, so
         // the reason has to be read to tell them apart.
-        if msg.contains("requires an argument") || msg.contains("missing value") {
+        if msg.contains("conflicts with") {
+            ErrorCode::ConflictingFlags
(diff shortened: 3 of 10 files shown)


jdx:
Both findings were real; the short-only one was the more serious.

**Short-only conflicts dropped** — fixed. `get_long()` gives nothing for `-q`, so a conflict pointing at a short-only flag was discarded and the spec was left accepting a combination clap rejects. Selectors already take `-q` as readily as `--long` (`flag_keys` renders both), so the bridge now falls back to the short form when it is the only name a flag has. Test added with a short-only target.

A conflict pointing at a *positional* is still dropped, and now deliberately, with a comment saying so: the spec has no way to name a positional in a conflict, and writing `--<name>` would produce a selector that matches nothing — which reads as a relationship that holds while doing nothing.

**Empty `conflicts` serialized** — already fixed, in the commit after the one reviewed. The render check caught it first: every flag in `docs/cli/reference/commands.json` had grown a `"conflicts": []`.

PR #820, opened 2026-08-12
feat(derive): declare which flags conflict and which require each other

jdx:
Stacked on #819, which taught the spec `conflicts`. This teaches the derive to declare it, along with the other two relationships that need only *whether* a flag was given: `required_if` and `required_unless`.

```rust
#[derive(usage::Cli)]
struct Rel {
    /// Read from a file
    #[usage(long, required_unless("--url", "--stdin"))]
    file: Option<String>,
    /// Read from standard input
    #[usage(short = 's', long, conflicts("--file", "--url"))]
    stdin: bool,
    /// Where to write
    #[usage(long, required_if = "--stdin")]
    out: Option<String>,
}
```

Each takes a flag the way the spec names one — `"--long"` or `"-s"` — with one as a value and several as a list. They reach `FlagMeta`, so the emitted KDL says what the struct says and `usage g markdown` and the completions can mention them.

## Worth a look

- **A selector naming no flag is a compile error**, and so is one naming its own field. This is the concrete advantage of declaring relationships in code rather than in KDL: a typo'd selector in a hand-written spec is a relationship that quietly does not hold, and nothing ever tells you. Unit-tested through `Cli::from_input`, including resolution by long, short, and negation spelling.
- **Env counts on both sides.** The checks read `__given_*` after the environment fallback, so a value from the environment is as good as a typed one — matching clap (verified directly, including env-on-both-sides) and #819's usage-lib behaviour. An asymmetric rule would make the same pair a conflict or not depending on which side happened to be typed.
- **Conflicts are reported before required-ness**, matching usage-lib: when a conflict has also left something unfilled, "you gave two flags that cannot go together" beats "give me one more flag". There's a test for that ordering.
- **`overrides` is deliberately not here.** It is the one relationship that asks which flag came *last* rather than whether both arrived, and nothing records arrival order yet — so it gets its own PR rather than a half-answer in this one.
- The derive docs had a duplicated "Typed values" bullet and still claimed `env` was unread and required-ness unreported, both of which stopped being true a few PRs ago. Corrected.

---

> [!NOTE]
> **Medium Risk**
> Adds new post-binding constraint enforcement that changes which command lines parse successfully, but the checks are well-tested and do not touch auth or data handling.
> 
> **Overview**
> Teaches the derive to declare **flag relationships** that only need whether a flag was given: `conflicts`, `required_if`, and `required_unless`.
> 
> ```rust
> #[usage(long, conflicts("--file", "--url"))]
> stdin: bool,
> #[usage(long, required_if = "--stdin")]
> out: Option<String>,
> ```
> 
> Selectors use the spec's spelling (`"--long"` / `"-s"`), resolve at compile time (unknown or self-referential names are errors), and flow into `FlagMeta` so emitted KDL, docs, and completions see them. Post-binding checks enforce them after env fallback — env counts on both sides — and report conflicts before missing-required errors. `overrides` is deliberately deferred; it needs arrival order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3d833d4f2969d6006e640216bfdccadf5b68f66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -112,9 +112,9 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       paragraph short, whole block long), and spec emission, from usage-native
       attributes. Unsupported field types are a compile error rather than a
       surprise, and the messages point at the offending field.
-- [ ] **Subcommands in the derive** — one command per struct today. Needs an enum
-      of variants, cross-type table references, and a nested command path through
-      the parse function.
+- [x] **Subcommands in the derive** — an enum of variants, each holding the struct
+      that declares its flags, nested to any depth. A command is selected by its
+      position in the parent's table, found from the address the parser hands back.
 - [ ] **Typed values** — fields are text today (`String`, `Option<String>`,
       `Vec<String>`, `bool`, counting integers). Parsing into other types needs an
       error type for value conversion, which is also where `env`, required-ness,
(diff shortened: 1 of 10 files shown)


jdx:
Four findings, all valid, all fixed. Two of them were the same underlying mistake seen from different angles.

**Relationships on positionals** (greptile P1, bugbot) — a positional could declare `conflicts`/`required_if`/`required_unless`, the check was generated, and `ArgMeta` had nowhere to record it. So the binary enforced something its own spec, docs and completions could not describe. `SpecArg` has no such fields either, so per the canonicality rule the derive refuses it rather than the spec growing them speculatively — the same treatment `count` and `negate` already get.

**`required_unless` on a bare `String`** (bugbot) — accepted and unreachable: the shape-based required check runs first and knows nothing about the condition, so the declared exception could never take effect. It needs somewhere to put "absent", which means an `Option`. Refused with that message.

**`default` plus a condition** (bugbot) — a defaulted field was reported missing by `required_if`/`required_unless`, because applying a default does not set `__given_*`. Plain required-ness has always skipped defaulted fields and usage-lib does too; the relationship checks now match. Runtime test added, and I mutation-checked it.

**Empty `conflicts()`** (greptile P2) — compiled into no relationship at all. Rejected, as `choices()` already was.

Each has a unit test asserting the message, since a compile error nobody can read is not much better than a silent one.

PR #821, opened 2026-08-12
feat(derive): let a flag displace another, the last one given winning

jdx:
Stacked on #820. The last of the four flag relationships, and the one that needed a different answer.

`conflicts`, `required_if` and `required_unless` only ask *whether* a flag was given, so they are checked once the last token has been read. `overrides` asks which of two flags came **last** — so it is applied while parsing, in the arm that binds the flag: the token that just arrived is the last one so far. Nothing has to record arrival order, which is what had been blocking it.

```rust
#[usage(long, default = "true", overrides = "--plain")]
color: bool,
```

## Semantics, measured rather than assumed

I ran clap 4 directly instead of trusting my memory of it:

| given | clap | here |
| --- | --- | --- |
| `--file a --stdin` | `file=None stdin=true` | same |
| `--stdin --file a` | `file="a" stdin=false` | same |

Two things fall out of that and are now tested:

- **The relationship is bidirectional from one declaration.** Only `--file` says `overrides`, and `--stdin --file a` still displaces `--stdin`. usage-lib does the same.
- **A displaced flag returns to its default, not to blank.** clap reports the loser as `false` rather than absent, so a `--color` that defaults on and loses to `--plain` is back to *on*, not off. The reset `start()` applies is now a named function, since this is its second caller.

A displaced `Vec` is emptied, and values given after the flag that displaced them are kept — order decides, not which side was named in a declaration.

With this the post-binding layer is complete: `required`, `choices`, `env` fallback, defaults, `var_min`/`var_max`, and all four relationships. PLAN.md updated accordingly.

---

> [!NOTE]
> **Medium Risk**
> Touches core derive codegen for flag binding and post-binding (env/required), so incorrect override handling could misbind CLI values. Behavior is covered by new conformance and corpus tests.
> 
> **Overview**
> Completes the post-binding layer by adding **`overrides`**: a flag can displace another, with the last one given winning.
> 
> Unlike `conflicts` / `required_if` / `required_unless`, this is applied **during** parse in the flag-binding arm — arrival order is what matters, so nothing needs to record it separately. The relationship is bidirectional from one declaration, matching clap and usage-lib.
> 
> A displaced flag resets to its **declared default** (not blank); collections are emptied. An `__overridden_*` bit keeps env fallback from reviving a loser and keeps required checks from treating displacement as missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfba697225f2c5750a3c83f5eebc6c633614a346. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -123,11 +123,12 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       (`requires`/`conflicts`/`overrides`/`required_unless`), `var`, `count`,
       `env`, defaults, delimiters, the `double_dash` modes, global flags, flatten,
       boxed subcommand variants, headings, `cfg`-gated variants.
-- [ ] **The post-binding layer** — `required`, `choices`, `env` fallback, defaults,
-      `var_min`/`var_max`, `conflicts`, `required_if` and `required_unless` are done;
-      `overrides` is what remains, and it is the one that needs arrival order rather
-      than only whether a flag was given. These need a value's type, so they belong
-      with the derive rather than in the parser.
+- [x] **The post-binding layer** — `required`, `choices`, `env` fallback, defaults,
+      `var_min`/`var_max`, `conflicts`, `required_if`, `required_unless`, and
+      `overrides`. All of them need a value's type, so they belong with the derive
+      rather than in the parser. `overrides` is the one that happens _during_ the
+      parse instead of after it: it asks which of two flags came last, which only the
+      arriving token knows.
 
 ### Spec gaps found on the way
 
--- a/conformance/tests/derive.rs
+++ b/conformance/tests/derive.rs
@@ -573,6 +573,12 @@ struct Related {
     /// Read from standard input
     #[usage(short = 's', long, conflicts("--file", "--url"))]
     stdin: bool,
+    /// Colorize output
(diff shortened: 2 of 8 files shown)


jdx:
Both bots found the same real bug, and bugbot's framing was the accurate one: clearing `__given_*` is not the same as recording that a flag *lost*.

The environment fallback runs after the parse and only asks whether a flag was given, so a displaced flag with an `env` var got refilled and marked given again — both sides of the override standing, which is the opposite of what the command line said. A displaced bare `String` was reported missing for the same reason.

Fixed the way usage-lib does it: a flag on either side of an `overrides` carries `__overridden_*`, and both the environment fallback and the required check leave it alone. Giving the flag again clears the mark, which is what makes alternating flags work — `--include a --all --include b`.

Only flags that can actually lose an override carry the marker, so a CLI with no `overrides` generates exactly what it did before.

Two tests in the env test binary (its own process, since env vars race across tests in one binary), plus a corpus vector pinning the interaction across implementations — usage-lib already agrees with it. I mutation-checked the guard by removing it and confirming the revival test fails.

PR #823, opened 2026-08-12
fix(derive): let a `--`-only argument follow a variadic

jdx:
Stacked on #821.

The derive refused any argument after a variadic, on the grounds that a variadic takes every remaining word. True except in the one case that matters: an argument only fillable after a `--`, since the `--` is what stops the variadic.

mise declares exactly that on `run`, `exec` and `git`:

```kdl
arg "[ARGS]…"      help="Arguments to pass to the tasks. Use \":::\" to separate tasks" var=#true
arg "[-- ARGS_LAST]…" var=#true hide=#true
```

So the check was refusing a shape three of mise's commands depend on. Both parsers were already right — usage-argv binds it, usage-lib binds it — only the derive's validation disagreed.

## How it was found

By generating a shadow CLI from mise's committed 5,592-line spec and trying to compile it. Four commands failed, and the parser was not at fault. That's what the shadow exercise is for, and it's the reason to do it before touching mise rather than after.

Two corpus vectors pin the behaviour, with and without the separator, and a derive test checks the emitted spec still says which argument is which — otherwise docs and completions would describe a different CLI from the one that runs.

---

> [!NOTE]
> **Medium Risk**
> Touches derive validation and key generation for all derived CLIs; behavior is corrective but key values change and incorrect validation could accept unreachable argument shapes.
> 
> **Overview**
> **Allows a `double_dash = "required"` argument to follow a variadic** in derive validation. The old rule rejected any argument after a variadic; the `--` that such an argument requires is exactly what stops the first variadic, which mise needs on `run`/`exec`/`git`. Anything after a separator-consuming variadic is still rejected.
> 
> **Folds `module_path!()` into generated keys** via a new `usage_argv::key_base`, so byte-identical declarations in different modules no longer collide. That collision used to fail `Spec::to_kdl`'s uniqueness assert on otherwise valid CLIs. Codegen now emits named key consts (needed for `match` patterns) instead of inlining the hash.
> 
> Adds corpus vectors and derive/identity tests for both behaviors, plus a note that `var_max` is a post-binding check and does not hand leftover words to the next argument.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd0436c3af6561012c3dece838eaef51d7bd9b52. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -333,6 +333,30 @@ pub enum Error<'t, 'v> {
     MissingSubcommand,
 }
 
+/// The high half of every key one declaration's items get.
+///
+/// A derive cannot see other expansions, so it cannot hand out keys from a shared
+/// counter: it hashes the declaration it was given instead. It cannot see a module path
+/// either, which is why the module is mixed in *here* — `module_path!()` is available to
+/// the generated code as a compile-time string, so two byte-identical declarations in
+/// different modules end up with different keys rather than colliding.
+///
+/// `declaration` is a hash the derive computed over the item's own tokens.
+pub const fn key_base(module: &str, declaration: u32) -> u64 {
+    // FNV-1a, continuing from the declaration's hash rather than starting over, so both
+    // halves contribute. Spelled out rather than taken from a `Hasher`, which is not
+    // guaranteed to be stable between compilations — and these are baked into a binary.
+    let mut hash: u32 = declaration;
+    let bytes = module.as_bytes();
+    let mut i = 0;
+    while i < bytes.len() {
+        hash ^= bytes[i] as u32;
+        hash = hash.wrapping_mul(0x0100_0193);
+        i += 1;
+    }
+    (hash as u64) << 32
+}
+
 /// Interpret a value as UTF-8.
 ///
 /// The parser hands back bytes borrowed from `argv`; this is the conversion most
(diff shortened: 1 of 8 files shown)


jdx:
Valid, and the framing was the useful part: I had written the exemption as a property of each argument when it is a property of the command line. Two variadics behind a `--` both passed, and the first would take every word after the only separator there is — exactly what the rule exists to refuse.

So a variadic that takes what follows the separator now ends the list: nothing may come after it, variadic or not. Two unit tests, one for a second `--`-required variadic and one for a plain argument trailing it.

jdx:
The `var_max` finding is re-posted from the earlier commit; it's answered in the previous round rather than outstanding. Short version, since it turned out to be more interesting than a yes/no: I measured it, and **usage-argv and usage-lib disagree** — usage-argv gives the variadic every remaining word, usage-lib stops it at `var_max` and fills the next argument. The grammar sides with usage-argv ("collects every word still available"), so the check stays and `docs/spec/argv.md` now says a bound is a check rather than a limit. The divergence is recorded as `a-bound-does-not-stop-a-variadic` in the corpus, with a note that **clap behaves the way usage-lib does** — so which one is right is a grammar question for @jdx, not a bug on either side.

The identity-test finding was right and worth more than its severity. `add::Op` and `remove::Op` were not byte-identical — their doc comments differed, and the fingerprint hashes the whole item — so their keys already diverged and the test passed with or without the fix. It asserted the fix while demonstrating nothing. They're identical now, and I checked it fails when the module is stubbed out of `key_base`: "two things in this CLI share a key", which is exactly what it should notice.

PR #824, opened 2026-08-12
feat(bench): generate a shadow CLI from a spec, and compile mise's

jdx:
Stacked on #823. This is the first half of the gate — the thing the whole experiment is supposed to be judged by.

`xtask gen-shadow <spec.kdl> <dir>` reads a usage spec and writes a crate of derived types declaring the same CLI. **mise's committed 5,592-line spec compiles**: 211 commands, 711 flags, 128 arguments, four levels deep, in 2.6 seconds.

## What it drops, and says so

```
dropped, because the derive cannot express it yet:
  `double_dash = "automatic"` on an argument: 3
  a default on a flag that collects values: 1
  a flag's second long or short form: 13
```

Seventeen out of roughly a thousand. The generator names each class and counts it, because a silent cap reads as "all of it was expressible".

## A first measurement

The gate parses one command line and exits, beside a null binary that does everything *except* parse. Subtracting the second from the first is what turns a process measurement into a parser measurement — about two thirds of a small binary's instructions are the dynamic loader and libc starting up, and neither parser is responsible for those.

| invocation | instructions to parse |
| --- | --- |
| `mise` | 99,281 |
| `mise use -g node@20` | 104,720 |
| `mise settings set experimental true` | 104,650 |
| `mise -vv ls --installed` | 105,754 |
| `mise tasks run build -- --verbose` | 104,650 |

Near constant regardless of depth, which is what static tables should look like: there is no tree to build. **The clap side is the next PR** — until it lands, this number stands on its own rather than as a ratio, and the ~3.1M figure from the original analysis is not something I have reproduced here.

## Two things worth review

- **The generated crate is excluded from the workspace.** A shadow of mise trips `large_enum_variant` on the commands with thirty flags, which the real mise answers by boxing its variants — something the derive cannot express yet, so it's on the roadmap rather than papered over with an `allow`. The gate depends on the shadow by path so it is still built, and the smoke tests live in the gate crate, where they get linted like anything else.
- **The shadow is checked in, not built by a `build.rs`.** The compile-time comparison wants a fixed input, and a generated file in the tree is a diff a reviewer can read when the derive's vocabulary changes. CI runs `mise run gen-shadow` and fails on a diff, the same way it does for `render`.

Along the way, a note on fidelity: mise's spec declares **no positionals on the top-level `run`**, and that is deliberate — `src/cli/usage.rs` clears them (`run.args = vec![]`) and adds a `mount` of `mise tasks --usage` plus `restart_token = ":::"`, so the project's real tasks are what complete after `run` rather than a static `[TASK]`. The shadow can carry neither the mount nor the restart token yet, so its `run` is slightly cheaper to parse than real mise's would be. The smoke tests use `tasks run` for the separator shape and the root's own `[TASK]` for a bare task.

(I first suspected the clap bridge was dropping them, since `Run(Box<run::Run>)` is the only boxed variant of the commands involved and the same struct registered unboxed under `tasks run` kept all three positionals. I tested that against clap 4 directly: boxing loses nothing, in clap or in the bridge. Recording it here because it is a tempting wrong answer.)

---

> [!NOTE]
> **Medium Risk**
> Large generated shadow and new workspace/CI wiring for the gate; not security-critical, but the shadow exercises the derive at full mise scale and CI will fail on generator drift.
> 
> **Overview**
> Adds the first half of the **performance gate**: generate a compileable shadow of a real CLI from its usage spec, then measure parsing at mise's full scale.
> 
> **`xtask gen-shadow`** turns any `.usage.kdl` into a crate of derived types. mise's checked-in 5,592-line spec compiles (211 commands, 711 flags, 128 args). Features the derive cannot express yet are counted and reported rather than silently dropped. The generated `shadow-mise` crate is committed and **excluded from the workspace** (it trips `large_enum_variant`); the gate depends on it by path.
> 
> The new **`gate`** crate provides one-shot parse binaries (`parse-usage` vs a null `parse-none` baseline) plus smoke tests for real mise invocations — including the `[ARGS]…` / `[-- ARGS_LAST]…` separator shape and nested/global-flag cases.
> 
> CI now runs `mise r gen-shadow` and fails if the checked-in shadow drifts, same as `render`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4470be0e264b5bb9ffc609b4d07513b14d6be8bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Both findings were right, and the alias one matters more than its severity suggests.

**Aliases dropped without a count.** The filter itself is correct — an alias and the name it points at share a map entry, so generating both would declare the same command twice — but not counting it made the report say 17 losses where there were 108: **67 visible aliases, 24 hidden ones**, plus 2 mounts and 2 restart tokens. The generator's whole job is to say what could not be expressed, so that was the one bug it could not afford. Aliases are now the largest known gap in the derive rather than an invisible one, and PLAN.md records them as the next thing mise needs.

**`subcommand_required` ignored.** Every command got an `Option<Commands>`, so the 27 of mise's commands that require a subcommand accepted invocations the real CLI refuses. Fixed, with a test on `bootstrap accounts` — which requires one while `bootstrap` itself does not, so reading the spec is the only way to get it right.

I also corrected the note at the end of the description. I had said mise's missing positionals on the top-level `run` "may be a loss in mise's own clap→spec dump rather than intent" — it is intent: `src/cli/usage.rs` clears them and adds a mount of `mise tasks --usage` so real task names complete after `run`. I had a plausible wrong theory (`Run(Box<run::Run>)` is the only boxed variant, and the same struct registered unboxed under `tasks run` kept its positionals) and tested it against clap 4 rather than reporting it: boxing loses nothing.

jdx:
Three of the four were real, and two were tests of mine that could not fail.

**The separator test** only checked that `tasks run` had been selected, while claiming to cover the `[ARGS]…` / `[-- ARGS_LAST]…` shape that made the derive's validation wrong. It now names each word on each side of the `--`, plus a second positional so a merge would be visible.

**The global-flag test** was worse: it asserted only that `ls` was selected. That could not fail — unknown flag-like words are values by default and `ls` has a variadic positional, so a global that stopped being recognized after the command would still parse, with `-C` and `/tmp` landing quietly in the variadic. It asserts the bound value now.

**Arg-level `env` and `help_heading`** were being dropped with no count, and both reach `ArgMeta`. Fixed — and since mise's spec puts `env` only on flags, the regenerated shadow doesn't change and nothing in it would have caught this. So `render` is now separate from `generate` and has tests of its own on specs written for the purpose, including one that checks both dialects make the same claims about the same flag.

`default_subcommand` was already fixed in the commit after the one reviewed.

jdx:
Good catch, and it was a regression from my own fix in the previous round — both shapes were visible in the regenerated shadow, exactly as described.

`--security`'s long help ends its first sentence with a period the short form leaves off, so stripping the short form left `.` behind and `trim` could not remove it from the middle: it became `/// .` on a line of its own. And where the remainder *opened* with an indented example, `trim` took the indentation along with the blank line — `mise activate --shims`'s `PATH="$HOME/…"` lost its four spaces.

Now only the orphaned period and the separating newlines go, since spaces are not in the set, and a long form that adds nothing but a period says nothing at all.

Both have tests, and the useful part: my first attempt at the indentation test **did not fail** when I restored the blunt version, because it put a line between the short form and the example — which is the one place `trim` cannot reach. Moved the example to the front of the remainder, and now both assertions fail without the fix.

PR #826, opened 2026-08-12
feat(argv): a bound stops a variadic

jdx:
Stacked on #825. Settles the question left open in #823, the way you called it.

## What was open

`var_max` had two coherent readings, and the two implementations had each quietly picked one. For `arg "[a]" var=#true var_max=1` then `arg "[b]"`, given `ex x y`:

| | result |
| --- | --- |
| usage-argv (before) | `a = ["x","y"]`, then `var_too_many` |
| usage-lib | `a = ["x"]`, `b = "y"` |
| clap (`num_args = 0..=1`) | same as usage-lib |

## The decision

**A limit, not a check.** clap behaves that way, every spec in the fleet is generated from a clap command, and it is the only reading under which `[a]… [b]` can be filled at all.

So `var_max` moves into the hot `Arg`/`Flag` tables — the ones kept deliberately free of anything but what binding needs — as `Option<u32>`, and the parser counts what a variadic has taken. `var_min` stays a post-binding check, since no single word tells you a variadic will end up short. `var_too_many` accordingly stops being reachable for a bounded variadic and now describes only a repeatable flag's occurrences.

**The cost: 55 instructions per parse, 0.1%.** Wall clock unchanged at 2.0µs. That was the thing worth checking, since the objection to this reading was that it puts a validation concept in the binder.

## What follows from it

- **The derive's rule relaxes.** An argument after an *unbounded* variadic is still refused; after a bounded one it is allowed, and a bounded variadic behind a `--` no longer spends the separator. The error message now names both things that stop a variadic, so it points at a fix rather than just a refusal.
- **Four corpus vectors**, at the **binding** layer where the question now lives: the bound hands over, the bound with nothing after it is an `unexpected_arg` rather than a silent drop, unbounded still takes everything, and the flag form.
- **One divergence recorded rather than fixed**: for a flag whose single occurrence collects, usage-lib does not collect at all — it takes one value, so the second word falls through to the positional. That predates this change (`long-variadic-flag-arg` documents the same gap) and bounding what usage-argv collects is consistent with it collecting in the first place.

`docs/spec/argv.md` states the rule, and PLAN.md records the decision with its reasoning so it does not get re-litigated.

---

> [!NOTE]
> **Medium Risk**
> Changes core argv binding semantics and parser hot-path behavior; mistakes could mis-route tokens or break CLIs relying on the old post-binding `var_max` for variadics, though coverage is broad.
> 
> **Overview**
> Settles **`var_max` as a binding limit**, not a post-parse count check, so bounded variadics stop collecting and following positionals can be filled (clap `num_args` / usage-lib behavior).
> 
> **usage-argv** adds `var_max: Option<u32>` on `Flag` and `Arg`, tracks per-occurrence counts (`collected` / `arg_taken`), resets counts when advancing positionals or jumping past `--`, and stops variadic flag collection when the bound is hit.
> 
> **usage-derive** emits `var_max` into hot tables for variadic args/flags only; compile-time rules allow an argument after a **bounded** variadic (or after `--`); post-binding `var_too_many` applies only to **repeatable** flags counting occurrences.
> 
> **Conformance bridge**, corpus vectors, spec docs, and **PLAN.md** are updated (including revised bench numbers); bounded variadic-flag behavior vs usage-lib remains a recorded divergence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 285503bb3221458e7ee2cbb9c653083503f63476. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added maximum value limits for variadic flags and positional arguments.
  * Bounded variadic arguments can pass remaining values to subsequent arguments.
  * Added conflict detection for incompatible flags.
  * Documented `var_max` behavior and clarified validation rules.

* **Bug Fixes**
  * Corrected variadic value collection, overflow handling, inline values, and parser state resets.
  * Preserved minimum-value validation after parsing.
  * Improved handling around separators and repeatable flag occurrences.

* **Tests**
  * Expanded coverage for bounded and unbounded variadics, positional fallthrough, overflow, and double-dash scenarios.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -124,8 +124,8 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       `env`, defaults, delimiters, the `double_dash` modes, global flags, flatten,
       boxed subcommand variants, headings, `cfg`-gated variants.
 - [x] **The post-binding layer** — `required`, `choices`, `env` fallback, defaults,
-      `var_min`/`var_max`, `conflicts`, `required_if`, `required_unless`, and
-      `overrides`. All of them need a value's type, so they belong with the derive
+      `var_min`, `conflicts`, `required_if`, `required_unless`, and `overrides` —
+      `var_max` moved to the binder, see the decision below. All of them need a value's type, so they belong with the derive
       rather than in the parser. `overrides` is the one that happens _during_ the
       parse instead of after it: it asks which of two flags came last, which only the
       arriving token knows.
@@ -151,6 +151,14 @@ carrying them rather than being worked around.
       shape mise uses on `run`, `exec` and `git`: `[ARGS]…` for the words before the
       separator and `[-- ARGS_LAST]…` for the ones after. Both parsers already bound it;
       only the derive's validation disagreed. Found by compiling the mise shadow.
+- [x] **Is `var_max` a limit or a check?** — usage-argv let a variadic take every word and
+      judged the count afterwards; usage-lib stopped the variadic at the bound and gave the
+      rest to the next argument. Two coherent readings, and nothing had recorded that the
+      two implementations disagreed. Settled as a **limit**, matching usage-lib and clap's
+      `num_args`, since every spec in the fleet is generated from a clap command and it is
+      the only reading under which `[a]… [b]` can be filled. `var_max` therefore moved into
+      the table binding reads; `var_min` stays a check, because no single word tells you a
+      variadic will end up short. Costs 55 instructions per parse, or 0.1%.
 - [ ] **A mount on the root command** — the spec accepts `mount` only inside a
       `cmd` block, so a CLI whose _top-level_ subcommands are discovered by running
       something cannot say so. Worth deciding whether that is a gap or a deliberate
--- a/PLAN.md
+++ b/PLAN.md
@@ -200,9 +200,9 @@ checked-in `mise.usage.kdl` for both parsers.
       same properties. Three release binaries — one per parser, one that does
       everything except parse — measured by `tak`, which gates the counts in CI.
 - [x] **Perf report** — at mise's full scale, a cold parse costs **50.9k instructions
-      against clap's 5.96M: 117× fewer**, and **2.0µs against 500µs of wall clock**.
+      against clap's 5.96M: 117× fewer**, and **2.1µs against 490µs of wall clock**.
       Measured by differencing two runs of the _same_ binary over how many parses it
-      does, so nothing but the parse varies. clap's 500µs is 315µs building its command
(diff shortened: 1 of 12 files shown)


jdx:
Three real ones, and the first was the serious one.

**`arg_taken` leaked across the separator** (bugbot High, CodeRabbit Major). Jumping to a `--`-required argument moved `arg_pos` without clearing the count belonging to the argument it left, so a bounded variadic before the separator lent its total to the one after it. `[first]… var_max=3` then `[-- rest]… var_max=2`, given `a b -- c d`, stopped `rest` immediately and reported `d` unexpected. There's a corpus vector for it, and I confirmed it reproduces before the fix:

```
- a-bound-before-the-separator-is-not-charged-after-it
     expected: Ok(args: {first: [a, b], rest: [c, d]})
     got:      Failed(UnexpectedArg)
```

**`var_max` was narrowed with `as u32`** (greptile). A spec declaring `4294967296` — no real limit — became zero, which reads as "stop at once": the worst possible misreading. It saturates now, in the derive and the conformance harness both.

**Two documentation gaps**, both correct: the grammar's variadic-flag section didn't mention that a bound also stops collection, and PLAN.md still listed `var_max` among the post-binding checks when this PR is the decision to move it. Fixed.

I also rebuilt this branch on merged `main`. Its history predated the merge of #818–#825, so GitHub was showing the whole stack again in the diff; it's 11 files now, all of them this PR's.

jdx:
Three findings, and the first two were the reading only half-applied.

**A vector still expected the old semantics.** `arg-variadic-max` wanted post-binding `var_too_many` for a variadic positional given more words than its bound — which under the limit reading is an `unexpected_arg` while binding. It passed only because usage-argv skips post-binding vectors while usage-lib still answered the old way, which is exactly the sort of hole the corpus is supposed to close. Corrected — and its `diverges` label is gone, because adopting clap's reading is what made the two implementations agree. That is the nicest outcome available: settling the question deleted a recorded divergence.

**The derive still emitted the post-binding maximum.** For a variadic that check is unreachable, since binding stops at the bound. For a *repeated* variadic it was actively wrong: each occurrence respects the limit, and the total across occurrences would fail a check the invocation never broke. It is now emitted only for the flag whose bound counts occurrences, which no single token can decide. A test gives a bounded flag twice and expects all four values.

**The ratios in PLAN.md disagreed with each other** — the table said 117× while the prose still said 103×, and the wall figure was a stale 245×. All from one measured run now: 50,858 against 5,959,392 instructions, 2.1µs against 490µs.

PR #827, opened 2026-08-12
feat(derive): let a command answer to more than one name

jdx:
Stacked on #826. Closes the largest gap the mise shadow was reporting.

**91 of mise's commands have a second name** — `i` for `install`, `x` for `exec`, `rm`/`remove`/`delete`/`del` for `unset` — and the derive could not declare one, so a shadow of mise rejected invocations the real thing accepts. `gen-shadow` was counting them all as dropped.

## Where the gap actually was

Everything below the derive already worked: usage-argv's `Command` carries `aliases` and `find_subcommand` matches them, and the KDL writer emits them. Only the *declaration* was missing.

It belongs on the variant, not the struct, because the enum is where a command is named — the wrapped struct has no idea what its parent called it. That's the same seam `#[usage(name = "...")]` already used.

```rust
#[derive(usage::Subcommands)]
enum Commands {
    /// Install a tool
    #[usage(alias = "i", alias_hidden = "add")]
    Install(InstallArgs),
    /// Remove a tool
    #[usage(alias("rm", "uninstall"))]
    Remove(RemoveArgs),
}
```

The table gets both kinds, because the parser matches both; the metadata records which are hidden, since that is the only thing the distinction affects — whether help and completions mention them. An alias equal to the command's own name, or an empty one, is a compile error rather than a line that quietly does nothing.

## The shadow carries them now

Both dialects, which took a detail: clap spells one alias and several differently (`visible_alias = "x"` vs `visible_aliases = [...]`), and repeating the singular form is not the same as the plural one. `mise i node@20` and `mise x -- node -e 1` now parse through the shadow, with a gate test for each.

## Cost

353 instructions of the 51.3k parse (clap's grew by 33k, so the 117× ratio holds), and the dropped-property list is 91 entries shorter — the remaining ones are `default_subcommand`, 2 mounts, 2 restart tokens, 13 secondary flag forms, 3 `double_dash="automatic"` and one default on a collecting flag.

---

> [!NOTE]
> **Medium Risk**
> Touches core subcommand routing and codegen for all derived CLIs; behavior change is broad (91 mise commands) but covered by conformance and gate tests, with a known small parse-cost increase.
> 
> **Overview**
> **Subcommands can declare alternate names** via `alias` / `alias_hidden` on enum variants (single value or list). Codegen wires both into parse tables and records hidden aliases in metadata so help/completions only advertise the visible set. Compile-time checks reject duplicate names across variants and aliases.
> 
> **Shadow generation** no longer counts command aliases as unsupported—it emits matching `#[usage(...)]` / clap `visible_alias` / `alias` attributes for both mise shadows. **PLAN.md** drops aliases from the shadow gap list and adds follow-up items: a fixed perf sentinel (shadow growth tripped the 1% gate by ~1.52%) and non-linear subcommand lookup.
> 
> **Tests** cover alias parsing (`mise i`, hidden `x`), KDL/spec hidden-alias metadata, usage-lib resolution, and gate integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2def5d2282ebf46ff280e0340ae4354b5c64b776. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added visible and hidden aliases for commands and subcommands, including common shorthand forms.
  - Visible aliases now appear in help and completions, while hidden aliases remain available for compatibility.
  - Alias metadata is preserved across generated CLI integrations.

- **Bug Fixes**
  - Improved command alias routing, including aliases used with arguments after `--`.
  - Variadic argument handling now consistently processes subsequent values after the separator.

- **Documentation**
  - Updated command alias and variadic argument behavior documentation.

- **Tests**
  - Added coverage for alias resolution, hidden aliases, collisions, and separator handling.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -188,7 +188,7 @@ checked-in `mise.usage.kdl` for both parsers.
 - [x] **Shadow generation** — `xtask gen-shadow` turns any `.usage.kdl` into a crate
       of derived types. mise's committed 5,592-line spec compiles: 211 commands, 711
       flags, 128 arguments, four levels deep, in 2.6s. What it cannot express, it
-      counts: 91 command aliases (67 visible, 24 hidden), 13 secondary flag aliases, 3
+      counts: 13 secondary flag aliases, 3
       `double_dash="automatic"`, 2 mounts, 2 restart tokens, 1 `default_subcommand`, 1
       default on a collecting flag. Aliases are the largest gap and the next thing the
       derive needs for mise; `default_subcommand` is the one that changes the _root's_
@@ -205,6 +205,17 @@ checked-in `mise.usage.kdl` for both parsers.
       does, so nothing but the parse varies. clap's 490µs is 305µs building its command
       tree, ~160µs validating it, and ~24µs actually parsing — so even against clap's
       parse alone, with the tree already built and paid for, this is 12× faster.
+- [ ] **A sentinel the fixture cannot trip** — the three `gate-*` benchmarks measure the
+      mise shadow, and tak gates everything in `tak.toml` at 1%. But the shadow is
+      _designed_ to grow: teaching the derive command aliases added 91 names to it and cost
+      1.52%, which the gate reported as a regression when it was the fixture getting
+      bigger. A regression sentinel has to be a fixed fixture — a small hand-written CLI
+      that never changes — with the shadow benches reported but not gated. Until then a
+      red gate on a PR that grows the shadow is expected, and the PR should say why.
+- [ ] **A subcommand lookup that is not linear** — the 1.52% above is real work: finding a
+      command scans the parent's list, and each candidate now checks its aliases too. At
+      mise's ~100 top-level commands that is worth a sorted table or a hash, and it is the
+      largest remaining item in the parse.
 - [ ] **Differential fuzzing** — proptest over argv against usage-lib on the mise
       spec, to find disagreements the corpus did not think of.
 - [ ] **Perf report** — published honestly, whichever way it goes.
--- a/benches/gate/tests/parse.rs
+++ b/benches/gate/tests/parse.rs
@@ -130,3 +130,20 @@ fn a_command_that_requires_a_subcommand_refuses_to_stand_alone() {
     let a = argv(["build"]);
     Cli::parse_from(&a).expect("a bare task should parse");
 }
+
+#[test]
+fn a_command_answers_to_its_alias() {
+    // 91 of mise's commands have a second name, and until the derive could declare one
+    // the shadow rejected invocations the real thing accepts. `mise i` is `mise install`.
+    let a = argv(["i", "node@20"]);
+    let cli = Cli::parse_from(&a).expect("`mise i` should parse");
+    let Some(Commands::Install(install)) = cli.command else {
+        panic!("`i` should have selected install")
+    };
+    assert_eq!(install.tool_version, ["node@20"]);
+
+    // And a hidden one, which is matched but kept out of help.
+    let a = argv(["x", "--", "node", "-e", "1"]);
+    let cli = Cli::parse_from(&a).expect("`mise x` should parse");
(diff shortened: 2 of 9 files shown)


jdx:
Right, and it was the one validation I'd left out: I checked an alias against its *own* command's name and never against its siblings'.

The parser takes the first table entry that matches, so `Install` plus an `Add` aliased to `install` leaves one of them unreachable — silently, which is the part worth refusing. Every name a command answers to now goes into the same duplicate check, aliases included, so a collision is a compile error naming the word that collides. Tests cover a name-versus-alias collision, two aliases colliding with each other, and the distinct case still compiling.

Also rebuilt on merged `main`, so the diff is this PR's 13 files rather than the whole stack again.

jdx:
The gate on this PR is red, and it is correct to be — but it is measuring the fixture, not the parser. Worth being explicit since I built the gate.

```
| gate-usage | 468,384 → 475,486 | +1.52% ⚠️ |
| gate-clap  | 6,318,720 → 6,346,108 | +0.43% |
```

The mise shadow gained **91 command aliases**, so both parsers have more to scan for the same invocation — about 78 instructions per alias, on a parse that uses none of them. The ratio is unchanged. This is the shadow getting bigger, which is the point of the PR.

Two things follow, both recorded in PLAN.md rather than fixed by quietly raising the threshold:

- **A sentinel has to be a fixed fixture.** tak gates everything in `tak.toml` at 1% with no per-benchmark control, and the shadow is *designed* to grow as the derive learns to express more — mounts and flatten will do this again. The durable fix is a small hand-written CLI as the gated benchmark, with the shadow benches reported but not gated. Until that lands, a red gate on a PR that grows the shadow is expected and the PR should say so, which is what this comment is.
- **The cost points at a real optimization.** Finding a command scans the parent's list and each candidate now checks its aliases too. At mise's ~100 top-level commands that wants a sorted table or a hash, and it is the largest remaining item in the parse itself.

Happy to do either as its own PR if you'd rather see the gate green before merging this — say which.

jdx:
jdx's call: these should be informative, not gated. Done in #831 — the three shadow benchmarks leave `tak.toml`, so usage's own `startup` and `markdown` stay gated as they always were, and the shadow comparison becomes a table appended to this same comment.

Once #831 is in, the gate on this PR goes green for the right reason: nothing here slows the parser, the fixture it measures simply got 91 names bigger.

One thing that came out of implementing it, in case it is useful for tak: `gate = false` on a `[bench.*]` entry is accepted and ignored, as any unknown key is — I checked with a deliberately nonsense key, which was also accepted silently. So there is no way to express "measure but do not gate" in `tak.toml` today.

jdx:
These were not stale re-posts — the bots were right and I was wrong to have assumed otherwise. **This branch had silently lost several of #826's fixes.**

Bugbot's wording was the giveaway: *"advances `arg_pos` but no longer clears `arg_taken`… the corpus case that guarded this path was removed in the same change."* It was. Restoring the vector on this branch made it fail, which is how I confirmed the code fix was gone too. Checking the rest of the file set turned up more of the same:

| #826 fix | was on this branch |
| --- | --- |
| `arg_taken` cleared on the separator jump | **missing** |
| the corpus vector guarding it | **missing** |
| grammar: a bound also stops a variadic flag | **missing** |
| `var_max` saturating instead of narrowing (derive + harness) | **missing** |

The cause was mine: restacking this branch onto the corrected #826 by replaying a content patch captured against the *older* base. A patch built from an old base reverts anything newer that touches the same lines, and it did so quietly — the branch still compiled and its own tests still passed, because the reverted pieces belong to #826's tests, which live on #826.

Rebuilt properly: each branch's own work applied to the corrected parent, with the shadows regenerated rather than merged, and the four fixes verified present on all four branches by grepping the pushed refs rather than trusting the process. CodeRabbit's `docs/spec/argv.md` finding is the same reversion and is fixed by the same restore.

PR #828, opened 2026-08-12
feat(derive): let a variant hold its command in a `Box`

jdx:
Stacked on #827. This is the one you pointed at: *"the purpose behind the boxing was to avoid clap issues when commands get too big."*

An enum is as large as its biggest variant, so one thirty-flag subcommand makes every invocation of a mise-sized CLI move that much stack — and at that size clap itself starts to struggle, which is why mise boxes its largest commands.

```rust
#[derive(usage::Subcommands)]
enum Commands {
    Install(Box<Install>),   // boxed
    Nudge(Nudge),            // not, and the two work side by side
}
```

The box is read from the written form, as `Option<T>` is elsewhere. It is an artifact of *how the variant holds the struct*: the tables, the partial and `build` all speak to the struct itself, and the box goes back on at the single point a value is made. A test checks the emitted spec cannot tell the difference — boxing is not something a CLI has.

## What it unlocks

**The generated shadows are ordinary workspace members again.** `clippy::large_enum_variant` was the reason they sat outside the workspace, and boxing answers the lint instead of silencing it — so 12k lines of generated code are now linted like everything else, and there is still no `allow` anywhere in the repo.

## It is also slightly faster

Not the point, but worth recording: **51.1k instructions against 51.3k**, because moving a `Commands` as large as its biggest variant costs more than a malloc and a pointer move. Wall clock 2.05µs, ratio still 117×.

## One note on process

My first version shadowed `name` — the *command's* name — with the held type's name, so every command was suddenly called after its struct and no subcommand matched. The existing subcommand tests failed immediately, which is exactly the argument for having them; it took one diff to find.

---

> [!NOTE]
> **Medium Risk**
> Changes core derive codegen and regenerates very large benchmark shadows; parsing/spec behavior is covered by new conformance tests, but any derive bug affects all adopters.
> 
> **Overview**
> **`usage-derive` now accepts `Install(Box<Install>)` (and mixed boxed/unboxed enums).** The macro strips the `Box` for tables, partials, and `build`, then re-wraps with `Box::new` when constructing the enum variant. Types like `Box<cmds::Deep>` are parsed syntactically so generated code keeps the full path. Emitted KDL/spec is unchanged.
> 
> **Shadow crates are first-class workspace members again.** `xtask gen-shadow` emits boxed subcommand variants (matching mise/clap practice), so `large_enum_variant` no longer forces `exclude = ["benches/shadows"]`. Regenerated `mise` and `mise-clap` shadows are the bulk of the diff.
> 
> Conformance adds tests that boxed variants parse like unboxed ones and that boxing does not leak into the spec.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d9f107e1a649f1b8371884c1929f1b4a3a12c39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for boxed subcommand argument structures in generated CLI definitions.
  * Boxed and unboxed subcommands are both parsed correctly, including aliases and nested command paths.
  * Command names, aliases, options, and generated specifications remain unchanged.

* **Documentation**
  * Clarified boxed subcommand support and its effect on enum storage.

* **Tests**
  * Added end-to-end coverage for boxed variants and compatibility with existing unboxed variants.


jdx:
Correct, and it was a bad shortcut on my part: I read the box out of a *rendered* type string, and `type_name` keeps only the last segment. `Box<cmds::Deep>` came out as `Deep`, so the generated code named a type that is not in scope, and the derive failed to compile unless the struct happened to be imported.

Taken apart syntactically now — the last path segment's angle-bracketed argument — which keeps the path and reads `std::boxed::Box<T>` as well as `Box<T>`. `Box<T, A>` names an allocator this cannot reason about, so it is left alone and fails as the type the user wrote, with their name in the error rather than mine.

There's a test with the command declared in another module and referred to as `Box<cmds::Deep>`.

jdx:
The `unbox` fix is present here — `fn unbox` takes the type apart syntactically, so `Box<cmds::Deep>` keeps its path — and there is a test with the command declared in another module. That finding is a thread re-anchored after a force-push.

Worth knowing what the force-push was for, though: this branch had silently lost four of #826's fixes, because I restacked it by replaying a content patch captured against the older base, which reverts newer changes touching the same lines. Rebuilt on the corrected parent, with the fixes verified on the pushed ref.

PR #831, opened 2026-08-12
chore(bench): report the shadow comparison instead of gating it

jdx:
Your call, implemented: the shadow benchmarks are informative, not gated.

## Why they cannot be gated

tak gates everything in `tak.toml` at 1%, and the mise shadow is a fixture that **grows on purpose** — every property the derive learns to express adds to it. Teaching it command aliases (#827) added 91 names and 1.5%, and the gate reported a regression when it was the fixture getting bigger. Mounts and flatten will do the same. A gate that fires for that teaches people to ignore it, which costs more than it catches.

So `gate-baseline`, `gate-usage` and `gate-clap` leave `tak.toml`. usage's own benchmarks — `startup` and `markdown` — are gated exactly as they were before I added mine.

## What replaces them

`mise run perf:shadow`, which prints a table and the in-process wall clock, and which the pull-request workflow appends to its existing comment:

```
### Shadow comparison

| | usage | clap | ratio |
|---|---:|---:|---:|
| instructions, cold parse | 50,858 | 5,959,392 | 117x |

usage: argv -> struct                            2058 ns      2.06 µs
clap: build tree + parse -> struct             489708 ns    489.71 µs
clap: parse -> struct, tree reused              23610 ns     23.61 µs
clap: build tree only                          305085 ns    305.08 µs
```

The ratio is the thing to watch: it is a property of the parsers and holds steady while the fixture changes. Absent on a runner without valgrind, and the workflow treats that as nothing to add rather than as an error.

## One thing worth knowing about tak

My first instinct was `gate = false` on the bench in `tak.toml`. **tak accepts it and ignores it** — as it ignores any unknown key; I confirmed by feeding it `frobnicate = true`, which it also accepted silently. So that would have looked like a fix and gated anyway. If per-benchmark gating is something you'd want in tak itself, this is the use case for it.

---

> [!NOTE]
> **Low Risk**
> CI and benchmark configuration only; no runtime parser or product behavior changes, and shadow measurement failures cannot fail the PR gate.
> 
> **Overview**
> **Shadow mise-vs-clap benchmarks are no longer part of tak’s 1% gate.** `gate-baseline`, `gate-usage`, and `gate-clap` are removed from `tak.toml` so fixture growth (as the derive expresses more of mise’s spec) does not show up as parser regressions. Gated `startup` and `markdown` benches are unchanged.
> 
> **Reporting replaces gating.** New `tasks/perf-shadow.sh` (and `mise run perf:shadow` for local runs) measures cold-parse instruction counts via valgrind on `parse-n` vs `parse-n-clap`, prints a usage/clap/ratio table, and appends `time-parse` wall-clock output. The perf PR workflow runs this after `perf:record` with `|| true`, ships `shadow.md` in the artifact, and appends it to the sticky PR comment when present—missing valgrind or a failed run is ignored, not an error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ee5825d6cfe5ab7df4ca2c55b3fb779794dfcc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added optional shadow performance comparisons for parsing workloads.
  * Reports include instruction-count ratios and wall-clock benchmark results in Markdown format.
* **Improvements**
  * Pull-request performance comments can include shadow results when available.
  * Measurements continue gracefully when performance tools or individual measurements are unavailable.
  * Shadow benchmarks are tracked separately from gated benchmark checks.

### DIFF
--- a/.github/workflows/perf-pr.yml
+++ b/.github/workflows/perf-pr.yml
@@ -78,6 +78,14 @@ jobs:
       - name: Measure this pull request
         run: mise run perf:record
 
+      # Reported, never gated: the shadow is a fixture that grows as the derive learns to
+      # express more, so a gate on it fires for a bigger fixture rather than a slower
+      # parser. `|| true` because an informative table is not worth failing a run over.
+      # The script rather than the task: the release build is already done by the step
+      # above, and the task exists for running this by hand.
+      - name: Measure the shadow against clap
+        run: ./tasks/perf-shadow.sh /tmp/shadow-report.md || true
+
       - name: Find the base commit
         id: base
         env:
@@ -123,6 +131,9 @@ jobs:
           # a different failure from a regression and must not be reported as
           # one. The gate below maps them to different messages.
           cp /tmp/tak-gate-status /tmp/tak-out/status 2>/dev/null || echo 2 > /tmp/tak-out/status
+          # Absent when valgrind is missing or the measurement failed, which the reporting
+          # job treats as nothing to add rather than as an error.
+          cp /tmp/shadow-report.md /tmp/tak-out/shadow.md 2>/dev/null || true
           printf '%s %s\n' "${HEAD_SHA:0:12}" "${BASE_SHA:0:12}" > /tmp/tak-out/shas
 
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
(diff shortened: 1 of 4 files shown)


jdx:
Both right, and the first one had teeth.

**A partial report could have been published.** The workflow calls this with `|| true` on purpose — an informative table is not worth failing a run over — so a failure part way through the script would have left a truncated table, and the comment step would have published it as though it were a comparison. The report is built in a temporary file and moved into place only on success now. Checked by running it with a broken `PATH`: it exits 127 and leaves the destination exactly as it found it, so CI finds nothing to append rather than half a table.

**The fixture's size is out of the report.** "211 commands, 711 flags, four levels deep" is true today, and the shadow grows on purpose — the count would have gone stale the next time the derive learns something. `mise run gen-shadow` prints it, which is where a number about the fixture belongs.

PR #832, opened 2026-08-12
fix(spec): make the config block survive being written out

jdx:
First of the config v2 stack. Four defects in `config { prop … }`, all in the ground v2 builds on.

Nothing renders the block into docs yet (`lib/src/docs/models.rs:12` has `// pub config: SpecConfig,` commented out), but **the SDK generators do consume it** — and their workarounds are the evidence for two of these.

## `data_type` was parsed and never written

It survived exactly one hop:

```
prop "jobs" data_type="integer"   →  parse → Integer → write → (nothing) → parse → Null
```

`SpecDataTypes` gains a `Display` (the KDL spelling) and the writer emits it.

## `default` stored KDL source text, not a value

It kept `KdlValue::to_string()`, so `default=#true` was read as the five characters `#true` and `default="4"` as `"4"` *with* its quotes — and writing it out added another pair, one layer per round trip. The old snapshot in this file recorded exactly that (`default="#true"`, `default="4"`), which is how the bug stayed.

It's a typed `SpecConfigValue` now (bool/int/float/string), which is also what the v2 vocabulary needs for typed defaults. The **Python SDK generator had been undoing the damage by hand**:

```rust
SpecDataTypes::Boolean => match d.as_str() {
    "#true" | "true" => " = True".to_string(),      // the leaked KDL spelling
    …
SpecDataTypes::Integer | SpecDataTypes::Float => {
    let numeric = d.trim_matches('"');              // the added quotes
```

Both gone; it reads the value and renders by the declared `data_type` where there is one — which keeps the `data_type=float default="1.5"` case rendering as `1.5` rather than `"1.5"` (my first attempt at this regressed that, caught by its snapshot).

## A nested `prop` block parsed and lost its children

`prop "status" { prop "missing_tools" }` kept `status` and silently dropped the rest — the property loop reads properties only. Refused now, naming the dotted spelling to use instead.

## `merge` was first-wins per prop

Everything else in `Spec::merge` is other-wins (`merge_opt!`); config was the one place an `include` could add a prop but never correct one. Last-wins now, whole prop at a time.

## Verification

A test per defect, and **I checked each one fails when its fix is reverted** (four mutations, four failures). Full workspace green, clippy clean, `mise run render` produces no diff.

---

> [!NOTE]
> **Medium Risk**
> Changes spec parsing, serialization, and generated SDK code paths where bad defaults could previously break imports or silently corrupt values; new validation may reject previously tolerated specs.
> 
> **Overview**
> **Config `prop` defaults are typed values now**, not KDL source strings, so parse → write → parse keeps booleans, numbers, and strings without extra quoting layers. The writer also emits **`data_type`** again (via `Display` on `SpecDataTypes`), and defaults go out through typed KDL entries with **`string_entry`** for strings so control characters still reparse.
> 
> **Parsing is stricter:** nested `prop { … }` is rejected, defaults must fit the declared type (and finite floats / i64 range), with coercion after all props on the node are read. **`SpecConfig::merge`** is last-wins per prop, matching other `Spec::merge` behavior.
> 
> **SDK output:** shared **`escape_string_literal`** escapes control characters for Python/TS; the Python config dataclass renders defaults from **`SpecConfigValue`** (quoted strings only) instead of `#true`/quote-stripping hacks, with a test that malicious string defaults cannot become import-time code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 706e618de5dfa06d86c94291af850b85366969e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Improvements**
  * Configuration defaults now preserve native types, including booleans, integers, decimals, and strings.
  * Generated Python configuration values use correctly typed, safely formatted defaults.
  * String values are safely escaped across generated Python and TypeScript output.
  * Configuration serialization and round trips are more consistent and reliable.
  * Invalid nested configuration properties are rejected with clearer parse errors.
  * Repeated configuration properties now use the latest declaration.
  * Integer ranges and declared configuration types are validated more reliably.

### DIFF
--- a/lib/src/sdk/mod.rs
+++ b/lib/src/sdk/mod.rs
@@ -52,14 +52,38 @@ pub(crate) fn escape_py_docstring(s: &str) -> String {
     s.replace('\\', r"\\").replace(r#"""""#, r#"\"\"\""#)
 }
 
-/// Escape backslashes and double quotes for Python string literals.
-pub(crate) fn escape_py_string(s: &str) -> String {
-    s.replace('\\', r"\\").replace('"', r#"\""#)
+/// Escape a string for a double-quoted literal.
+///
+/// Backslashes, quotes, and control characters — the last of these because neither language
+/// can carry one literally inside a quoted string, so a value with a newline in it wrote a
+/// module that fails to import rather than one that says something wrong. Help text and
+/// config defaults both really do contain them.
+///
+/// Python and TypeScript spell all of these the same way, so one function serves both.
+fn escape_string_literal(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    for c in s.chars() {
+        match c {
+            '\\' => out.push_str(r"\\"),
+            '"' => out.push_str(r#"\""#),
+            '\n' => out.push_str(r"\n"),
+            '\r' => out.push_str(r"\r"),
+            '\t' => out.push_str(r"\t"),
+            c if c.is_control() => out.push_str(&format!("\\x{:02x}", c as u32)),
+            c => out.push(c),
+        }
+    }
+    out
 }
 
-/// Escape backslashes and double quotes for TypeScript string literals.
+/// Escape a string for a Python literal.
+pub(crate) fn escape_py_string(s: &str) -> String {
+    escape_string_literal(s)
+}
+
+/// Escape a string for a TypeScript literal.
 pub(crate) fn escape_ts_string(s: &str) -> String {
-    s.replace('\\', r"\\").replace('"', r#"\""#)
+    escape_string_literal(s)
 }
 
 /// A simple code writer with indentation management.
--- a/lib/src/sdk/python/mod.rs
+++ b/lib/src/sdk/python/mod.rs
@@ -138,22 +138,24 @@ fn render_types(spec: &Spec, package_name: &str, source_file: &Option<String>) -
             // none is declared — a spec may say `data_type=float default="1.5"`, and the
             // declaration is what the generated field is typed as. No `#true`-spelling or
             // quote-stripping to undo any more: the spec keeps values, not source text.
-            let default = match (&prop.default, prop.data_type) {
-                (None, _) => String::new(),
-                (Some(v), SpecDataTypes::Boolean) => match v {
-                    SpecConfigValue::Bool(false) => " = False".to_string(),
-                    SpecConfigValue::String(s) if s == "false" => " = False".to_string(),
-                    _ => " = True".to_string(),
-                },
-                (Some(v), SpecDataTypes::Integer | SpecDataTypes::Float) => {
-                    format!(" = {}", v.display())
-                }
-                (Some(SpecConfigValue::Bool(b)), SpecDataTypes::Null) => {
+            // A bare Python literal only for a value that *is* one. Anything else is a
+            // quoted string, escaped.
+            //
+            // This is generated code that gets imported, so text from the spec must never
+            // reach it unquoted: a `data_type="integer"` whose default is the string
+            // `__import__("os").system(…)` would otherwise be written as an expression and
+            // run on import. Numeric strings are read as numbers when the spec parses, so
+            // by here a `String` is a string.
+            let default = match &prop.default {
+                None => String::new(),
+                Some(SpecConfigValue::Bool(b)) => {
                     format!(" = {}", if *b { "True" } else { "False" })
                 }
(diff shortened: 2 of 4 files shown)


jdx:
Three real ones and one that doesn't reproduce. The Critical was the important one, and it was a consequence of my own fix in this PR.

**String default emitted as a Python expression** (CodeRabbit, Critical). Rendering by the declared `data_type` meant a `data_type=integer` whose default was arbitrary text wrote that text into `types.py` unquoted — and `types.py` gets imported, so `default="__import__('os').system('touch /tmp/pwned')"` would have run on import. Fixed at both ends:

- the spec reads a default as the declared type *only where the text is that type*, so `data_type=float default="1.5"` becomes `Float(1.5)` and `data_type=integer default="__import__('os')"` stays a `String`;
- the generator emits a bare literal only for a value that *is* one, quoting and escaping anything else.

There's an end-to-end test with that exact payload asserting the output contains no `= __import__` and does contain the quoted form, plus a default containing `"` staying escaped.

**Out-of-range integer silently dropped** (greptile + CodeRabbit, same bug). KDL parses `i128`; `try_from(…).ok()` turned an out-of-range integer into "no default", losing a number somebody wrote — and then the writer and the SDKs both reported the property as having none. It's a parse error now (`config default does not fit in a 64-bit integer`), and `None` means an explicit `#null`.

**Whole floats becoming ints** (Bugbot, Medium) — **does not reproduce.** I measured it before and after:

```
default=1.0   → written: default=1.0     → back: Float(1.0)
default=2.50  → written: default=2.5     → back: Float(2.5)
default=1e3   → written: default=1000.0  → back: Float(1000.0)
```

kdl 6.5's float rendering already keeps the decimal point. I had written a `value_repr` fix for this and then found the mutation test *passed* without it — which is what told me the fix was unnecessary rather than untested. It's removed; the test stays as a pin, labelled as one, with the reasoning in a comment so nobody re-adds it.

All four fixes mutation-checked: revert the fix, the test fails (except the float one, which is how I caught it).

jdx:
No, I had not seen that one — it landed on the post-rebase commit, after my last sweep. It was right, and so were the two other findings I had missed on this PR. All three are fixed in the amended head.

**String defaults skipped the round-trip helper.** Exactly as described, and I reproduced it before fixing: a default containing a colour escape wrote a spec that no longer parses.

```
config {
    prop prompt default="a[0mb"   ← the escape went out literally
}
```

`string_entry` exists for this and the typed-default writer built its entry by hand. It goes through the helper now, as does the prop's own key.

**Non-finite float defaults.** `#inf` and `#nan` parse in KDL and nothing downstream can carry one. Measured, not assumed: `usage g json` reported `"default": null` — the default silently gone — because `serde_json` writes a non-finite float as `null`, and the Python generator emitted a bare `inf`, which is a `NameError`. Refused at parse now, by both roads (`default=#inf` and `default="inf"` on a float-typed prop).

**Out-of-range string defaults, and the wider case behind it.** Greptile's example was a 20-digit number in quotes on an integer-typed prop, which slipped past the range check the same number unquoted would have hit. The narrow fix would have been a range check on the string; instead a string the declared type cannot read is now refused outright.

That reverses a decision I made deliberately last time — keeping it as a string, so nothing would treat its text as a number — so it is worth saying why. Keeping it left the spec asserting two contradictory things, and the generator then wrote an `int` field defaulting to a quoted sentence. Refusing is safe *and* honest. I checked it refuses nothing anybody has written: across mise's 280 settings, the largest registry in the fleet, every string default reads as its declared type.

The security test moves rather than disappears. Its original case no longer parses, but a *string*-typed default legitimately holds arbitrary text, so the generator must still quote — two defences, and that is what it now tests.

One note on method: `UsageErr::InvalidInput` renders as "Invalid usage config" whatever went wrong, so the new tests assert on the diagnostic's label. A test that only checks `is_err()` passes just as happily when the spec was refused for an unrelated reason.

jdx:
Two more, both real.

**Control characters broke the generated Python.** `escape_py_string` handled backslashes and quotes only, so a default containing a newline was written into `types.py` unescaped and the module fails to *import* — a worse failure than one that merely says something wrong. Fixed in the shared escaper, which `escape_ts_string` was also using, so the TypeScript side and every other string these generators emit (help text included) are covered by the same change.

**A native default under a declared `string`.** `type="string" default=4` left the value a number, so the field was typed `str` and defaulted to `4`. Coercion now runs in that direction too. Every other declared type is a number or a boolean, where a bare value is already the right shape.

PR #833, opened 2026-08-12
feat(derive): let a field be the type it means

jdx:
The largest thing standing between the derive and mise. Independent of #827–#829, so it can go in either order.

mise names `PathBuf` **227 times** and its tool-version type **83 times** in `src/cli/`. A derive that only holds `String` cannot hold mise's commands, however faithfully it expresses mise's spec.

```rust
#[usage(short = 'j', long)] jobs: Option<usize>,
#[usage(long)]              out: Option<PathBuf>,
#[usage(long)]              port: u16,          // bare: the type says it is required
#[usage(long, var)]         tool: Vec<Tool>,    // any FromStr of your own
#[usage(long, var)]         search: Option<Vec<PathBuf>>,
```

`Option<Vec<T>>` earns its place: "never given" and "given nothing" are different answers, and mise's root asks for that distinction three times.

## What keeps it cheap

The layering does not move. **Binding still collects text** — a word's meaning is not a question about where it lands — and conversion happens once, where the struct is built. So defaults, `env`, `choices` and the variadic bounds all keep working on text, untouched.

**Measured on the same fixture as main: 40,129 instructions against 39,963.** +166, or 0.4%, nearly all of it bookkeeping the identity conversion cannot avoid.

That comparison is worth a note. On #827/#828 this parse measures ~50.9k, and I nearly reported this branch's 40k as a 20% win — it is not, it is a *smaller shadow*, because main has neither those branches' 91 aliases nor their boxed variants. Same trap the informative-benchmark change was made for; the honest number needs the same fixture on both sides.

## Errors say what the type says

```rust
Err(Error::InvalidValue(bad)) => {
    bad.name;    // "jobs"
    bad.value;   // "lots"
    bad.reason;  // "invalid digit found in string"
}
```

Better than anything this crate could invent about someone else's type: `u16` explains "number too large than can fit", and a tool argument explains itself. `Error` gained one variant and **lost `Copy`**, since that variant owns two strings — but the payload is boxed, so `Error` is still **40 bytes** and the `Result` it rides in on the hot path did not grow. I checked that rather than assuming it.

A type no word could become is a compile error naming *that* type:

```
error[E0277]: the trait bound `HashMap<String, String>: FromStr` is not satisfied
```

## One gap, deliberate and documented

A word reaches a field through `from_utf8_lossy`, so a `PathBuf` holding a path that is not UTF-8 gets replacement characters rather than bytes. Rare, and wrong when it happens. The fix is for the partial to hold `OsString` and let `build` decide — exact for `PathBuf` and `OsString`, an error for `String` rather than a silent mangling — which is the next PR rather than a bigger one now.

Also in here, from jdx's note that mise carries hacks purely to work around clap: PLAN.md now lists what adoption should let mise **delete**, checked against mise rather than assumed. Two of the five do not survive contact — `command_effects.rs` argues its own case for staying one central list, and `mise-extra.usage.kdl` turns out to be a docs link template rather than a workaround.

---

> [!NOTE]
> **Medium Risk**
> Public `usage_argv::Error` API changes (`Copy` removed, new variant) and derive codegen affects all parsers, but behavior is covered by new conformance tests and conversion stays off the successful parse hot path.
> 
> **Overview**
> **Typed CLI fields** move beyond string-only binding: a field can be any `T: FromStr`, plus `Option<T>`, `Vec<T>`, and `Option<Vec<T>>` (so “flag never given” vs “given but empty” stays distinct). The argv layer still binds text; **`usage-derive` converts when building the struct**, with a fast path that moves `std::string::String` without re-parsing.
> 
> Failed conversions surface as **`Error::InvalidValue`** (boxed **`InvalidValue`** with field name, raw text, and the type’s `Display` message). **`Error` no longer implements `Copy`**; the new variant is boxed so the enum stays **40 bytes**.
> 
> **Subcommand `Args` builds** use the same **`field_final`** logic as the root CLI (fixing typed fields on nested commands).
> 
> **`PLAN.md`** marks typed values done, notes the next step (non–UTF-8 via `OsString` in partials), and adds a mise adoption section on what clap workarounds can be removed vs what should stay.
> 
> **`conformance/tests/typed.rs`** covers numeric/path/custom types, `Option<Vec<_>>`, spec/KDL unchanged by Rust types, and subcommand conversion errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0501bb551c09766e5075e0df8938ff6fa1fcc03c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added type-driven conversion for CLI values, supporting numeric, path, custom `FromStr`, optional, and repeated fields.
  * Added support for optional collections, distinguishing absent values from empty collections.
  * Conversion failures now report the field, rejected value, and reason.

* **Documentation**
  * Updated documentation for supported typed-value patterns and current non-UTF-8 handling limitations.

* **Tests**
  * Added comprehensive coverage for typed fields, collections, conversion errors, and numeric validation.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -115,10 +115,17 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
 - [x] **Subcommands in the derive** — an enum of variants, each holding the struct
       that declares its flags, nested to any depth. A command is selected by its
       position in the parent's table, found from the address the parser hands back.
-- [ ] **Typed values** — fields are text today (`String`, `Option<String>`,
-      `Vec<String>`, `bool`, counting integers). Parsing into other types needs an
-      error type for value conversion, which is also where `env`, required-ness,
-      and `choices` get enforced — see the post-binding layer below.
+- [x] **Typed values** — a field is any `T: FromStr`, plus `Option<T>`, `Vec<T>` and
+      `Option<Vec<T>>`, which is what lets a command struct hold the `PathBuf` mise names
+      227 times and the tool-version type it names 83 times. Binding still collects text;
+      the conversion happens where the struct is built, and a value that will not convert
+      reports the text and the type's own message. `Error` grew one boxed variant and lost
+      `Copy`, and stayed 40 bytes, so nothing on the hot path grew.
+- [ ] **Values that are not valid UTF-8** — a word reaches a field through
+      `from_utf8_lossy`, so a `PathBuf` holding a non-UTF-8 path gets replacement
+      characters instead of bytes. The partial should hold `OsString` and let `build`
+      decide: exact for `PathBuf` and `OsString`, an error for `String` rather than a
+      silent mangling. Small, and the next thing.
 - [ ] **`usage-derive` v1** — everything mise needs: constraints
       (`requires`/`conflicts`/`overrides`/`required_unless`), `var`, `count`,
       `env`, defaults, delimiters, the `double_dash` modes, global flags, flatten,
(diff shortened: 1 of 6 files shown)


jdx:
Real, and I priced the alternative rather than arguing about likelihood.

If an adopter writes `use my_crate::String` over a `FromStr` type of their own, that field is *moved* rather than converted and their crate fails to compile. A macro cannot resolve a name, so recognising `String` by how it is written is the only option available — the question is what the fix costs.

Converting unconditionally instead: **3 allocations become 5, and the parse goes 40,129 → 41,039 instructions (+2.3%)** on `mise use -g node@20`. That is the price of protecting against someone shadowing `String`, which is a thing nobody should do and nobody in mise does.

So the fast path stays, the hazard is documented alongside the other known gaps, and it **stops being a choice at all in the next PR**: once the partial holds bytes rather than text, a `String` field converts like every other type and there is no identity case left to recognise. That change is already planned for the non-UTF-8 fix, and this is a second reason for it.

Measuring it turned up something worth having anyway. The collecting path was rebuilding a `Vec<String>` element by element — allocating a second `Vec` to hold what the first already held, one allocation per collecting field, and mise's commands collect a lot. It is moved whole now, which is what took the count back to 3.

PR #834, opened 2026-08-12
feat(derive): declare the words a value may be

jdx:
Stacked on #833. jdx: *"yes let's do the ValueEnum derive, mise has a lot of those"* — nine, to be exact.

```rust
#[derive(usage::ValueEnum)]
enum Interpreter {
    Bash,
    Zsh,
    Fish,
    #[usage(name = "pwsh")]
    PowerShell,
}

#[usage(short = 's', long, value_enum)] shell: Option<Interpreter>,
#[usage(long, var, value_enum)]         also: Vec<Interpreter>,
```

## The point is that the list exists once

A value enum is a list of words that a CLI normally writes twice — once as the type, once as whatever tells the framework about it. Here the type is the only copy, and everything reads from it:

- the **emitted spec** carries the words, so `usage g markdown` and the completions offer exactly what the parse accepts;
- the **check** that rejects a wrong value reads the same `CHOICES`;
- the generated **`FromStr`** is what typed values already call, so a value enum is not a special path — it is a type whose values happen to be listed.

A wrong word gives `InvalidChoice { name: "shell", choices: ["bash", "zsh", "fish", "pwsh"] }` — the flag and the expected words — rather than a conversion error mentioning a type the user never named. And converting one by hand still gets a message with the words in it, since that check belongs to the parser rather than the type.

## What it refuses

Each of these is either one list in two places or a word nothing answers to:

| declaration | why not |
| --- | --- |
| `value_enum` with `choices(…)` | the same list twice, and they can drift |
| `value_enum` on a `bool` | a switch has no value for a word to be |
| a variant holding fields | a value is one word; there is nothing to build fields from |
| two variants with the same `name` | one of them could never be reached |
| an enum with no variants | accepts no value at all |

Also in here: the three review findings from #833 (a subcommand's values were not being converted, the `String` fast path matched a last path segment, and the `Display` requirement on a conversion error is now documented) — those are on #833 itself; this branch just sits on top.

---

> [!NOTE]
> **Low Risk**
> Additive derive/macro and trait API with conformance tests; no changes to the hot argv binding path beyond existing post-binding choice checks.
> 
> **Overview**
> Adds **`#[derive(usage::ValueEnum)]`** and a **`value_enum`** field attribute so allowed flag/arg words live on the enum type instead of a duplicate `choices(...)` list on each field.
> 
> The derive emits **`ValueEnum::CHOICES`** (for spec/help/completions) and **`FromStr`**, with kebab-case variant names and optional **`#[usage(name = "...")]`**. **`usage-argv`** gains a **`ValueEnum`** trait; codegen wires **`value_enum`** fields to that const for metadata and post-parse validation, so invalid tokens surface as **`InvalidChoice`** with the expected words rather than a type conversion error.
> 
> Compile-time guards reject mixing **`value_enum`** with **`choices`**, using it on bool/count fields, non-unit variants, duplicate names, **`cfg`** on variants, and empty enums. Conformance tests cover parse, spec emission, and manual **`FromStr`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1fa8cef80e0ff4f30ddd107681b04d7aa369fce1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -121,6 +121,11 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       the conversion happens where the struct is built, and a value that will not convert
       reports the text and the type's own message. `Error` grew one boxed variant and lost
       `Copy`, and stayed 40 bytes, so nothing on the hot path grew.
+- [x] **Enumerated values** — `#[derive(usage::ValueEnum)]` on an enum of bare variants
+      gives the words a value may be, and a field says `value_enum` to use them. mise has
+      nine of these. The list is declared once, on the type: the spec, the help, the
+      completions and the check that rejects a wrong word all read it from there, so none of
+      them can drift from the type the way a second list on the field would.
 - [ ] **Values that are not valid UTF-8** — a word reaches a field through
       `from_utf8_lossy`, so a `PathBuf` holding a non-UTF-8 path gets replacement
       characters instead of bytes. The partial should hold `OsString` and let `build`
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -760,6 +760,17 @@ fn quoted(value: &str) -> String {
 /// generated parse needs to name the type that accumulates a subcommand's values
 /// while parsing, and cannot know which module the derive put it in — an
 /// associated type is how it names it regardless.
+/// A type whose values are a fixed set of words.
+///
+/// What a CLI calls an enum: `--shell bash`. The words are what the spec lists as
+/// `choices`, so declaring them once on the type keeps help, completions and the check that
+/// rejects a wrong value reading from the same place — rather than a list in an attribute
+/// that has to be kept in step with the type by hand.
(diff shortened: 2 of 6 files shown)


jdx:
Correct — and there is no way to emit what a `cfg`-ed-out variant would need.

The `FromStr` arm would name a variant that is not there, and `CHOICES` is a `const` array: array literals take no attributes on their elements, so the word list cannot have a matching hole. Emitting the arms conditionally and leaving the list intact would be worse than the compile error — it would offer a word nothing answers to.

So it is refused, with a message saying why and what to do instead (`cfg` the whole enum), and a test for it. I checked mise before choosing: nine value enums, no `cfg` on any variant, so nothing needs this today. If something does, the fix is to make the word list a function rather than a `const`, which is contained.

PR #835, opened 2026-08-12
feat(spec): the config vocabulary

jdx:
Stacked on #832. What a `config` block can say — so docs, a JSON schema, completions and a resolver come from one declaration instead of the five hand-written copies the fleet has now.

## Three node kinds

```kdl
config {
    // Kinds usage knows nothing about, with the metadata docs need to describe them.
    source "git" name="git config" doc_hint="git config `{key}`" set_hint="git config {key} {value}"
    source "pkl" name="hk.pkl"

    // The rc-style chain, ascending precedence.
    file "~/.config/hk/config.pkl" scope="global"
    file "hk.pkl" findup=#true

    prop "jobs" type="uint" default=0 default_note="0 = auto-detect" help="…" since="1.0.0" {
        cli "--jobs" "-j"
        env "HK_JOBS" "HK_JOB"        // aliases, highest first
        source "git" "hk.jobs"        // its key in a declared kind
        example "hk check --jobs 4"
    }
    prop "exclude" type="list<string>" merge="union" { default "target" "node_modules" }
    prop "stash" type="string" { choices { choice "git" help="Use `git stash`" } }
    prop "trusted" type="bool" scope="global"       // never from a project file
    prop "ci" type="bool" hide=#true scope="env" { x "mise.rust_type" "BoolOrString" }
    prop "old.key" deprecated="Use new.key" renamed_to="new.key" \
        deprecated_warn_at="2026.12.0" deprecated_remove_at="2027.12.0"
}
```

`source` is what lets docs say "settable with `git config hk.check`" without usage having any idea what git is. `scope="global"` is not decoration — mise treats "a checked-in file must not be able to change this" as a security property, so it is declared rather than left to each tool.

## Types

```
base := bool | string | int | uint | float | path | url | duration | object
type := base | list<type> | set<type> | map<base, type> | option<type> | type "|" type
```

An unknown name is **kept verbatim** rather than refused — that's the type-level escape hatch, so a spec naming a type only its own tool understands still loads everywhere. Old spellings (`data_type=`, `boolean`, `integer`, `number`, `usize`, `array<…>`, `optional<…>`) are accepted and normalized.

## `x` extensions

`x "ns.key" value`: preserved in order, written back out, in `usage g json`, interpreted by nothing here. Where `mise.rust_type`, `mise.parse_env` and `aube.npm_shared` live. This is the seam that lets a CLI with special rules describe its settings without usage having to model those rules.

## Verification

- **The whole vocabulary round-trips**, field by field, in one test — anything the spec cannot write out is something a tool would lose by saving its own file.
- **The parsed model is committed as a snapshot** — the artifact `usage g json` hands downstream, and what a port in another language diffs against.
- **Every KDL example on the reference page is parsed by a test.** That page previously documented `file`, `findup`, `default "k" "v"`, `alias` and `config_file` — *none of which the parser has ever accepted* — while never mentioning `prop`, the one thing it did. It's in the site nav. It cannot drift like that again.
- Unknown vocabulary is refused rather than half-read (five cases tested), which is what `min_usage_version` is for.

One finding for elsewhere: running that docs test against the other reference pages reports five failures. Four are catalogue blocks — lists of alternative spellings that aren't specs (one even puts an argument after a variadic, which the parser rightly refuses). The fifth is real: **`flag "flag1"` in `cmd.md` is missing its dashes** and the parser rejects it. Left for its own change rather than smuggled in here.

---

> [!NOTE]
> **Medium Risk**
> Large expansion of the spec parser and serialized config shape; behavior changes (strict errors, merge rules for files/sources) could break hand-written specs that relied on old permissive or undocumented behavior, though compatibility aliases and tests aim to limit that.
> 
> **Overview**
> Adds a full **config** vocabulary to usage specs so settings, file chains, and external source kinds can be declared once for docs, JSON export, and resolvers.
> 
> **`config` block** now supports **`source`** (opaque kinds like git/pkl with doc hints), **`file`** (precedence chain with `findup`, `scope`, `format`), and richer **`prop`** nodes: `type` grammar, `cli`/`env` aliases, bindings, `choices`, `merge`/`scope`, deprecation/renames, `x` extensions, and list defaults.
> 
> Introduces **`SpecConfigType`** (`list<>`, `map<>`, unions, legacy synonym normalization) alongside the old five-value `data_type`, with stricter parsing (unknown keys refused, no silent child drops, union defaults not coerced via legacy type).
> 
> **`docs/spec/reference/config.md`** is rewritten to match the parser; **`lib/tests/docs_examples.rs`** parses every KDL example on that page. Generated CLI JSON gains empty `sources`/`files` on `config`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ae32fe0ed7596b9554551b4d110cb402b5ab5cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/cli/reference/commands.json
+++ b/docs/cli/reference/commands.json
@@ -1153,7 +1153,9 @@
     "examples": []
   },
   "config": {
-    "props": {}
+    "props": {},
+    "sources": {},
+    "files": []
   },
   "version": "5.1.0",
   "usage": "Usage: usage-cli [OPTIONS] [COMPLETIONS] <COMMAND>",
--- a/docs/spec/reference/config.md
+++ b/docs/spec/reference/config.md
@@ -108,7 +108,7 @@ And as child nodes, for anything multi-valued or long:
 | `cli "--jobs" "-j"`               | flags that set it                                             |
 | `env "A" "B"`                     | environment variables, highest precedence first               |
 | `source "git" "a.b"`              | its keys in a declared source kind                            |
-| `default "a" "b"`                 | a list default                                                |
+| `default "a" "b"`                 | a list default, values typed as written (`default 80 443`)    |
 | `long_help "…"`                   | the long form, when a raw string reads better than a property |
 | `example "…"`                     | one invocation worth showing                                  |
 | `choices { choice "a" help="…" }` | the values it accepts, each with its own help                 |
--- a/docs/spec/reference/config.md
+++ b/docs/spec/reference/config.md
@@ -1,53 +1,172 @@
 # Configuration
 
-The default priority for configuration properties in usage is the following:
-
-- CLI flag (e.g. `--user alice`)
-- Environment variable (e.g. `MYCLI_USER=alice`)
-- Config file (e.g. `~/.mycli.toml`)
-- Default value
-
-## Environment Variables
-
-TODO
-
-## Config Files
+A `config` block describes a CLI's settings: what they are called, what they hold, where a
+value can come from, and what to say about them in documentation. It is what
+`usage g markdown` renders as a settings reference, what a JSON schema for the config file
+is generated from, and what a runtime resolves values against.
 
 ```kdl
 config {
-    // system
-    file "/etc/mycli.toml"
-    file "/etc/mycli.json"
-
-    // global
-    file "~/.config/mycli.toml"
-    file "~/.config/mycli.json"
-
-    // local
-    file ".config/mycli.toml" findup=#true
-    file ".config/mycli.json" findup=#true
-    file ".mycli.dist.toml" findup=#true
-    file ".mycli.dist.json" findup=#true
-    file ".mycli.toml" findup=#true
-    file ".mycli.json" findup=#true
-    file ".myclirc" findup=#true format="ini"
-
-    // e.g.: .mycli.dev.toml, .mycli.prod.toml
-    file ".mycli.$MYCLI_ENV.toml" findup=#true
-
-    default "user" "admin"
-    default "work_dir" "/tmp"
-    default "yes" false
-
-    alias "user" "username"
+    prop "jobs" type="uint" default=0 help="Number of parallel jobs" {
+        cli "--jobs" "-j"
+        env "MYCLI_JOBS"
+    }
 }
 ```
 
-## Alias Config Keys
+Keys are dotted paths. `prop` blocks do not nest — write `prop "status.missing_tools"`
+rather than a `status` block containing a `missing_tools` one — because one spelling keeps
+merging, ordering and round-tripping unambiguous, and a schema generator can always
+re-nest on the dots.
 
-Config keys can be aliased to other keys. This is useful for backwards compatibility.
+## Where values come from
+
+Highest precedence first, and only the layers a CLI actually has:
+
+- the command line
+- the environment
+- config files, nearest first, with a `.local` variant outranking its base
+- the user's own configuration
+- files installed for the whole machine
+- the default a `prop` declares
+
+A property can name its sources explicitly; the order they are written in is the order they
+are consulted.
 
 ```kdl
-config_file ".mycli.toml" findup=#true
-config_alias "user" "username"
+prop "check" type="bool" {
+    cli "--check"              // flags that set it, as declared elsewhere in this spec
+    env "HK_CHECK" "HK_LINT"   // several: aliases, highest precedence first
+    source "git" "hk.check"    // a source kind declared below
+}
 ```
+
+## `source` — kinds usage does not know about
+
+usage reads the command line, the environment, and config files. Everything else — a git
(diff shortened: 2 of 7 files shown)


jdx:
All three findings were real, and there was a fourth hiding behind the third. Fixed in b41d239.

**Config merge drops new fields** — confirmed, and worse than described: `is_empty` gating the writer meant a `config` block declaring *only* `file` locations was reported empty and omitted entirely. `merge` now merges `sources` per kind (a set, like props) and replaces `files` when the other spec declares any. Files are an ordered precedence chain rather than a set, and interleaving two chains has no meaningful answer — in the case `include` exists for, only one side declares them at all. That reasoning is in the code.

**Malformed types normalize silently** — confirmed for all three inputs, and `int>` was the worst of them: it became `Base::Custom("int>")`, a "type" every consumer treats as a string. `Custom` is the escape hatch for names a *newer* usage understands; it must not swallow an expression that failed to parse. Empty union members are no longer dropped, and a base name is refused if it is empty or contains any of `< > , |`. That last check found the fourth case: `map<,string>` was parsing as a map keyed by `Custom("")`.

**Non-string bindings become strings** — confirmed. `env`, `cli`, `source` and `example` hold names, so `string_args` now uses `ensure_string` like the rest of the parser instead of `to_string()`.

The one place where "convert rather than refuse" was actually wrong in the *other* direction was the list default: `default 80 443` for a `list<int>` is two numbers, and forcing strings there would make the JSON schema in #839 describe the setting wrongly. So `default_list` is now typed like the scalar `default` is, which the schema generator picks up for free — it emits `[80, 443]` rather than `["80", "443"]`.

Five tests, seven mutations, each one verified to fail with its fix reverted. The three descendant PRs are rebased; the docs model needed one follow-on change for the typed list, folded into #837.

jdx:
Also fixed the builder inconsistency Bugbot flagged (`Builder env skips envs list`). It was real: `env` and `envs` are two spellings of one thing — `env="X"` and `env "A" "B"` — and the parser reconciles them so a consumer can read either field, but `SpecConfigProp::env()` set only `env`. A spec assembled in Rust therefore serialized with an empty `envs`, and every reader of that field saw a setting no variable could set.

`env()` now maintains both, and calling it more than once means the alias list, highest precedence first — the same thing `env "HK_JOBS" "HK_JOB"` means in a spec. One test covers all three construction paths (builder, property, child list) so the invariant cannot hold in one and not the others, mutation-verified.

The `Config merge drops sources and files` finding is the same one Greptile raised and is fixed in the commit above.

jdx:
Three more findings from the later review round, all real, all fixed in the amended head.

**A union claimed a legacy type it does not have.** `data_type_of` ran unions through `simplified()`, so `bool|string` reported `Boolean` — and that legacy field decides how a default is read, so `type="bool|string" default="true"` came back as a *boolean*, contradicting the very `value_type` it was derived from. A union has no counterpart among the old five values, so it maps to `Null` like every other composite.

**`source` and `file` ignored their children.** They validated their properties and never looked at a child block, so nested vocabulary was dropped in silence — the one thing this block is strict about not doing, and which `prop` already refused. Both now refuse a block outright.

**The two env spellings could disagree.** The sync ran only when one side was empty, so a prop writing both `env="FIRST"` and `env "SECOND"` left `env` and `envs` asserting different things — `usage g json` exposing the pair, and a writer that chooses between them by list length, so a round trip dropped a value. `env=` is now simply shorthand for a one-element list: both forms feed `envs` in written order and `env` is always its first entry, so they agree by construction rather than by a sync that might not run. Tested through all four construction paths — builder, property, child list, and both at once — plus a round trip.

The builder inconsistency from the earlier round is fixed here too.

jdx:
Two more, both real.

**`#null` in an extension did not round-trip.** `x` nodes promise to come back out exactly as they went in, and `#null` had nowhere to come back to — it was stored as an empty string and written as `""`, so tool-private metadata was altered by saving the file. Refused now, which is the only honest answer when the value has no representation.

**A second `env` or `cli` node dropped the first.** `example` and `source` accumulate across nodes; these two assigned, so a spec writing them on two lines silently lost the first line's aliases. They extend now, like their neighbours.

jdx:
Right, and it is the case I missed when I made `env` and `cli` accumulate last round: `default` still cleared the list before filling it, so a prop declaring its list default over two lines kept only the second. It accumulates now, like its neighbours. Mutation-verified.

jdx:
Right — and it is the same class as `source` and `file`, which I fixed a round ago with a helper I then failed to apply to the one node left. A `choice` is properties-only too, so a nested block is refused rather than dropped, through the same `refuse_children`. Mutation-verified, and the existing test grew a third case so the three nodes are checked together.

PR #837, opened 2026-08-12
feat(docs): render the config block

jdx:
Stacked on #835. The settings a spec declares now reach documentation — `lib/src/docs/models.rs:12` had `// pub config: SpecConfig,` commented out, so a config block could be written, read, round-tripped, and never appear anywhere a user would look.

## What it looks like

```markdown
# Configuration

Read from, in ascending precedence — the last one that names a setting wins:

- `~/.config/hk/config.toml` (global)
- `hk.toml` — and in every parent directory

## `exclude`

- **type**: `list<string>`
- **default**: `target, node_modules`
- **merged**: values from every source are combined
- **set with**: `HK_EXCLUDE`

Patterns to skip

## `old`

::: warning Deprecated
Use jobs instead. Removed in 2027.12.0.
:::

## Performance

### `jobs`

- **type**: `uint`
- **default**: 0 = auto-detect
- **since**: 1.0.0
- **set with**: `--jobs`, `-j`, `HK_JOBS` (or `HK_JOB`), git config `hk.jobs`
```

That last line is the point of the `source` declarations: **"git config `hk.jobs`" is rendered from the kind's `doc_hint`**, so a page describes a git config, a pkl file or an `.npmrc` without usage knowing what any of them are.

## Decided in the model, not the templates

So that markdown, man pages and anything later all agree: hidden props are dropped once, props are grouped by `help_heading` (reusing the helper flags already use, unheaded first), and the "set with" line is worded in one place.

## Where it appears

- `usage g markdown`: after the commands in single-file mode; a `settings.md` page in `--multi` mode.
- `usage g manpage`: a `CONFIGURATION` section after the commands, before the author.

Both only when there is something to say — a CLI with no settings gains neither an empty section nor an empty file, and there's a test for each.

The man page is deliberately terser than the web page: type, default, how to set it, deprecation. Prose belongs on the web page, and markdown backticks would be literal in a terminal — a test asserts none survive. Heading levels never skip: a prop under a heading is one level deeper, a prop without one sits directly under the section.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to documentation generation and templates; no runtime CLI behavior or security-sensitive paths, with collision failures failing closed before partial writes.
> 
> **Overview**
> **Config blocks from usage specs now show up in generated docs** instead of being dropped from the docs model.
> 
> Markdown gets a **Configuration** section (single-file output after commands; dedicated `configuration.md` in `--multi` mode) with file precedence, grouped settings, facts, choices, and “set with” lines built from CLI/env/custom `doc_hint` sources. The multi-file index links to that page when config content exists. **`--multi` refuses to run** if a visible top-level `configuration` command would collide with `configuration.md`, and does so before writing any files.
> 
> Man pages gain a matching **CONFIGURATION** section (files-only configs included, props grouped by `help_heading`, terminal-friendly facts without markdown backticks).
> 
> Rendering fixes keep config in sync with other docs: config is re-derived from `raw_config` so options like `--replace-pre-with-code-fences` apply, and templates gate empty config so no blank sections or files appear.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d84af71a276464aca361df9b65b39d352a6033a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added configuration documentation to generated Markdown and manpage output.
  - Documented configuration files, lookup precedence, property groups, types, defaults, sources, choices, scopes, and deprecation notices.
  - Added links to the generated configuration page when configuration details are available.
- **Bug Fixes**
  - Empty configurations no longer generate blank pages or configuration sections.
  - Hidden configuration properties remain excluded.
  - Configuration pages now take precedence when their path conflicts with a command page.
- **Documentation**
  - Improved formatting and heading structure for configuration content.

### DIFF
--- a/cli/src/cli/generate/markdown.rs
+++ b/cli/src/cli/generate/markdown.rs
@@ -65,6 +65,22 @@ impl Markdown {
         }
         if self.multi {
             ctx = ctx.with_multi(true);
+            // Before anything is written, so a refusal does not leave a half-built directory.
+            //
+            // Refused rather than warned about: writing anyway destroys the command's own
+            // documentation, and a warning printed into a docs build is a warning nobody reads.
+            // No CLI in the fleet has a `configuration` command — they call it `config` — so
+            // this asks the one author who ever hits it to pick a name, rather than quietly
+            // losing a page for everybody who does not.
+            if !ctx.render_config()?.trim().is_empty() {
+                if let Some(cmd) = ctx.config_page_collision() {
+                    miette::bail!(
+                        "the `{cmd}` command's page and the settings page would both be written \
+                         to {}; rename the command, or hide it, to generate both",
+                        ctx.config_page()
+                    );
+                }
+            }
             let commands = spec.cmd.all_subcommands().into_iter().filter(|c| !c.hide);
             for cmd in commands {
                 let md = ctx.render_cmd(cmd)?;
@@ -86,6 +102,15 @@ impl Markdown {
             let md_idx = ctx.render_index()?;
             let path_idx = self.out_dir.as_ref().unwrap().join("index.md");
             write(&path_idx, &md_idx)?;
+            // Its own page, and only when there is something to put on it: a CLI with no
+            // settings should not gain an empty file in its docs directory.
+            let md_config = ctx.render_config()?;
+            if !md_config.trim().is_empty() {
+                // The same call the index links by, so the two cannot disagree about where the
+                // page went.
+                let path = self.out_dir.as_ref().unwrap().join(ctx.config_page());
+                write(&path, &md_config)?;
+            }
         } else {
             let md = ctx.render_spec()?;
             write_or_stdout(self.out_file.as_deref(), &render(&md))?;
--- a/cli/tests/markdown.rs
+++ b/cli/tests/markdown.rs
@@ -320,3 +320,55 @@ fn test_source_code_links_exist() {
 
     fs::remove_file(&out_file).unwrap();
 }
+
+#[test]
+fn a_page_collision_is_refused_before_anything_is_written() {
+    // A CLI with a literal `configuration` command: its page and the settings page want the same
+    // file. Writing anyway destroys the command's documentation, so the build stops — and stops
+    // *before* writing, so a refused run does not leave a half-built directory behind.
+    let dir = std::env::temp_dir().join(format!("usage_md_collision_{}", std::process::id()));
+    let _ = fs::remove_dir_all(&dir);
+
+    let specs = std::env::temp_dir().join(format!("usage_md_specs_{}", std::process::id()));
+    fs::create_dir_all(&specs).unwrap();
+    let clashing = specs.join("clash.usage.kdl");
+    fs::write(
+        &clashing,
+        "name \"ex\"\nbin \"ex\"\nconfig {\n  prop \"jobs\" type=\"uint\"\n}\ncmd \"configuration\" help=\"Odd\"\n",
+    )
+    .unwrap();
+    let ordinary = specs.join("ok.usage.kdl");
+    fs::write(
+        &ordinary,
+        "name \"ex\"\nbin \"ex\"\nconfig {\n  prop \"jobs\" type=\"uint\"\n}\ncmd \"settings\" help=\"Manage\"\n",
+    )
+    .unwrap();
+
+    usage_cmd()
+        .args(["generate", "markdown", "-f"])
+        .arg(&clashing)
+        .args(["--multi", "--out-dir", dir.to_str().unwrap()])
+        .assert()
+        .failure()
+        // A fragment miette does not wrap: it hard-wraps the message across lines.
+        .stderr(predicates::str::contains("rename the command"));
+    assert!(
+        !dir.exists() || fs::read_dir(&dir).unwrap().count() == 0,
+        "a refused run wrote files into {}",
+        dir.display()
+    );
+
+    // And the ordinary case — including a `settings` command, which no longer collides — still
+    // writes both pages.
+    usage_cmd()
+        .args(["generate", "markdown", "-f"])
+        .arg(&ordinary)
+        .args(["--multi", "--out-dir", dir.to_str().unwrap()])
+        .assert()
+        .success();
+    assert!(dir.join("configuration.md").exists());
+    assert!(dir.join("settings.md").exists(), "the command's own page");
+
+    let _ = fs::remove_dir_all(&dir);
+    let _ = fs::remove_dir_all(&specs);
+}
--- a/lib/src/docs/manpage/renderer.rs
+++ b/lib/src/docs/manpage/renderer.rs
@@ -78,6 +78,9 @@ impl ManpageRenderer {
             }
         }
 
+        // CONFIGURATION section
+        self.render_configuration(&mut roff);
+
         // AUTHOR section (if present)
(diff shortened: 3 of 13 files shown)


jdx:
One of the two is real and is fixed in the amended head; the other is a pre-existing property of the generator rather than of this change.

**Files-only configuration is suppressed** — correct, and the inconsistency was worse than a missing section: three output paths render the same docs model, and they disagreed about when there was anything to render. The standalone page asked `props.is_empty() && files.is_empty()`; the single-file template and the manpage asked about props alone. So a CLI that documents its config file chain before declaring its first setting — a reasonable thing, since the chain is the part a reader cannot guess — got it on one page and nowhere else. Both now use the model's own `is_empty`, with a test that renders a files-only spec through all three and two mutations confirming it fails without each fix.

This is the same defect class as the `is_empty` fix on #835 below it, which is where the block was being dropped by the *writer* for the same reason.

**Stale settings page survives regeneration** — accurate as a description, but `generate markdown --multi` has never pruned anything: there is no `remove_file` anywhere in it, so removing a *command* leaves its page behind in exactly the same way. Special-casing `settings.md` to delete itself would make it the one page in the directory that does, which is a worse inconsistency than the one it fixes. Pruning an out-dir is a reasonable feature and it should apply to every generated page at once — worth its own change, not this one. Leaving it as is deliberately.

jdx:
Fixed the second Bugbot finding too (`Multi-mode index omits settings link`) — my own incompleteness: `--multi` wrote `settings.md` and nothing pointed at it, so a reader who starts at the index, which is what an index is for, never reached the page.

`index.md` now links it, gated on exactly the condition that decides whether the page is written at all, so the index can never point at a file that was not created. Tested in both directions and mutation-verified in both — dropping the link fails, and linking unconditionally fails too.

Bugbot's `Files-only config skipped inconsistently` is the same finding as Greptile's, fixed in the commit above.

jdx:
One more finding, and it was a good catch: **config help never got the rendering options it was asked for.**

`MarkdownRenderer::new` renders the whole docs model eagerly — which happens *before* the builder methods that set `replace_pre_with_code_fences` — and rendering marks each item done, so a later pass no-ops. `render_cmd` sidesteps this by re-deriving from the raw command its caller hands it; config had no such argument, so it took the already-rendered copy and the option silently did nothing to a setting's long help while working everywhere else on the same page.

The renderer now keeps the raw config block, and all three paths — the settings page, the single-file page, and the index — derive from it.

Worth reporting how close this came to being verified badly. My first test asserted that `<pre>` became a fence, on the strength of the option's *name*; it failed, and the reason was that `replace_pre_with_code_fences` actually converts four-space-indented blocks and has nothing to do with `<pre>` tags. Had the premise been wrong in the other direction the test would have passed while measuring nothing. Then the fix for `render_spec` survived its mutation, because only `render_config` was covered — so the test now exercises both paths, and both mutations fail.

jdx:
Two more, both real.

**The manpage ignored `help_heading`.** The docs model already partitions settings by heading so the formats stay aligned; the manpage walked the flat list, so every heading was dropped and headed settings were interleaved with unheaded ones in one alphabetical run. It now renders the groups, with `.SS` per heading.

**The facts list could start against the line above it.** The blank line that opens the list was emitted inside the `type_` branch, so a prop with a default and no declared type put its first list item straight against the heading — and a deprecated one put it against the admonition's closing `:::`, where a renderer can read it as part of the admonition. The condition now mirrors exactly the branches that emit items.

Both mutation-verified, and both needed a test written for them: the existing snapshot fixture happened to contain no prop that exercised either case, so the fix passed its first mutation while measuring nothing.

jdx:
Three more, and the first two were found by CodeRabbit, which has started reviewing this stack.

**The settings heading was glued to the line above it, at the wrong level.** CodeRabbit flagged the `header_level` decrement; reproducing it showed a second problem hiding behind the first — `{%- include %}` also stripped the blank line, so the single-file page actually read:

```
Run# Configuration
```

`Run` being the last line of the preceding command's help. Both fixed: a plain `include` with the blank line kept, and no decrement, so the section is `## Configuration` under the document title with its settings at `###`.

**The manpage never rendered `choices`.** A constrained setting had no valid-value documentation in the terminal at all. Its values now join the facts line as `one of: git, none`; the per-choice help stays on the web page, where there is room for it.

**The manpage dropped `deprecated_remove_at`.** Markdown says "Removed in 2027.12.0" and the man page said only the reason — leaving a reader with nothing to plan around, in the one place deprecation is supposed to surface.

All three mutation-verified.

jdx:
Real, and it lands on the very CLI this feature is aimed at: **mise has a `settings` command**, so its command page and the settings page both wanted `settings.md`, config was written second, and the command's page was silently replaced — with the index then linking both entries to the same file.

The page is now written to `configuration.md` when a top-level `settings` command already occupies `settings.md`, and the index links whatever name was chosen through the same call, so the two cannot disagree about where the file went. Verified end to end: both pages exist, the command's page is intact, and the index points at each correctly. Mutation-verified in both directions.

jdx:
Accurate, and it made me reconsider the fix rather than patch it. Verified both halves first:

```console
$ usage g markdown -f with-settings-cmd.kdl --multi --out-dir d   # → settings.md + configuration.md
$ usage g markdown -f without.kdl          --multi --out-dir d    # → configuration.md is now stale
```

Choosing between two names by whether a command is in the way trades a silent overwrite for a silent stale page. So there is no choice any more: the settings page is **always** `configuration.md`. A fixed name cannot be abandoned between runs, and `configuration` is a name CLIs give to a *file* — the command is called `config`.

The one collision left — a CLI with a literal `configuration` command — is reported on stderr rather than left silent, since silence is exactly what made the `settings.md` case hard to see:

```
warning: the `configuration` command's page and the settings page both write configuration.md — the settings page wins
```

Re-verified end to end: two runs across the command being added and removed leave no page the generator abandoned. (A removed *command* still leaves its own page, as it always has — `gone.md` survives a run where `gone` no longer exists. That is worth fixing as one change covering every generated page, and I would support it, but it is not this one.)

jdx:
Both bots flagged the same thing and they are right — it is a weak assertion of mine, and exactly the kind I have been checking for elsewhere. The no-settings case asserted the absence of `settings.md`, a name nothing writes any more, so a broken gate emitting a link to `configuration.md` sailed through it.

It now asserts against `renderer.config_page()`, so the assertion cannot drift from the name again. Verified by mutation: forcing the gate open (`{%- if true %}`) fails the test, which it did not before.

jdx:
Right — warning and then writing over the command's page still loses the documentation, and a warning inside a docs build is a warning nobody reads. It is refused now, and refused *before* anything is written, so a stopped build does not leave a half-populated directory:

```console
$ usage g markdown -f clash.usage.kdl --multi --out-dir d
Error: the `configuration` command's page and the settings page would both be written to
       configuration.md; rename the command, or hide it, to generate both
$ ls d          # nothing
```

I went with refusing rather than picking a winner because the situation has no correct answer — two documents want one file, and both were asked for. No CLI in the fleet has a `configuration` command (they call it `config`), so this asks the one author who ever hits it to choose, rather than quietly losing a page for everyone who does not. A CLI-level test covers the refusal, the empty output directory, and that the ordinary case — including a `settings` command, which no longer collides — still writes both pages.

PR #841, opened 2026-08-12
feat(derive): hold the bytes a word arrived as

jdx:
A value used to reach its field through `from_utf8_lossy`, so a path with a stray byte in it
became a path with `U+FFFD` in it — a different file, silently. That is the worst shape a bug
can take: no error, wrong answer.

The partial holds the bytes now and the conversion happens once, where the struct is built.
`apply` cannot report anything — it answers whether an event belonged to this command — which
is why the bytes have to travel that far before anyone can complain about them. A word that
is not UTF-8 is an `InvalidValue` naming the field, showing the value lossily *for the message
only*, and saying what was wrong with it.

It also retires a hazard review raised on the last PR: `String` was recognised by how it was
written so that its value could be moved rather than converted, which broke for an adopter
who shadowed the name. There is no identity case left — everything converts — so recognising
the spelling only skips a second step now, and getting it wrong is a compile error rather
than a mangled value.

Costs +656 instructions (1.6%) and one allocation against main, measured on the same fixture:
the collecting path rebuilds a `Vec<String>` from bytes rather than moving one. That is what
not corrupting a value is worth.

Reporting such a value is not the same as accepting it, and accepting it needs
`OsStr::from_encoded_bytes_unchecked` — `unsafe`, in a crate that has none. The call would be
sound and PLAN.md says why, but that is jdx`s call rather than mine.

---

---

> [!NOTE]
> **Medium Risk**
> Changes how all CLI string values are stored and converted during parse; behavior improves for invalid UTF-8 but touches hot derive codegen paths with a small measured perf cost.
> 
> **Overview**
> **Argv values are kept as raw bytes through binding** instead of being turned into strings with `from_utf8_lossy`. The derive partial now stores `Vec<u8>` for string-like fields; conversion to `String` / `FromStr` types happens once in `build`, where invalid UTF-8 becomes **`InvalidValue`** (the error message may show the value lossily for display only).
> 
> **`value_enum` / `choices` checks** skip bytes that are not valid UTF-8 so users get a UTF-8 error instead of a misleading **`InvalidChoice`** list.
> 
> The old **`String` identity fast path** is removed—every typed field goes through the same UTF-8 validation path. **Defaults and `env` fallbacks** write UTF-8 bytes into the partial to match.
> 
> Conformance tests cover non-UTF-8 `--out` paths and enumerated `--shell` values. **PLAN.md** marks “report, don’t mangle” done and notes accepting exact non-UTF-8 paths is still future work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea7b427785427d58f7b40a7e3d840248f0aa24a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Cost

Measured on the mise-scale shadow (`use -g node@20`), differencing two runs of the same
binary against the same fixture on `main`:

| | main | this branch |
|---|---:|---:|
| instructions, cold parse | 39,828 | 40,484 |
| allocations, bound parse | 3 | 4 |
| allocations, bare parse | 0 | 0 |

+656 instructions, or 1.6%, and one allocation — the `Vec<u8>` a value is held in before it
becomes a `String`. That is the price of not silently substituting a different filename, and
it is paid only by invocations that carry a value.

## What this does not do

A non-UTF-8 value is *reported*, not accepted: `PathBuf` on Unix could hold those bytes
faithfully, but constructing an `OsStr` from them needs `OsStr::from_encoded_bytes_unchecked`,
and usage-argv has no `unsafe` in it today. The bytes do come from `as_encoded_bytes` in this
same process and every split the parser makes is at an ASCII byte, so the call would be sound
— but that is a policy decision about the crate rather than a detail of this change, so this
branch takes the conservative half and names the field in the error instead.

**Resolved in #844**, which is stacked on this one: jdx approved the `unsafe`, and it then
turned out to be needed only on Windows — on Unix the safe `OsString::from_vec` does the job.
A `PathBuf` field there accepts the bytes exactly, and does so for 567 instructions against a
`String` field's 660.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved command-line, default, and environment values without lossy UTF-8 replacement.
  * Added clear validation errors for invalid UTF-8 values, including typed options, paths, strings, and collections.
  * Improved distinction between invalid values and valid text that does not match available choices.

* **Documentation**
  * Updated documentation to explain byte-preserving value handling and current limitations.

* **Tests**
  * Added coverage for valid paths and text, invalid UTF-8 input, enum validation, and field-specific errors.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -126,11 +126,18 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       nine of these. The list is declared once, on the type: the spec, the help, the
       completions and the check that rejects a wrong word all read it from there, so none of
       them can drift from the type the way a second list on the field would.
-- [ ] **Values that are not valid UTF-8** — a word reaches a field through
-      `from_utf8_lossy`, so a `PathBuf` holding a non-UTF-8 path gets replacement
-      characters instead of bytes. The partial should hold `OsString` and let `build`
-      decide: exact for `PathBuf` and `OsString`, an error for `String` rather than a
-      silent mangling. Small, and the next thing.
+- [x] **Values that are not valid UTF-8 are reported, not mangled** — the partial holds the
+      bytes a word arrived as, and the conversion happens once where the struct is built, so
+      `--out /tmp/\xff` says so instead of handing a `PathBuf` a path with `U+FFFD` in it —
+      a different file, silently. Costs +656 instructions (1.6%) and one allocation, which
+      is what not corrupting a value is worth. It also retired the hazard of recognising
+      `String` by its spelling, since there is no identity case left.
+- [ ] **Accepting a value that is not valid UTF-8** — reporting it is not the same as taking
+      it. `PathBuf` could hold the exact bytes, but recovering an `OsString` from them needs
+      `OsStr::from_encoded_bytes_unchecked`, which is `unsafe`, and this crate has none. The
+      call would be sound — the bytes come from `as_encoded_bytes` in the same process, and
+      every split the parser makes is at an ASCII byte, so no multi-byte sequence is ever
+      cut — but introducing `unsafe` is jdx's call to make, not mine.
 - [ ] **`usage-derive` v1** — everything mise needs: constraints
       (`requires`/`conflicts`/`overrides`/`required_unless`), `var`, `count`,
       `env`, defaults, delimiters, the `double_dash` modes, global flags, flatten,
--- a/conformance/tests/typed.rs
+++ b/conformance/tests/typed.rs
@@ -299,6 +299,42 @@ fn the_words_reach_the_spec_from_the_type() {
 }
 
 #[test]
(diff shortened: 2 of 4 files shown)


jdx:
Addressed in the latest push.

**\`choices\` masked a UTF-8 failure** — all three of you found this independently, which it deserved. A non-UTF-8 value became \`""\` through \`unwrap_or_default()\`, matched no choice, and came back as \`InvalidChoice\`: a message listing words, about a value that is not a word. The comment beside it already claimed \`build\` reported this; it did not. Bytes that will not convert are now passed over and left for \`build\`, which names the field and gives the UTF-8 error as the reason. A value that *is* text and is not one of the choices still gets the list. Reproduced first, then fixed — \`ea7b427\`.

PR #842, opened 2026-08-12
feat(derive): declare the properties mise patches in by hand

jdx:
`src/cli/usage.rs` exists because clap cannot say three things: that a bare `mise
build` means `mise run build`, that `:::` restarts an invocation, and that `run`'s
subcommands come from `mise tasks --usage`. mise generates its spec and then edits
those in afterwards.

Declared instead, as `#[usage(default_subcommand = …)]` on the root and
`#[usage(restart_token = …, mount = …)]` on a command, so they travel with the code
that defines the command — which is the point of emitting a spec rather than patching
one.

All three are emission-only: a mount costs a subprocess and belongs to completions,
which is the cold path; a restart token is read by whoever splits a command line into
several invocations; and a default subcommand tells a completion engine what a bare
word means. Nothing here changes how a word binds, and the test asserts that as
directly as it asserts the KDL — including that a sibling declaring none of the three
binds exactly as it did.

`default_subcommand` is refused at compile time on a root with no subcommand field.
Which command it names cannot be checked from one expansion, but a root that can never
satisfy it can be caught where it is written.

The shadow generator carries them in the usage dialect and counts them as dropped in
the clap dialect, so the two shadows no longer lose the same set and the report says
which side lost what — a silent drop would make the clap side look like a faithful
translation of a spec it cannot represent.

---

---

> [!NOTE]
> **Medium Risk**
> Macro derive and spec emission changes affect all generated CLIs; argv binding is intentionally unchanged, but mis-declared attributes are now compile errors and default-subcommand routing divergence vs usage-lib remains until parser work lands.
> 
> **Overview**
> **`usage-derive` can now declare three command-level spec fields** that mise used to patch after emission: `default_subcommand` on the root `#[derive(Cli)]`, and `restart_token` / `mount` on `#[derive(Args)]` subcommands. Parsed attributes flow into generated `Spec` / `CommandMeta` so KDL emission carries them; **`check_position`** rejects misplaced declarations (e.g. `default_subcommand` without subcommands, or `mount`/`restart_token` on the root).
> 
> Conformance test **`root_grammar.rs`** checks the emitted KDL and that parsing/bindings are unchanged for mounts and restart tokens; it **documents** that **`default_subcommand` is not routed** during parse yet (unlike usage-lib).
> 
> **`xtask` shadow generation** emits these in the usage dialect and **counts them as dropped** in the clap dialect; shared context is refactored into a **`Run`** struct. The mise usage shadow and **PLAN.md** are updated to reflect expressible vs. still-missing behavior (routing on `default_subcommand`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcfafd9d180cc1d9c6801ba7b89ffaf1fc846b5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## What mise gets to delete

`src/cli/usage.rs` post-processes the emitted spec to add exactly these three things. With
them declarable, what remains in that file is clearing `run`'s arguments, which is a
consequence of the mount rather than a separate workaround. It already records one hack that
went away when jdx/usage#738 landed.

## Verification

- The three properties reach the emitted KDL, and a sibling command that declares none of
  them gets none of them.
- Binding is unchanged: `run --dry-run` still binds `run`'s own flag with a mount declared,
  the undeclared sibling binds identically, and a bare word still fills the root's own
  argument rather than being routed through the default subcommand.
- `default_subcommand` on a root with no subcommand field is refused at compile time.
  Mutation-checked: with the condition forced false, the test fails.
- The shadow generator carries them in the usage dialect and counts them as dropped in the
  clap dialect, each asserted separately. The "only the root declares one" test is
  mutation-checked against a child inheriting the root's value — the mistake the shared
  traversal context makes easy.
- The generator's argument list is now a `Run` struct, which let its pre-existing
  `#[allow(clippy::too_many_arguments)]` be deleted rather than extended.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -178,6 +178,14 @@ carrying them rather than being worked around.
       the only reading under which `[a]… [b]` can be filled. `var_max` therefore moved into
       the table binding reads; `var_min` stays a check, because no single word tells you a
       variadic will end up short. Costs 55 instructions per parse, or 0.1%.
+- [x] **The command-level properties mise patches in by hand** — `default_subcommand`,
+      `restart_token` and `mount`, declared with `#[usage(...)]` instead of edited into the
+      spec afterwards by `src/cli/usage.rs`. None of the three changes how a word binds: a
+      mount costs a subprocess and belongs to completions, a restart token is read by
+      whoever splits an invocation into several, and a default subcommand tells a completion
+      engine what a bare word means. So they are emission-only, and the test asserts both
+      that they reach the KDL and that binding is unaffected. `default_subcommand` is
+      checked at compile time to require a subcommand field, since it names one.
 - [ ] **A mount on the root command** — the spec accepts `mount` only inside a
       `cmd` block, so a CLI whose _top-level_ subcommands are discovered by running
       something cannot say so. Worth deciding whether that is a gap or a deliberate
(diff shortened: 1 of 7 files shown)


jdx:
Addressed in the latest push.

**All four findings were right.**

- \`mount\`/\`restart_token\` on the root: confirmed against usage-lib, which rejects a spec with either at the top, so these were headed for a \`debug_assert!\` in the writer or a silent drop in release. Refused at the declaration now.
- \`default_subcommand\` on a nested \`Args\` (Bugbot and Greptile both): refused rather than dropped. Both derives share one parse, so the rule moved to a \`check_position\` each entry point calls for the place it stands in — before, a nested struct got whichever check ran first, which named the wrong problem.
- The corrupted comments: entirely self-inflicted, a regex over the function body rewrote prose as well as code. Fixed all three.

PR #844, opened 2026-08-12
feat(derive): accept a value the OS accepts and UTF-8 does not

jdx:
Reporting a non-UTF-8 value was the safe half of the fix. It is not the whole one:
`/tmp/\xff` is a filename the operating system accepts, and a CLI that cannot receive
one cannot open the file. A `PathBuf` or `OsString` field now takes the bytes exactly.

`usage_argv::os_string_from_bytes` is the reverse of `as_encoded_bytes`. On Unix it is
the *safe* `OsString::from_vec` — an `OsString` there is an arbitrary byte sequence, so
the conversion is total and no `unsafe` arises. Only Windows needs
`from_encoded_bytes_unchecked`, because WTF-8 makes the conversion partial.

That call rests on an invariant, now written where it can be checked: every sub-slice
the parser produces is cut at an ASCII byte — after `-`, after `--`, at `=`, and between
the letters of a short-flag cluster — and an ASCII byte never occurs inside a multi-byte
sequence. The one way to break it is a non-ASCII `Flag::shorts` entry, which would let a
cluster split mid-character. The derive already refused that; the rule is now documented
as load-bearing on both sides and has the test it was missing.

The crate goes from `forbid(unsafe_code)` to `deny`, so the single audited exception has
to name itself rather than the crate pretending it has none. Parsing still contains no
`unsafe` at all.

Byte-exactness is cheaper than the text path, not dearer: a `PathBuf` field costs 567
instructions per parse against a `String` field's 660, because it skips the UTF-8
validation pass, and both allocate once. The gate fixture cannot show this — a spec
carries no Rust types, so every shadow field is a `String` — so it is measured directly.

A `String` field still reports rather than substitutes, which is the right answer for a
type that cannot hold those bytes.

CI gains a Windows compile check, since every test here is `#[cfg(unix)]` and nothing in
the pipeline was building the branch that carries the `unsafe`.

---

---

> [!NOTE]
> **Medium Risk**
> Touches argv-to-field conversion for path types and platform-specific behavior on Windows; Unix behavior is well-tested but the Windows path is compile-checked only, not exercised in tests.
> 
> **Overview**
> **`PathBuf` and `OsString` fields now keep argv bytes instead of requiring UTF-8**, so paths like `/tmp/\xff` parse as the OS sees them rather than being rejected or mangled. The derive routes those types through new **`usage_argv::os_string_from_bytes`**: lossless **`OsString::from_vec` on Unix**, and on **Windows** a **safe UTF-8 round-trip** (invalid bytes become **`InvalidValue`**) rather than **`from_encoded_bytes_unchecked`**, so the crate stays **`forbid(unsafe_code)`**.
> 
> Docs tie **ASCII-only short flags** to parser slicing (clusters and `=` cuts), with derive validation and a test reinforcing that. **Unix conformance tests** cover detached, attached, short, repeated, and **`OsString`** forms; **`String`** still errors on non-UTF-8. **CI** adds a **`x86_64-pc-windows-msvc` `cargo check`** for the non-Unix branch. **PLAN.md** marks the milestone done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 426f1bc06c4c3a8b4034128c749711b6f464432a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Cost

A `PathBuf` field is now **cheaper** than the same field as a `String`, because it skips the
UTF-8 validation pass:

| field type | instructions per parse | allocations per parse |
|---|---:|---:|
| `Option<PathBuf>` | 567 | 1 |
| `Option<String>` | 660 | 1 |

The gate fixture cannot show this and was not used for it: a spec carries no Rust types, so
every field in the generated shadow is a `String` and the shadow's numbers are unchanged
(40,472 against 40,484 on the parent — the same code path, and the difference is noise).
Measured instead on a two-field probe, differencing `PARSE_N=0` against `PARSE_N=100` runs of
the same binary and dividing.

## Why this is sound

On **Unix** there is no `unsafe`: an `OsString` is an arbitrary byte sequence, so the
conversion is the safe `OsString::from_vec` and no invariant is required. This is worth
stating plainly because it means the approval was only needed for one platform.

On **Windows** the encoding is WTF-8, not every byte sequence is valid, and only
`from_encoded_bytes_unchecked` will take one. That rests on an invariant of the crate:

> Every sub-slice the parser produces is cut at an ASCII byte.

A token is split after `-`, after `--`, at `=`, and between the letters of a short-flag
cluster. All four are ASCII, and an ASCII byte never occurs inside a multi-byte WTF-8 or UTF-8
sequence, so no cut can land mid-character. Values passed through whole are trivially fine.

The one way to break it is a non-ASCII `Flag::shorts` entry, which would let a cluster split
mid-character. The derive already rejected that — for ergonomic reasons — so the rule was
already true; what this PR adds is that it is now *documented as load-bearing* on both the
field and the check, and has the test it was missing.

## Verification

- Every route a value can take keeps its bytes: detached (`--out <bad>`), attached
  (`--out=<bad>`, cut at the `=`), short (`-o<bad>`), short-with-equals (`-o=<bad>`),
  repeated into a `Vec<PathBuf>`, and a bare `OsString` field. The cluster cases are the ones
  the soundness argument leans on hardest, which is why they are enumerated rather than
  sampled.
- Mutation-checked: with the conversion switched to `from_utf8_lossy`, both byte-exactness
  tests fail.
- `Option<Vec<PathBuf>>` still tells "never given" from "given nothing" — that decision is
  made before the conversion and had to survive the new path.
- A `String` field still reports rather than substitutes, since it cannot hold those bytes.
- The non-ASCII `short` rejection now has a test, and the `unsafe` branch is compile-checked
  for `x86_64-pc-windows-msvc`.

## One gap worth naming

Every test here is `#[cfg(unix)]`, and CI had no Windows job — so the branch carrying the
`unsafe` was the one thing in the pipeline nothing built. This adds a Windows compile check
to `test.yml`. A compile check is not coverage: the Windows conversion is still unexercised
by any test, and reaching it needs an unpaired surrogate in `argv`. If that matters enough to
you, the next step would be a Windows test job rather than a check.

## Summary by CodeRabbit

- **New Features**
  - Added support for preserving non-UTF-8 command-line values in path and operating-system string inputs on compatible platforms.
  - Added short-form output support, repeated path inputs, raw values, and optional repeated exclusions.
  - Added clear errors when values cannot be converted on supported platforms.

- **Bug Fixes**
  - Fixed path argument handling so raw bytes are preserved instead of being altered.
  - Improved validation and error reporting for non-ASCII short flags.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,9 +40,9 @@ jobs:
           fi
       - run: mise r build
       - run: mise r test
-      # The one code path no test can reach from here: `os_string_from_bytes` is safe on
-      # Unix and `unsafe` on Windows, and every test of it is `#[cfg(unix)]`. A compile
-      # check is not coverage, but it keeps the branch carrying the `unsafe` from rotting
+      # `os_string_from_bytes` has a `cfg` branch per platform, and every test of it is
+      # `#[cfg(unix)]` — so the Windows one is not compiled anywhere else in this pipeline.
+      # A compile check is not coverage, but it is what keeps that branch from rotting
       # unseen.
       - name: check the Windows-only conversion still compiles
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,14 @@ jobs:
           fi
       - run: mise r build
       - run: mise r test
+      # The one code path no test can reach from here: `os_string_from_bytes` is safe on
+      # Unix and `unsafe` on Windows, and every test of it is `#[cfg(unix)]`. A compile
+      # check is not coverage, but it keeps the branch carrying the `unsafe` from rotting
+      # unseen.
+      - name: check the Windows-only conversion still compiles
+        run: |
+          rustup target add x86_64-pc-windows-msvc
+          cargo check -p usage-argv --all-features --target x86_64-pc-windows-msvc
       - run: mise r render
       # Same reasoning as `render`: the shadow is checked in, so a change to the derive's
       # vocabulary that would alter it has to be committed rather than discovered later.
--- a/PLAN.md
+++ b/PLAN.md
@@ -132,12 +132,26 @@ manpages, and SDKs — never a runtime dependency of somebody else's program.
       a different file, silently. Costs +656 instructions (1.6%) and one allocation, which
       is what not corrupting a value is worth. It also retired the hazard of recognising
       `String` by its spelling, since there is no identity case left.
-- [ ] **Accepting a value that is not valid UTF-8** — reporting it is not the same as taking
-      it. `PathBuf` could hold the exact bytes, but recovering an `OsString` from them needs
-      `OsStr::from_encoded_bytes_unchecked`, which is `unsafe`, and this crate has none. The
-      call would be sound — the bytes come from `as_encoded_bytes` in the same process, and
-      every split the parser makes is at an ASCII byte, so no multi-byte sequence is ever
-      cut — but introducing `unsafe` is jdx's call to make, not mine.
+- [x] **Accepting a value that is not valid UTF-8** — reporting it was the safe half;
+      accepting it is the whole fix, because the operating system does accept `/tmp/\xff`
+      as a filename and a CLI that cannot receive one cannot open the file. A `PathBuf` or
+      `OsString` field now takes the bytes exactly, through
+      `usage_argv::os_string_from_bytes`. On Unix that is the **safe** `OsString::from_vec`,
+      so the question of `unsafe` never arises there; only Windows needs
+      `from_encoded_bytes_unchecked`, because WTF-8 makes the conversion partial. jdx
+      approved the `unsafe` for that case.
+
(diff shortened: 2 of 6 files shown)


jdx:
Addressed in the latest push.

**The P1 was correct and the fix was to remove the \`unsafe\`, not to justify it better.** A safe function taking a \`Vec<u8>\` cannot enforce a "these came from \`as_encoded_bytes\`" precondition, so no amount of documentation made it sound.

On Unix an \`OsString\` is an arbitrary byte sequence, so \`OsString::from_vec\` is exact and safe — that is the case that matters, since non-UTF-8 filenames are ordinary there. On Windows the bytes go through UTF-8 and one that will not convert comes back in the \`Err\` to be reported. The crate \`forbid(unsafe_code)\`s again. Byte-exactness is still cheaper than the text path: 553 instructions per parse against 674.

PR #848, opened 2026-08-12
feat(argv): route a word that names nothing to the default subcommand

jdx:
`default_subcommand` was emitted into the spec and ignored by the parser, which the
previous commit recorded as a gap rather than a decision. usage-lib routes on it, and
this is the rule behind `mise build` meaning `mise run build`.

A word naming no subcommand now descends into the named one, and the cursor steps back so
the word is examined again against the command it reached — which lets it be that
command's argument or one of *its* subcommands, without this deciding which and without
two events for one word. The declaring command's own positional does not win, which is
what makes the property more than a synonym for an argument, and is mise's shape exactly:
the root has `[TASK]` and so does `run`.

Taken at most once per parse, as usage-lib does, so a chain of defaults cannot walk a CLI
deeper than anything the user typed.

The name is resolved by `find_subcommand` during const evaluation, so a name no
subcommand answers to is a compile error. The previous commit said the name could not be
checked because the variants are another expansion; that is true of the macro and not of
const eval, which can search the list the parent already holds.

Seven corpus vectors, six agreeing with usage-lib and one recording where it does not:
usage-lib holds the single name spec-wide and applies it at whichever command it stands
on, so `ex config zzz` descends into `config ls` when an unrelated command happens to
have a subcommand of that name. Read here as a property of the root, which is the only
place a spec can declare it.

`default_subcommand` on a nested `Args` is now refused outright instead of being dropped
in silence. Both derives share one parse, so the rule moved to where the context is
known — before, a nested struct got whichever check ran first, which named the wrong
problem.

No measurable cost: 40,370 instructions against 40,472, the branch being reached only by
a word that matched no subcommand. Allocations unchanged — 0 bare, 4 bound.

One thing this does not buy: `mise build` still fails in the shadow, because mise's spec
gives `run` no positional — its arguments are cleared and a mount supplies task names,
which usage-argv does not execute. Routing plus mounts is what lets mise delete its
hand-rolled dispatch; the gate test now says so rather than asserting the old answer.

---

---

> [!NOTE]
> **Medium Risk**
> Changes core subcommand routing and argv binding semantics (default descent, once-per-parse latch, token rewind), which can alter how bare words and edge cases parse across CLIs that declare `default_subcommand`.
> 
> **Overview**
> **Implements `default_subcommand` in the hot-path parser** so a bare word like `build` can mean “descend into `run` and bind there,” matching usage-lib and mise’s `mise build` → `mise run build` shape.
> 
> `Command` gains `default_subcommand` plus **`find_subcommand`** (const-time name/alias lookup, compile error if the name is missing). In `word()`, after a failed subcommand match, the parser may descend once per parse (`default_taken`), **rewind the argv cursor**, and re-read the same token under the default command. Routing is skipped for flag-like tokens, after `--`, and when a positional already filled; unknown flags still bind at the current command.
> 
> **Derive and conformance** wire the root-level spec property into static tables; nested `default_subcommand` on `Args` is rejected. **Corpus** adds `09-default-subcommand.json` (including one recorded usage-lib divergence: root-only vs spec-wide default). **Gate shadow** now expects `mise build` to route into `run` then fail with `UnexpectedArg` until mounts exist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46fff423795cb879f7789aaa1c29d638bb8286c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## The decision behind this

Investigating `flatten` turned up that five of mise's ten `#[clap(flatten)]` sites are a
subcommand field beside a flatten of *the same type the `Ls` variant holds* — `config`,
`tasks`, `settings`, `deps`, `tool` all hand-rolling "with no subcommand, behave as `ls`".
Which raised the question of what `default_subcommand` actually means, and the answer was that
#842 got it wrong: usage-lib routes on it while parsing.

```
["mise", "build"] -> cmds ["mise", "run"], args [("TASK", String("build"))]
```

Even though the root declares its own `arg "[TASK]"`. So the property is a binding rule, and
#842's test asserting otherwise was corrected in its own commit before this one.

## Cost

| | before | after |
|---|---:|---:|
| instructions, cold parse | 40,472 | 40,370 |
| allocations, bare / bound | 0 / 4 | 0 / 4 |

No measurable cost. The branch is reached only by a word that already failed a subcommand
lookup, and the ±100 is code layout rather than work done — cachegrind is deterministic and
gave the same figure on repeated runs.

## The name is now checked at compile time

#842 said "which command it names cannot be checked here — the enum's variants are in another
expansion". True of the macro, not of const evaluation: `find_subcommand` searches the list the
parent already holds, so

```
error[E0080]: evaluation panicked: `default_subcommand` names a command that this one does not have
```

A wrong name is a compile error rather than a spec property nothing validates.

## What this does not buy

`mise build` still does not work end to end in the shadow, and it is worth being plain about
why: mise's spec gives `run` **no positional at all** — `src/cli/usage.rs` clears them and adds
`mount run="mise tasks --usage"`, so task names are meant to come from running that.
usage-argv does not execute mounts, so after descending there is nothing to bind to. Routing
*plus* mounts is what would let mise delete its hand-rolled dispatch; this is half of it. The
gate test now records that instead of asserting the old answer.

## Verification

- Seven corpus vectors. Six agree with usage-lib — checked automatically, since an absent
  `reference` label is an assertion the reference test enforces in both directions.
- Six parser unit tests: routing, a named subcommand not routed, re-examination against the
  reached command, the once-per-parse latch, a flag before the word, and nothing routing past
  `--`.
- Mutation-checked: removing the latch fails the latch test; removing the cursor rewind fails
  four.
- End-to-end through the derive, plus a test that the resolved pointer is the table's own entry
  rather than a copy.

## One divergence, and a question

`default-is-declared-for-the-root` records it: usage-lib holds the single name for the whole
spec and applies it at **whichever command it is standing on**, so with `default_subcommand
"ls"` declared at the top, `ex config zzz` descends into `config ls` — an unrelated command
acquiring a default because a name happened to match one level down.

```
["config", "zzz"] -> cmds ["ex", "config", "ls"], args [("WHAT", String("zzz"))]
```

This parser reads it as a property of the root, which is the only place a spec can declare one.
That looks like the intended meaning and usage-lib's looks accidental — but it is your call,
and if you agree it is a small fix in `lib/src/parse.rs`. The corpus label makes the difference
fail loudly the moment it changes either way.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -49,7 +49,7 @@ usage-lib ── interprets a spec       usage-derive ── reads a Rust type
       │      at runtime                     │
       │                                     ├─→ static tables ─→ usage-argv (hot)
       └─→ the oracle the corpus             └─→ static metadata ─→ help,
-          measures everyone against              completions, spec output (cold)
+      measures everyone against              completions, spec output (cold)
 ```
 
 Four rules hold this together.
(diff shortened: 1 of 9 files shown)


jdx:
Addressed in the latest push.

**Both findings were real, and both confirmed against usage-lib before fixing.**

- Unknown flags routing (Bugbot): \`ex --wat\` in usage-lib comes back as commands \`["ex"]\` with the token bound to the root positional. Only a token that could have named a command routes now, which also excludes the \`--\` a \`preserve\` argument asks for as a value.
- Aliases (Greptile): usage-lib resolves the name against names, aliases *and* hidden aliases, so \`find_subcommand\` does too — before, pointing \`default_subcommand\` at an alias failed to compile.

Three corpus vectors, all agreeing with usage-lib, and both fixes mutation-checked.

PR #853, opened 2026-08-13
feat(derive): say three things about a CLI the spec could and the derive could not

jdx:
Found by rendering mise's help from the shadow's metadata and diffing every line against
usage-lib's. 23 of 211 differed, and every difference traced back to something mise's KDL
declared, the derive had no vocabulary for, and `gen-shadow` dropped **without counting**.
That last part is what made them invisible: the dropped-property report exists to prevent
exactly this, and it was reporting nothing.

`value_name` — the placeholder for a flag's value. `FlagMeta` had the field and the KDL
writer already read it; nothing could set it. So a spec saying `--tool <TOOL>` came back
saying `--tool <tool>`, because the flag's own name was all that was left. 211 of mise's
flags carry one.

`required` on a collecting argument — `<TARGET>…`, meaning one or more. Required-ness is
normally the type's to say, and a `Vec` cannot: it has no bare-versus-`Option` shape. This is
the one place it is declared rather than inferred, and it is refused anywhere the type
already answers the question, in either direction.

`var` on a counted flag — inferred rather than declared, because a count *is* repetition:
`-vvv` is three occurrences and counting them is the point. Left out, the emitted spec said
`count` without `var` where mise's says both.

These matter past help text. The emitted spec is what docs, manpages, completions and the SDK
generators read, so a property that cannot survive the round trip is one every downstream
consumer is wrong about.

`gen-shadow` emits all three now, and the report for the usage dialect drops from six kinds
of loss to three — the remaining ones being genuine gaps with nothing yet to express them.

---

---

> [!NOTE]
> **Medium Risk**
> Touches derive parsing, post-bind validation, and shadow codegen across the full mise CLI surface; behavior changes for empty required collections and help/spec placeholders, but bounded by new tests and spec diffing.
> 
> **Overview**
> Closes three spec→derive gaps that were silently dropped by `gen-shadow` (and invisible in the drop report): flag value placeholders, one-or-more collections, and repeatability on counted flags.
> 
> **`value_name`** on value-taking flags so help/spec show `--tool <TOOL>` instead of defaulting to the flag name. **`required`** on `Vec` positionals and collecting flags so `<TARGET>…` is enforced at parse time and in emitted KDL, aligned with metadata via a single required expression. **`var`** is inferred for `count` flags (and plain repeatable valueless flags get `var` in shadow gen) so `-vvv` matches the reference spec.
> 
> `gen-shadow` now emits these attributes; the mise shadow gains widespread `value_name` / `required` annotations. Compile-time validation rejects contradictory uses (`required` on `Option`/bare types, `value_name` on positionals or valueless flags, `required` vs `required_if`/`Option<Vec>`). Conformance tests cover metadata and required-collection parsing; `PLAN.md` marks the milestone done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93b378f7fae83000ed47486b03e2f6bc6cb7c6ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## How these were found

By starting the next item — help rendering — and immediately checking it against the
reference: render mise's usage line from the shadow's metadata, compare all 211 against
usage-lib's. **23 differed**, and every one traced back to something mise's KDL declared, the
derive had no vocabulary for, and `gen-shadow` dropped *without counting it*.

That last part is the real finding. The dropped-property report exists so the shadow can never
look like a faithful translation of a spec it cannot represent — and for these three it was
reporting nothing at all.

```
mise config get   [-f --file <file>] [KEY]   ← ours
                  [-f --file <FILE>] [KEY]   ← usage-lib

mise deps remove  [PACKAGES]…               ← ours
                  <PACKAGES>…               ← usage-lib

mise cache prune  [-v --verbose]            ← ours
                  [-v --verbose…]           ← usage-lib
```

## The three

| | what it is | why the derive could not say it |
|---|---|---|
| `value_name` | the placeholder for a flag's value | `FlagMeta` had the field and the KDL writer already read it — nothing could *set* it, so the flag's own name stood in |
| `required` on a collection | `<TARGET>…`, one or more | required-ness is the type's to say, and a `Vec` has no bare-versus-`Option` shape to say it with |
| `var` on a counted flag | that it may be given again | now inferred, not declared: a count **is** repetition — `-vvv` is three occurrences |

`required` is refused anywhere the type already answers the question, in either direction — on
an `Option` it contradicts, on a bare type it is redundant, and a declaration that changes
nothing is one someone will eventually trust.

## Why it matters past help text

The emitted spec is what docs, manpages, completions and the SDK generators read. A property
that cannot survive the round trip is one every downstream consumer is wrong about — which is
also why the fix belongs here rather than in the renderer.

The usage dialect's dropped-property report goes from six kinds of loss to three, and the three
that remain are genuine gaps with nothing yet to express them.

## Verification

- Four tests on a CLI declaring all three, asserting each reaches the parsed-back spec.
- All three mutation-checked: reverting each fix fails exactly its own test.
- Two refusal tests for `required` in the places the type already answers.
- `mise run gen-shadow` regenerated; 211 value names and 16 required markers now survive.

## Summary by CodeRabbit

* **New Features**
  * Generated CLI help and completion metadata now preserve explicit value names.
  * Required variadic arguments and collections are represented accurately.
  * Counted and repeatable flags now expose correct repeatability metadata.
  * Generated specifications better reflect command usage with required markers and value placeholders.

* **Bug Fixes**
  * Fixed discrepancies between generated specifications and rendered help output.
  * Improved metadata handling for flags, collections, flattened fields, and subcommands.
  * Added validation for unsupported or conflicting argument metadata.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -252,6 +252,17 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
       something cannot say so. Worth deciding whether that is a gap or a deliberate
       restriction.
 
+- [x] **Three things a spec could say that the derive could not** — a flag's value name
+      (`--tool <TOOL>` came back as `--tool <tool>`, since the flag's own name was all there
+      was), a collecting argument that needs at least one value (`<TARGET>…`, which a `Vec`
+      cannot express because it has no bare-versus-`Option` shape), and `var` on a counted flag
+      (a count _is_ repetition, so it is inferred now rather than needing told). All three were
+      dropped by `gen-shadow` without being counted, which is the part that made them invisible:
+      the report is the thing that was supposed to prevent exactly this. Found by rendering
+      mise's help from the shadow and diffing every line against usage-lib's — 23 of 211 differed,
+      and every difference traced to one of these. They matter past help text, since the emitted
+      spec is what docs, manpages, completions and the SDK generators read.
+
 ### Then: what a CLI framework has to have
 
 - [ ] **Help rendering** — `--help` and `-h` from the static metadata, with
--- a/benches/shadows/mise/src/lib.rs
+++ b/benches/shadows/mise/src/lib.rs
@@ -33,6 +33,7 @@ pub struct ActivateArgs {
         long = "shell",
         short = 's',
         hide,
+        value_name = "SHELL",
         choices("bash", "elvish", "fish", "nu", "xonsh", "zsh", "pwsh")
     )]
     pub shell: ::std::option::Option<::std::string::String>,
@@ -126,7 +127,7 @@ pub struct ToolAliasUnsetArgs {
 #[derive(Args)]
 pub struct ToolAliasArgs {
     /// Filter aliases by tool
-    #[usage(long = "tool", short = 'p')]
+    #[usage(long = "tool", short = 'p', value_name = "TOOL")]
     pub tool: ::std::option::Option<::std::string::String>,
     /// Don't show table header
     #[usage(long = "no-header")]
@@ -313,7 +314,7 @@ pub struct BootstrapDotfilesAddArgs {
     #[usage(long = "local", short = 'l')]
     pub local: bool,
     /// Dotfile mode to write
-    #[usage(long = "mode", short = 'm')]
+    #[usage(long = "mode", short = 'm', value_name = "MODE")]
     pub mode: ::std::option::Option<::std::string::String>,
     /// Print the config/source updates without writing anything
     #[usage(long = "dry-run", short = 'n')]
@@ -322,16 +323,16 @@ pub struct BootstrapDotfilesAddArgs {
     #[usage(long = "no-apply")]
     pub no_apply: bool,
     /// Write to this config file or directory
-    #[usage(long = "path", short = 'p')]
+    #[usage(long = "path", short = 'p', value_name = "PATH")]
     pub path: ::std::option::Option<::std::string::String>,
     /// Source path to use for a single target
-    #[usage(long = "source", short = 's')]
+    #[usage(long = "source", short = 's', value_name = "PATH")]
     pub source: ::std::option::Option<::std::string::String>,
     /// Skip the confirmation prompt
     #[usage(long = "yes", short = 'y')]
     pub yes: bool,
     /// Targets to add or update
-    #[usage(arg, name = "TARGET")]
+    #[usage(arg, name = "TARGET", required)]
     pub target: ::std::vec::Vec<::std::string::String>,
 }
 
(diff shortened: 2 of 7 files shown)


jdx:
Three findings, all real and all the same shape — a declaration written down, emitted into the spec, and then not honoured.

**`required` on a collection was documentation only** (Bugbot, Greptile). The required check looked at the field's *shape*, and a `Vec` has the same shape whether or not it declared `required` — so the spec, the help, the manpage and the completions all said one-or-more while the CLI that ran accepted none. It now uses the same expression the metadata is built from:

```rust
if !(f.shape == Shape::Required || f.required_collection) || f.default.is_some() {
```

That is deliberately the identical expression to the one at the two `*_meta` sites: one expression cannot disagree with itself.

**`value_name` was accepted where nothing could show it** (Greptile, CodeRabbit — same finding). It names the placeholder a flag's *value* gets in help; `arg_meta` never emits it and a `bool`/`count` flag has no placeholder, so on a positional or a valueless flag the declaration compiled and was dropped. Refused now, with a message that says what to use instead — `name` for a positional.

**The shadow generator dropped `required` from a collecting flag** (CodeRabbit). A scalar flag says required by *type* — `String` rather than `Option<String>` — and a `Vec` has no such spelling, so `required` is the only way to say it. Without emitting it, a spec demanding one-or-more values became a shadow accepting none, whose regenerated spec read `[VALUE]…` where the source said `…`. Note this one depends on the first fix: emitting `required` there is only honest now that the derive enforces it, which is why they belong in the same commit.

Each has a test in the layer that owns the behaviour — parse-time in the conformance suite, attribute validation in the model's own tests, generator output in the shadow's — and each test is checked against a reverted fix.

jdx:
Both real, and both the same thing as the first round: a declaration that compiles and cannot mean what it says. Refused now, beside the `value_name` check.

**`required` beside `required_if`/`required_unless`.** `required` on a collection means one or more values *always*, and the check for it runs unconditionally — so a sibling condition was accepted, emitted into the spec, and never consulted. For a scalar the two cannot meet: `required_unless` needs somewhere to put "absent" and so only goes on an `Option`. A collection has no type to keep them apart, so validation does.

**`required` on an `Option<Vec<_>>`.** Shaped like any other collection, so it was accepted — after which the field can never be `None` and the `Option` means nothing at all.

```
`required` on a collection means one or more values, always, but
`required_if` says it is required only sometimes
```

Three cases tested, three mutations, and a fourth assertion that a plain `#[usage(long, required)] Vec<String>` still compiles — which is the thing the refusals are protecting.

jdx:
Fixed. A line continuation had lost its backslash, so the one message that tells an author what a field accepts had a long run of spaces through the middle of it. The test asserts the absence of a double space rather than the exact list, so it stays true as options are added.

jdx:
All addressed, though two of them twice over — another session was fixing the same findings in parallel, and reconciling the two is most of what this took.

**Required collection not enforced (High)** — real, and fixed: `required_checks` now keys off the same expression the metadata is built from (`shape == Required || required_collection`), so the spec and the parser cannot disagree about it. One expression cannot contradict itself.

**Required ignores conditional exceptions / Required allowed on `Option<Vec>` (Medium ×2)** — both real. `required` on a collection means one or more values *always*, so it contradicts `required_if`, `required_unless`, and `Option<Vec<_>>` — each of which says absence is representable. All three refused at the declaration now.

**`value_name` silently discarded** and **`required` not preserved for collecting flags in the generator** — both fixed, with a generator regression test for the second.

**Malformed unknown-option error text (Low)** — the run of spaces was a stray line continuation, already gone.

One thing worth recording: the parallel fixes collided. Both sessions added a guard refusing the self-contradicting `required`; mine ran first and its wording did not match the test the other had written, and the reconciliation then dropped the *other* rule — refusing `required` where the type already answers — while keeping its variables. clippy caught the orphans (`assigned but never used`), and the test caught the missing rule. Both are back, and the duplicate guard is gone.

PR #854, opened 2026-08-13
feat(argv): render the usage line, byte-identical to usage-lib's

jdx:
`Usage: mise use [OPTIONS] <TOOL@VERSION>…` — the line at the top of `--help`, and the first
thing a CLI framework has to be able to produce. Built from the same `&'static` metadata a
parse ignores, so a binary that never asks for help pays nothing for being able to.

usage-lib renders this through a tera template over a runtime model. This crate has no `Spec`
at run time, only tables, so the rules are reimplemented — and reimplemented rules drift. The
test runs both over mise's real spec and compares all 211 lines, because an adopter's help
text changing is a visible regression even when the change is one bracket. The shadow is
generated from the same KDL usage-lib is handed, so the two describe the same CLI by
construction rather than by a fixture kept in step by hand.

Two things it caught:

An argument that only takes what follows a `--` is bracketed *around* the separator —
`[-- COMMAND]…`, one optional thing — where I had written `-- [COMMAND]…`, a literal `--`
followed by an optional word. Five of mise's commands read that way.

A flag is named after the form it answers to, not after the Rust field holding it. usage-lib
derives the name from the first long form, and the derive was kebab-casing the field ident —
so `type_` gave a flag called `type-`, which help printed as `type-: -t --type` and errors
reported as `type-`. An explicit `name` still wins, and a flag with no long form keeps its
short, as usage-lib does. Selectors resolve by form rather than by name, so nothing that
refers to a flag changes.

---

---

> [!NOTE]
> **Low Risk**
> Cold-path help formatting behind `spec`, with broad parity tests; derive naming changes align emitted KDL/help with usage-lib and do not alter parse-time selector matching.
> 
> **Overview**
> Adds **`usage_argv::help::usage_line`** behind the existing **`spec`** feature so the top-of-help invocation summary is built from static **`CommandMeta`** tables without a runtime spec model.
> 
> **Parity:** The gate bench compares every mise command’s line (211) against **usage-lib** from the same KDL; **PLAN.md** marks the usage line as done.
> 
> **Rendering fixes:** Defaulted flags/args render as optional (`required && default.is_empty()`), not raw `required`. **`--`** positionals use **`[-- NAME]…`** with brackets outside the separator. Flags/args collapse to **`[FLAGS]`** / **`[ARGS]…`** past an inline limit of 2.
> 
> **Derive:** Flags without an explicit **`name`** are named from the first long or short form (fixes **`type_` → `type-`**); short-only value flags keep the field ident as **`value_name`**.
> 
> **Tests:** Gate **`help.rs`**, conformance **`derive`** / **`metadata`** for defaulted and renamed flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32ff41b59316517f41e0d9ebaff5652f9a82fd1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## The standard

**All 211 of mise's commands render byte-identically to usage-lib.**

usage-lib builds this through a tera template over a runtime model. usage-argv has no `Spec` at
run time — only `&'static` tables — so the rules are reimplemented here, and reimplemented rules
drift. The test runs both over mise's real spec and compares every line, because an adopter's
help text changing is a visible regression even when the change is one bracket. The shadow is
generated from the same KDL usage-lib is handed, so the two describe the same CLI by
construction rather than by a fixture kept in step by hand.

The parity test is also what found the four metadata gaps: three in #853 below this, and one
here.

## What it caught here

**A `--` argument is bracketed around the separator.** usage-lib writes `[-- COMMAND]…` — one
optional thing — where I had `-- [COMMAND]…`, a literal `--` followed by an optional word. Five
of mise's commands read that way.

**A flag is named after the form it answers to, not the field holding it.** usage-lib derives
the name from the first long form; the derive was kebab-casing the Rust ident. So `type_` gave a
flag called `type-`, printed by help as `type-: -t --type` and reported by errors as `type-`.
An explicit `name` still wins, and a flag with no long keeps its short, as usage-lib does.
Selectors resolve by form rather than by name, so nothing that *refers* to a flag changes — and
error messages get better for free.

## Cost

None on the parse path: this reads the cold metadata a successful parse never touches, and lives
behind the existing `spec` feature. A binary that never renders help does not carry the
formatting code.

## Not yet

The line only. The sections below it — about, arguments, flags grouped by heading, subcommands,
examples — and the `--help`/`-h` wiring in the derive are the next two PRs. The line comes first
because everything else is arranged around it, and because it is the piece with an exact oracle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -793,6 +793,7 @@ dependencies = [
  "shadow-mise",
  "shadow-mise-clap",
  "usage-argv",
+ "usage-lib",
 ]
 
 [[package]]
--- a/PLAN.md
+++ b/PLAN.md
@@ -265,8 +265,16 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
 
 ### Then: what a CLI framework has to have
 
-- [ ] **Help rendering** — `--help` and `-h` from the static metadata, with
-      headings and hidden-item filtering.
+- [ ] **Help rendering** — `--help` and `-h` from the static metadata, with headings and
+      hidden-item filtering. The **usage line** is done: `usage_argv::help::usage_line` builds
+      it from the tables, and all 211 of mise's commands come out byte-identical to usage-lib's,
+      which is the standard that matters — an adopter's help changing is a visible regression
+      even when it is one bracket. usage-lib renders through a tera template over a runtime
+      model, which this crate has no equivalent of, so the rules are reimplemented and the
+      parity test is what keeps them honest. Getting there also forced a fourth naming fix: a
+      flag is named after the form it answers to, as usage-lib names it, not after the Rust
+      field holding it — `type_` had been giving a flag called `type-`, which help printed and
+      errors reported.
 - [ ] **Completions, self-contained** — `<bin> completion <shell>` emits the
       script; a hidden `<bin> complete-word` serves requests from the binary's own
       embedded spec. Same dispatch shape usage-cli uses today, without requiring
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -70,12 +70,12 @@ pub fn usage_line(path: &[&str], meta: &CommandMeta<'_>) -> String {
     // what a user is invited to type.
     let flags: usize = meta.flags.iter().filter(|f| !f.hide).count();
     if flags > 0 {
-        let required = meta.flags.iter().any(|f| !f.hide && f.required);
+        let required = meta.flags.iter().any(|f| !f.hide && flag_demanded(f));
         if flags <= INLINE_LIMIT {
             for flag in meta.flags.iter().filter(|f| !f.hide) {
                 // A required flag is angled, like a required argument: the brackets are what
                 // say whether leaving it out is allowed.
-                let (open, close) = if flag.required {
+                let (open, close) = if flag_demanded(flag) {
                     ('<', '>')
                 } else {
                     ('[', ']')
@@ -91,7 +91,7 @@ pub fn usage_line(path: &[&str], meta: &CommandMeta<'_>) -> String {
 
     let args: usize = meta.args.iter().filter(|a| !a.hide).count();
     if args > 0 {
-        let required = meta.args.iter().any(|a| !a.hide && a.required);
+        let required = meta.args.iter().any(|a| !a.hide && demanded(a));
         if args <= INLINE_LIMIT {
             for arg in meta.args.iter().filter(|a| !a.hide) {
                 let _ = write!(out, " {}", arg_usage(arg));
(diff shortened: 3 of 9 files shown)


jdx:
Real, and it goes to the heart of what this PR claims. usage-lib clears `required` while *parsing* a spec that declares a default, then renders the usage line from `required` alone — so it writes `[dir]` where this wrote `<dir>`, for an argument the parser is perfectly happy to omit. The same signal decides whether a collapsed list reads `<ARGS>…` or `[ARGS]…`, so a command of entirely defaulted arguments was marked required there too.

Fixed where the line is rendered rather than by clearing the flag in the metadata: the metadata is what the emitted spec is built from, and `required` there means what the author wrote.

Worth explaining why the 211-line parity test sails past this. Every defaulted positional in mise's spec is written `required=#false default=…`, so there is nothing to normalise and both sides already agree. A *derived* argument is required by its type and defaulted by its attribute — that is the shape that diverges, and mise's spec contains none. So the new test compares the two renderers over one derived declaration, and then parses it with no arguments at all to show the brackets are telling the truth.

Both call sites mutation-checked. The collapse threshold needed a case of its own: `INLINE_LIMIT` is 2, so it takes three defaulted arguments to reach the branch, and without one the mutation survived.

jdx:
Right — I fixed arguments and left flags reading `required` on its own, which is the same divergence in the other half of the line. usage-lib clears `required` on a flag that declares a default too, so a derived value flag with one was shown as `<--out>` when the parser fills it happily.

Both the brackets and the collapse threshold now use the same rule, and the fixture grew a defaulted flag plus three more so the collapsed `[FLAGS]` case is reached — without that third flag the threshold mutation survived, exactly as it did for arguments.

jdx:
Real, and worth doing — but the code it names is not in this PR's own commits, and I have deliberately not touched it.

The block that sets `name` from `shorts.first()` for a short-only flag arrived in a later commit on the branch this stack has since moved onto (`agent/help-short`/`agent/help-body`), and another session is actively editing that worktree — it committed under me while I was reading this file. Two of us writing there is how work gets lost, so here is the analysis instead of a half-applied fix:

The *name* is right and should stay: usage-lib derives it the same way, so `-j` alone is called `j` on both sides, and changing that would break the parity this stack exists for. The problem is only that `value_name` falls back to `name`, so the placeholder becomes `<j>` rather than something a reader can act on. The derive knows the field's own name, so defaulting `value_name` to it for a flag with **no long form and no explicit `value_name`** fixes the placeholder without touching what the flag is called — and a test wants a short-only value flag, checking both the emitted KDL and the usage line.

Happy to apply it here if that branch goes idle, or it can be picked up by whoever owns it.

jdx:
Right, and the fix belongs here rather than in a descendant — copying the field ident into the value name ran for *every* unnamed value flag, so a field called \`type_\` with \`long = "type"\` rendered \`--type <type->\`. Narrowed to the case it was for: a short-only flag, where the name becomes one character and \`-j <j>\` says nothing. A long form is already the descriptive name.

PR #856, opened 2026-08-13
feat(config): read config files as a layer

jdx:
Config files as a layer for `usage-config`: the rc-style chain a spec's `file` nodes describe — a system file, the user's own, then the project's, found by walking up from where the user is standing.

Every CLI in the fleet has written this walk by hand, and the walk is never the interesting part. The edges are, and each of these is a decision the fleet's copies make differently:

| case | what happens | why |
|---|---|---|
| the file is absent | nothing | a find-up chain is mostly directories with no config file in them |
| the file will not parse | **error** | the values the user believes are in effect are not; carrying on as though they wrote nothing is worse than saying so |
| a key nobody knows | warning, rest of the file applies | a file written for a newer binary must work with an older one, or upgrading is a coin toss — and silence makes a typo impossible to find |
| a value of the wrong type | warning, that key only | a typo in a *system-wide* file must not stop the CLI starting for every user on the machine |
| two files disagree | the nearest wins | the chain is ordered farthest-first because the merge takes the last writer |

`FileScope` is a required argument, not a defaulted one. A layer that forgot to say where it read from would be trusted as the operator's by accident — and that is precisely the check `scope="global"` exists for. Tested both ways: the same file refuses `trusted` when read as a project file and accepts it when read as the user's own.

## Two hooks the fleet actually needs

`under("settings")` reads from a table rather than the top level, because mise's settings share a file with its tools and tasks while hk's are at the top. Without it a CLI has to pre-extract the table — parsing the file twice, or teaching this layer the rest of the format. A file with no such table is not a broken file, and `[tools]` beside it is not reported as an unknown setting.

`preprocess` rewrites the text before parsing, which is where mise's tera goes without this crate learning what a template is. Text in, text out; an `Err` is a read failure naming the file, because a template that will not render is not something to carry on past.

## Values arrive as text

Deliberately. The *spec* decides what a value means: a `list<string>` with `parse="list_by_comma"` reads `"a,b"` as two items whether the file said so or not. A layer that pre-decided would disagree with the environment about the same setting — which is the class of drift this crate exists to end. Nested tables become dotted keys, which is how the registry is keyed.

## Still no dependencies by default

TOML and JSON are behind features, so `usage-config` with default features has an empty dependency tree — a CLI whose files are pkl or `.npmrc` writes its own layer against `Layer` and takes nothing it does not use. Built and tested in all four feature combinations; `cargo tree` confirms the default is bare.

Eleven tests. Four mutations of the properties that matter — the chain's order, the ceiling that stops it, a missing file, a file without the settings table — each verified to fail without its fix. One bug found while writing it: `str::parse::<toml::Value>` reads a single *value* in toml 0.9, so a whole document came back as "unexpected content, expected nothing"; it uses `toml::from_str`.

Next in this line: the `explain` renderer, then `usage-config-build` and the typed `Settings` struct.

---

> [!NOTE]
> **Medium Risk**
> New filesystem config path affects CLI startup and merge precedence; behavior is broad but gated by features and covered by extensive tests for symlinks, ceilings, and partial failure.
> 
> **Overview**
> Adds **`FileLayer`** to `usage-config` so TOML/JSON config files plug into the existing layer merge (optional **`toml`** / **`json`** features; default build stays dependency-free).
> 
> **Discovery & precedence:** load a fixed path or **`find_up`** with an inclusive ceiling; paths are normalized (symlinks, `..`, relative ceilings). Files are read farthest-first so the **nearest** file wins. Missing files are skipped; unreadable paths, parse failures, broken symlinks, and failed **`preprocess`** are hard errors.
> 
> **Semantics:** required **`FileScope`** for trust checks; unknown keys and per-key type mismatches become warnings. Scalars stay **text** for spec parsers; arrays and declared map/object settings use **`LayerCtx::entry_from_value`**. **`under`**, **`as_format`**, and **`preprocess`** support nested settings tables and extensionless files.
> 
> **`Ty`:** lists/maps no longer stringify into string-like types (supports shaped file values).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cabd065dc00fe965fddb020824169721269fff96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added optional TOML and JSON configuration file support.
  * Configuration files can be loaded from explicit paths or discovered through parent directories.
  * Added format selection, nested prefixes, preprocessing, scopes, and file precedence controls.
  * Structured values, including arrays and tables, are preserved during loading.
  * Missing files are ignored, while invalid files and processing errors are reported.
  * Unknown or incorrectly typed settings generate warnings without blocking other settings.
  * Format readers are opt-in and disabled by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1045,7 +1045,7 @@ dependencies = [
  "miette",
  "num-traits",
  "serde",
- "winnow",
+ "winnow 0.7.15",
 ]
 
 [[package]]
@@ -1597,6 +1597,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_spanned"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "serde_with"
 version = "3.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1946,6 +1955,45 @@ dependencies = [
 ]
 
 [[package]]
+name = "toml"
+version = "0.9.12+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf92845e79fc2e2def6a5d828f0801e29a2f8acc037becc5ab08595c7d5e9863"
+dependencies = [
+ "indexmap 2.14.0",
+ "serde_core",
+ "serde_spanned",
+ "toml_datetime",
+ "toml_parser",
+ "toml_writer",
+ "winnow 0.7.15",
+]
+
+[[package]]
+name = "toml_datetime"
+version = "0.7.5+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92e1cfed4a3038bc5a127e35a2d360f145e1f4b971b551a2ba5fd7aedf7e1347"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
+name = "toml_parser"
+version = "1.1.3+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d38ac1cf9b95face32296c0a3ede1fdc270627c9d9c02a7274dd6d960dc4d56"
+dependencies = [
+ "winnow 1.0.4",
+]
+
+[[package]]
+name = "toml_writer"
+version = "1.1.2+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d56353a2a665ad0f41a421187180aab746c8c325620617ad883a99a1cbe66d2"
+
+[[package]]
 name = "tracing"
 version = "0.1.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2038,6 +2086,10 @@ dependencies = [
 [[package]]
 name = "usage-config"
 version = "5.1.0"
+dependencies = [
+ "serde_json",
+ "toml",
+]
 
 [[package]]
 name = "usage-conformance"
@@ -2465,6 +2517,12 @@ checksum = "df79d97927682d2fd8adb29682d1140b343be4ac0f08fd68b7765d9c059d3945"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "winnow"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b97319f7b8343df12cc98938e5c3eb436064524c8d2b4e30a1d3a36eecdf81"
 
 [[package]]
 name = "xtask"
--- a/config/Cargo.toml
+++ b/config/Cargo.toml
@@ -11,9 +11,19 @@ authors = { workspace = true }
 license = { workspace = true }
 
 # Nothing here parses KDL: a spec is read at *build* time by usage-config-build, which emits
-# the registry as consts. This crate is what runs in the CLI, so it carries no spec parser,
-# no serde_json, and no format reader it was not asked for.
+# the registry as consts. This crate is what runs in the CLI, so it carries no spec parser
+# and no format reader it was not asked for.
 [dependencies]
+toml = { version = "0.9", optional = true }
+serde_json = { version = "1", optional = true }
+
+# Reading config files is opt-in, one format at a time, so the default is still no
+# dependencies at all: a CLI whose files are pkl or `.npmrc` writes its own layer against
+# `Layer` and takes nothing it does not use.
+[features]
+default = []
+toml = ["dep:toml"]
+json = ["dep:serde_json"]
 
 [package.metadata.release]
 shared-version = true
--- a/config/src/files.rs
+++ b/config/src/files.rs
@@ -1,0 +1,1298 @@
+//! Config files as a layer: where they are, and what reading one means.
+//!
+//! This is the rc-style chain a spec's `file` nodes describe — a system file, the user's own,
+//! then the project's, found by walking up from the working directory. Every CLI in the fleet
+//! has written this walk by hand, and the interesting part is never the walk: it is what
+//! happens at the edges. A file that does not exist is not an error. A file that exists and
+//! cannot be parsed *is*. A key nobody recognizes is a warning, because a config file written
+//! for a newer binary must still work with an older one. A value of the wrong type is a
+//! warning too, and only for that key — a typo in a system-wide file must not stop a CLI from
+//! starting for every user on the machine.
+//!
+//! Formats live behind features so the crate's default is still no dependencies at all: a CLI
+//! whose files are pkl or `.npmrc` writes its own layer against [`Layer`] and takes nothing it
+//! does not use.
+
+use std::path::{Component, Path, PathBuf};
+
+use crate::layer::{Layer, LayerCtx, LayerError, LayerOutput};
+use crate::source::{FileScope, Origin, SourceKind};
+use crate::value::Value;
+
+/// A hook that rewrites a file's text before it is parsed.
+///
+/// Named because clippy will not have it written inline, and it reads better here anyway: text
+/// in, text out, with the failure a string this crate does not interpret.
+type Preprocess = Box<dyn Fn(&str) -> Result<String, String> + Send + Sync>;
+
+/// How to read a file's bytes into keys and values.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Format {
+    #[cfg(feature = "toml")]
+    Toml,
+    #[cfg(feature = "json")]
+    Json,
+}
+
+impl Format {
+    /// The format an extension implies, when one does.
+    ///
+    /// A path with no extension, or one nobody claims, returns `None` — the caller names the
+    /// format itself rather than a guess being made on its behalf. `.myclirc` is TOML in one
+    /// CLI and INI in another, and neither is discoverable from the name.
+    pub fn of(path: &Path) -> Option<Self> {
+        match path.extension().and_then(|e| e.to_str()) {
+            #[cfg(feature = "toml")]
+            Some("toml") => Some(Self::Toml),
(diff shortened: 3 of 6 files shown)


jdx:
Three findings, all real, and the first is the one I got plainly wrong.

**Read errors were treated as a missing file** (Bugbot High, Greptile, CodeRabbit). `read_to_string` fails for far more than absence — permission denied, a directory at that path, content that is not UTF-8 — and all of them took the "not there, carry on" branch. This module's own doc says a file that exists and cannot be read *is* an error, so the code disagreed with the paragraph above it. Only `ErrorKind::NotFound` is silence now.

The test reads a **directory** where a file should be, rather than chmod'ing one: a permission bit proves nothing when the suite runs as root.

**The ceiling was exact path equality** (CodeRabbit, Greptile). A relative ceiling, or one reached through a symlink, matched no ancestor — so the walk ran to the filesystem root, reading files from above the boundary the caller asked for, which is the one thing a ceiling exists to prevent. Both sides are canonicalized now (best-effort: a path that does not exist yet is compared as written), and the walk also stops if it is no longer inside the ceiling. A ceiling that is not an ancestor at all now reads nothing rather than everything — of the two ways to be wrong about a boundary, that is the safer one.

**A `map` setting is a table, not a path to settings** (Bugbot). Flattening walked straight through it, so its entries became dotted keys nobody had declared: the map was never set and every key inside it was reported unknown — which also made `merge="deep"` unreachable from a file, and deep-merging map tables is a declared feature. Which tables are settings is the registry's answer, so the flattener now asks, and a declared table arrives already shaped rather than as text (there is no text a named parser could turn into one). Tested both that it arrives and that two files merge it key by key.

Three tests, three mutations, each verified to fail without its fix.

jdx:
Two more, both real, and the second one changed my mind about the design rather than patching it.

**A shaped value skipped the declared type** (Bugbot, CodeRabbit). The table path called `Entry::new` directly, so `Ty::coerce` never ran and a number inside a `map<string, string>` was stored rather than reported — the "a wrong type costs a warning, and only that key" promise broken for the settings that need it most. There is now a `LayerCtx::entry_from_value` beside `entry_for_key`, so a structured value goes through the declared type exactly as text does.

Writing the test for it found a second thing: `Ty::coerce` turned a *table* into the string `"nested=true"` for a text-typed setting. The rule it was following is a good one — a number written where text was expected is text that happens to look like a number, so `MISE_PYTHON_VERSION=3` should not fail — but a collection is not text, and rendering one produced a value nobody wrote. Narrowed to scalars.

**Arrays were joined into text and re-split** (CodeRabbit, Bugbot). Lossy in two ways that only appear alongside a spec:

- `exclude = ["a,b", "c"]` became `"a,b,c"`, and a comma parser then read three items where the file said two.
- A list declaring `list_by_colon` — or no parser at all — got *one* item holding the joined text.

So arrays keep the shape the file gave them. The rule underneath is now sharper than "values arrive as text": **a file has structure, and the named parsers are for sources that do not** — an environment variable. A *string* in a file still goes through the declared parser, which is what keeps `path = "/bin:/usr/bin"` meaning the same thing in a file and in `PATH`; the test asserts both halves.

Four mutations across the two, each verified.

jdx:
Right, and the finding sent me back to the predicate rather than adding a case to it — I had got it wrong in three ways, which is a sign the question was wrong:

- `Map | Object` only, so a table under a `Ty::Any` key — what a union or a tool-private type becomes — was walked into and every key inside reported unknown.
- An **empty** table under a scalar key produced no keys at all, so nothing was set and nothing was said.

A table is a *path* to settings until it is a setting itself, and the only thing that knows which is the registry. So the question is now simply "does this dotted key name a declared setting", and the declared type answers the rest: a `map` takes the table, a union takes it, a `uint` refuses it — and that refusal *is* the warning the rule promises. One predicate, three cases, nothing enumerated.

Two mutations — restricting it back to `Map | Object`, and never keeping a table — fail four tests between them.

jdx:
Right. `under("settings")` only checked whether the key was *present*, so a file whose `settings` is a scalar or an array was flattened from there — producing an **empty key**, reported as an unknown setting called nothing, while the settings that really were in the file went unread.

It is an error now, naming the key and the file. A file that is wrong in the one place this layer was pointed at is wrong in the same way as one that will not parse: the values the user believes are in effect are not.

```
could not read wrong.toml: `settings` should be a table of settings, and is not
```

Absent is still silence, which is the case the check was written for in the first place. Mutation-verified.

jdx:
Two findings; one real, one I could not reproduce.

**JSON `null` was stored as the text `null`** (Bugbot) — real. `null` is how a JSON file says "no value", so reading it as text meant a string setting held the *word* `null` and a list setting held one item of it: an optional field explicitly nulled *set* the setting instead of leaving it alone. A null key is now a key that is not there, and the same rule applies inside a declared table or list. Two mutations, both verified.

**Dead code with no format feature** (CodeRabbit) — does not reproduce. `files.rs` is gated at the module level in `lib.rs`:

```rust
#[cfg(any(feature = "toml", feature = "json"))]
pub mod files;
```

so with no feature there is nothing to warn about. Checked all four configurations rather than reasoning about it:

```console
$ cargo clippy -p usage-config --no-default-features --all-targets -- -D warnings   # clean
$ … --features toml    # clean
$ … --features json    # clean
$ … --features toml,json   # clean
```

If a warning shows up in a configuration I have not tried I would rather fix it than argue, so do point at one.

jdx:
Three findings, all real, and two of them are the same rule I only half-applied last round.

**`null` is a key that is not there — including the settings table itself.** I made `null` mean
"absent" everywhere inside a file, then left `under("settings")` checking only `is_object()`, so
`"settings": null` was the one spelling of "no settings table" that failed the read — while a file
omitting the table entirely was fine. Two ways to write the same thing, one of them a startup
error. `Some(Null) | None` now share the arm, above the `Some(_)` error.

**A JSON root that is not an object became the empty key.** `[1, 2]` or `"text"` flattened to an
entry keyed `""`, surfacing as ``unknown setting ` ` `` — a warning that names nothing, about a
file that is not a settings file at all. It's an error now, with the same message shape as the
prefix case. TOML can't reach this (its root is always a table), and a root of `null` still means
a file that says nothing rather than a file that is wrong.

**A ceiling that does not exist read no files at all.** This one is worse than the finding says.
The comment right above the code claims canonicalizing both sides is what stops a relative ceiling
from silently walking past the boundary — but `canonicalize` fails on a path that isn't on disk
yet, and the fallback kept that side relative. One relative side makes every `starts_with` false,
so the walk broke *before its first step* and the layer contributed nothing: not "read too much"
but "read nothing", which looks exactly like a project with no config in it. The fallback is
`std::path::absolute` now — the same normalization without the filesystem, so both sides are
comparable whether or not the directory exists yet. (Available since 1.79; the crate's
`rust-version` is 1.80.)

Each fix has a test that fails when the fix is reverted — I checked all four mutations, including
both directions of the root guard (rejecting a `null` root also fails). 60 tests in the crate,
workspace green, clippy clean with no features, `toml`, `json`, and both.

jdx:
Three more, and the third one is a hole my own last fix opened.

**A dangling symlink was read as no file.** `read_to_string` reports `NotFound` about the link's
*target*; the link itself is plainly there. This module's rule is that absence is normal and a file
that is there and cannot be read is an error — and a link into a directory the user has since moved
is the second case wearing the first case's error code. It now asks `symlink_metadata`, which is
the question actually being asked ("is there something at this path"), and only on the miss, so the
common no-file-here case pays one `lstat`.

**A ceiling that is not there *and* holds a `..` still read nothing.** `std::path::absolute` fixed
the relative half but leaves `..` in place as a component of its own, so the comparison stayed
false. Both sides now go through one `normalize`, and it does what canonicalizing cannot do alone:
resolve `.`/`..` lexically, then canonicalize *the deepest ancestor that actually exists* and put
the missing tail back on the end. That ordering is the point — a `..` resolved lexically through a
symlink would be wrong, and the components below the deepest existing ancestor don't exist, so they
cannot be links to anywhere. My previous comment claimed the prefix behavior before the code did
it; it does it now, and there is a test that fails if the pass is removed.

**A non-object JSON root was rejected without `under()` and ignored with it.** Last round I made a
list-or-scalar root an error — in the no-prefix branch only. With a prefix, `get("settings")` on a
list root answers `None`, which is indistinguishable from a file that simply has no settings table,
so the same wrong file was silently empty. The root question is asked once now, before the prefix
lookup, since a root that is not an object has no keys to read either way.

Four mutations, four dead tests: the dangling link read as absent, the root check removed, the
fallback left merely absolute (the `..` case), and the existing-prefix canonicalization dropped
(the link-then-missing-directory case). The two symlink tests are `cfg(unix)`. 63 tests in the
crate, workspace green, clippy clean with no features, `toml`, `json`, and both.

jdx:
Both right, and both are the deeper version of what I fixed an hour ago — the shallow case only.

**`link/..` was resolved in the wrong order.** My own doc comment on `normalize` says the thing
string work cannot do is `a/link/..`, which is wherever `link` points and not `a` — and then the
fallback resolved every `..` lexically *before* touching the filesystem, which is exactly that
mistake. Fixed by resolving one component at a time and canonicalizing each as it goes: a link is
followed before the `..` after it is applied, and a component that does not exist cannot be a link
to anywhere, so stepping back out of *that* lexically is right. This also subsumes the
deepest-existing-ancestor pass from the last round, so the function got shorter.

**A file below a dangling link was still silent.** `symlink_metadata` on the configured path
answers `NotFound` when the broken link is its parent — `~/.config/hk` pointed at a drive that is
no longer mounted, with `hk.toml` under it. The question is now asked of the path *and its
ancestors*, stopping at the first thing that exists: if that thing is a link that cannot be
followed, the read fails and the message names **the link**, which is what the user has to go and
fix. A directory nobody ever created still stays silent — that is the common case, and stopping at
the first existing ancestor is what keeps the two apart. The walk only runs after a read has
already failed, so the cost is a few `lstat`s on a miss.

Three mutations, each killing the right test: components pushed without following links (both the
`..`-through-a-link case and the missing-directory-under-a-link case), the ancestor walk removed,
and a plain missing file reported as broken (which also breaks the existing silence test). 65 tests,
workspace green, clippy clean in all four feature configurations.

PR #857, opened 2026-08-13
feat(config): explain why a setting has the value it has

jdx:
`config explain` for every adopter, as a renderer rather than a reimplementation. Second PR of stack #858; depends on #856 only for sitting on top of it.

hk's version is the best of its kind in the fleet — it names the winning source and the exact identifier, and does per-item provenance for lists. It also costs about two hundred lines, and it is written against a **second merge function** that exists only to answer this question. Two merges can disagree, and when they do the explanation describes a resolution that never happened. Here there is one merge and its provenance *is* the answer, so this module reads what the merge recorded and formats it.

```
jobs = 8
  set by  HK_JOBS
  type    uint
          How many jobs to run at once

  also considered, lowest precedence first:
    the default
    hk.toml#jobs

  environment  HK_JOBS, HK_JOB
  also         git hk.jobs
```

## What the wording is careful about

Each of these sends a user to the wrong place if it is careless:

- **A default is not something anybody set**, and neither is a post-merge rewrite — `default` and `derived` rather than `set by`. mise's `raw` implying `jobs = 1` must not read as though a file said so, or the user goes looking for the file.
- **The winner is not repeated** among the things it beat.
- **Asking after an old name answers about both**, and reads the deprecation notice from the declaration that *has* one — the old one. Reading it off the setting that replaced it printed nothing for the only case where it matters, which the test caught.
- **The type is the spec's spelling** (`uint`, `list<string>`), not the prose an error message uses. A reader searching the docs for "a positive integer" finds nothing. That is a new `Ty::name`, distinct from the existing `describe`.

Also `warnings`, pairing each message with the place that caused it, and `list` for a `config ls` — sorted by key, because a registry's order is the order somebody wrote a TOML file in, and excluding hidden settings and old names, which are documented nowhere and would surface here for the first time.

Plain text on purpose: a CLI that wants JSON has the same `Resolved` this reads, and better taste than a library about what its own output should look like.

Eight tests, five mutations — the verbs, the winner appearing among the also-considereds, the deprecation source, hidden settings in the listing, and the sort — each verified to fail without its fix.

---

> [!NOTE]
> **Low Risk**
> User-facing text and warning formatting only; resolution merge logic is unchanged aside from clearer type-error strings.
> 
> **Overview**
> Adds shared **`config explain`** rendering on top of existing merge provenance in [`Resolved`](config/src/resolve.rs)—no second merge path. New **`explain`**, **`warnings`**, and **`list`** format plain text for one key, resolution warnings, and `config ls` (sorted keys, hidden/renamed keys omitted).
> 
> **`explain`** prints the winning value with **`shown`**, provenance verbs (`default` / `derived` / `set by`), spec types via new **`Ty::name`**, lower-precedence contributors, env/bindings hints, and deprecation along rename chains.
> 
> Supporting changes: **`one_line`** / **`shown`** in `value.rs` keep line-oriented output safe (newlines, empty `[]` / `""` / `{}`); layer type-coercion warnings drop duplicated origin text so **`explain::warnings`** can append `(origin)` once; **`Ty`** coercion errors quote values with **`shown`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59e00a7b2b56b6436cf9dd386ddc172a09602f74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config/src/explain.rs
+++ b/config/src/explain.rs
@@ -1,0 +1,634 @@
+//! Why a setting has the value it has.
+//!
+//! hk's `config explain` is the best of its kind in the fleet — it names the winning source and
+//! the exact identifier, and it does per-item provenance for lists. It also costs about two
+//! hundred lines, and it is written against a *second* merge function that exists only to
+//! answer this question. Two merges can disagree, and when they do the explanation describes a
+//! resolution that never happened.
+//!
+//! Here there is one merge and its provenance is the answer, so this module is a renderer and
+//! nothing more. Every adopter gets the command; none of them writes it.
+//!
+//! The output is plain text on purpose. A CLI that wants JSON has the same [`Resolved`] this
+//! reads and better taste than a library about what its own JSON should look like.
+
+use std::fmt::Write as _;
+
+use crate::registry::{PropId, Registry};
+use crate::resolve::Resolved;
+use crate::source::SourceKind;
+use crate::value::{one_line, shown};
+
+/// One setting explained: what it is, where it came from, and what else tried.
+///
+/// ```text
+/// jobs = 4
+///   set by  MISE_JOBS
+///   type    uint
+///
+///   also considered, lowest precedence first:
+///     the default
+///     mise.toml#jobs
+/// ```
+///
+/// Returns `None` for a key the registry does not have, which the caller reports its own way —
+/// a CLI usually wants to suggest a near miss, and this module has no business guessing.
+pub fn explain(resolved: &Resolved, key: &str) -> Option<String> {
+    let registry = resolved.registry();
+    let found = registry.lookup(key)?;
+    let meta = registry.get(found.id);
+    let mut out = String::new();
+
+    // The name the *user* asked about, then the setting it turned out to be. Printing only the
+    // canonical name would answer a question nobody asked.
+    if let Some(old) = found.renamed_from {
+        let _ = writeln!(out, "{old} is now {}", meta.key);
+    }
+
+    match resolved.get(found.id) {
+        Some(value) => {
+            let _ = writeln!(out, "{} = {}", meta.key, shown(value));
+        }
+        None => {
+            let _ = writeln!(out, "{} is unset", meta.key);
+        }
+    }
+
+    if let Some(origin) = resolved.origin(found.id) {
+        // "set by the default" reads oddly, and a rewrite is not something anybody *set*.
+        let verb = match origin.kind {
+            SourceKind::DEFAULTS => "default",
+            SourceKind::COERCED => "derived",
+            _ => "set by",
+        };
+        let _ = writeln!(out, "  {verb:<8}{}", one_line(origin.describe()));
+    }
+    // The spec's own spelling, not the prose an error message uses: a reader searching the docs
+    // for "a positive integer" finds nothing, and `uint` is what the author wrote.
+    let _ = writeln!(out, "  {:<8}{}", "type", meta.ty.name());
+    if let Some(help) = meta.help {
+        // Through the same helper as everything else: an adopter's help is a doc comment, and a
+        // doc comment with a second paragraph in it split one fact into several records.
+        let _ = writeln!(out, "  {:<8}{}", "", one_line(help));
+    }
+
+    // Everything that contributed, which for a `union` or `deep` setting is how a list assembled
+    // from four files is accounted for. The winner is the last of them, so printing it again
+    // would be noise — but for anything with more than one contributor, *which* others there
+    // were is the question being asked.
+    let contributors = resolved.contributors(found.id);
+    if contributors.len() > 1 {
+        let _ = writeln!(out, "\n  also considered, lowest precedence first:");
+        for origin in &contributors[..contributors.len() - 1] {
+            // No padding: the column it was aligning for is one this crate cannot fill. A
+            // contributor's *value* is not kept — provenance is origins — so the width only ever
+            // added trailing spaces to the last thing on the line, and a path copied out of an
+            // explanation stopped being the path the merge recorded.
+            let _ = writeln!(out, "    {}", one_line(origin.describe()));
+        }
+    }
+
+    // Where else it *could* come from, which is the other half of the question: a user who does
+    // not like the answer needs to know what to change.
+    // The blank line opens whichever of these comes first, rather than belonging to the
+    // environment: a setting with git or pkl bindings and no environment variables had its
+    // `also` line jammed against the type or the help above it, reading as part of them.
+    if !meta.envs.is_empty() || !meta.bindings.is_empty() {
+        let _ = writeln!(out);
+    }
+    if !meta.envs.is_empty() {
+        let _ = writeln!(out, "  environment  {}", one_line(&meta.envs.join(", ")));
+    }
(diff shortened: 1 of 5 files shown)


jdx:
Both real, and both about output that lies about its own shape.

**A newline in a value broke the record boundary.** Both renderers here are line-oriented — one setting per line for a listing, one fact per line for an explanation — and a multi-line string is a perfectly ordinary thing to put in a TOML file. Its continuation read as another setting, or as provenance. Values now render with newlines (and carriage returns, and the backslashes that would be ambiguous) escaped.

**A setting with bindings and no environment variable had a dangling `also`.** The blank line belonged to the environment section, so a git- or pkl-bound setting with no variables had its `also` line jammed against the type or help above it. The blank line now opens whichever section comes first.

Three mutations, each verified. The listing test asserts the line count as well as the content, so a value that splits into two records fails rather than merely looking odd.

jdx:
Right, and the fix belongs on the message rather than the renderer: a type error named the place in its own text *and* the `Warning` carried it, so the rendered line said the same file twice — for exactly the warnings that had bothered to be specific.

The message now says what is wrong and the renderer says where, which also means every warning renders the same way instead of two kinds needing two treatments. Test asserts the place appears exactly once; mutation-verified by putting it back.

jdx:
Both right, and both the same class as the value fix — applied to the other two things interpolated into a line-oriented output.

An **origin** can carry a newline (a path may contain one), and a **warning message** quotes the value it rejected, which comes out of a file and can hold anything. Either one splitting its line makes the remainder read as another record; for warnings it hides every warning after it. So the escaping moved off the value and onto the shared helper, and all three go through it.

Two things about the verification. My first test covered only one of the two places an origin is rendered — the winner's line and the also-considered list are different branches — and the mutation for the other survived; both paths now carry a newline in the test. And I pushed once with a clippy failure in between (an import left unused by the change), because my gate command let the push run after the check reported. Caught on the next pass and fixed before this push; the branch is clean now.

jdx:
Right, and worse than the problem it was solving: `C:\Users\me\hk.toml` came out with doubled separators, so a reader copying the path lands nowhere. The escaping now covers newlines and carriage returns only. The ambiguity that buys — a literal `\n` in text is indistinguishable from an escaped newline — is the right trade for output a human reads, whose one requirement is that a record stays on its line.

Test asserts a Windows path round-trips as written, and the mutation that reintroduces backslash doubling fails.

jdx:
Right — same class again, and this time for the string most likely to contain a newline: an adopter's help is a doc comment, and a doc comment with a second paragraph is the ordinary case rather than an exotic one. Help, the deprecation notice, and the joined environment and binding lists all go through the helper now.

The test asserts the *line count* of a whole explanation rather than only the content, so anything splitting a record fails whether or not I thought to look for it. Both mutations verified.

That is four rounds on one small renderer, each finding a different string I had not routed through the escaping — which is a fair signal that the helper should have been applied at the point where a line is written rather than at each interpolation. If a fifth turns up I will restructure it that way instead of patching another call site.

jdx:
Right, and it is the shape I only half-handled. Reading the notice off the *replacement* printed
nothing for the case that matters (a rename with a notice on the old name); reading it off the
asked-about name alone prints nothing when the notice sits further along — `threads` renamed to
`concurrency`, and `concurrency` the step carrying "use jobs instead". A rename is not itself an
explanation of what to do instead, so the notice worth printing is the first one along the chain.

`deprecation_along` walks from the asked-about name and stops at the first notice, bounded by the
number of settings there are — the same guard `Registry::lookup` uses, so a registry whose renames
form a cycle stops instead of hanging. (`usage-config-build` will refuse that registry outright;
until then, nothing here can be made to loop.)

Also in this commit: `one_line` moved to `value.rs` as a crate-internal helper, because #862 needs
the same rule for its failure lines and one rule shared by every renderer is one rule.

79 tests. Mutation: restoring the old `lookup_exact`-only read fails the new test.

jdx:
One real, one I am going to leave as it is with the reasoning.

**The padding was aligning a column that does not exist.** Real, and worse than trailing
whitespace: the doc example above `explain` shows each contributor with its *value* beside it —
`the default        4` — and the code cannot print that, because provenance keeps origins and not
values. So the width only ever padded the last thing on the line, and a path copied out of an
explanation stopped being the path the merge recorded. The padding is gone, the doc example now
shows what the function actually prints, and the contributors test asserts no line of an
explanation ends in whitespace.

**The `\n` ambiguity I am keeping.** It is real — a rendered newline and a literal backslash-`n`
look the same — but it is the deliberate trade recorded in the comment on `one_line`, and I got
there by trying the alternative. Escaping backslashes too renders every Windows path with doubled
separators (`C:\\Users\\me\\hk.toml`), so the common case becomes a path the reader cannot copy in
order to disambiguate the pathological one (a real newline inside a path). This output is for a
human, and the property it needs is that one record stays on one line. A CLI that needs values
unambiguously has the same `Resolved` and can render JSON, which is the note already in the module
doc.

79 tests. Mutation: restoring the padding fails the new assertion.

jdx:
Right, and my own no-trailing-space assertion from the last round did not catch it — the fixture
had no empty value in it, and an empty value is a *value*.

`Value::display` writes a value the way a user would type it, and three of them are typed as nothing
at all: the empty string, the empty list, the empty map. Interpolated into `key = {}` that ends the
line after the `=`, with a trailing space and a truncated look — for a state that is perfectly
ordinary, since clearing a list is how a declared default gets turned off (`HK_EXCLUDE=`). Set is
not unset, and both are worth saying:

```
exclude = []
  set by  HK_EXCLUDE
stash = ""
```

The rule is one helper, `shown`, beside `one_line` in `value.rs`, and it is applied everywhere a
value is quoted into a message a human reads — `explain`, `list`, and the type errors `ty::coerce`
builds, which the merge's warnings render and which had the same "expected a list but has ``"
problem. Deliberately **not** folded into `display` itself: that is also what `config get` prints,
where an empty setting printing nothing is exactly right and `[]` would be a value nobody wrote.

The same fix lands in #862 for the reader's own errors, since that module quotes values the same way
and is in the same stack.

80 tests. Mutation: returning the value's text unconditionally fails the new test, which covers the
empty list and the empty string, in an explanation and in a listing, and re-checks that no line ends
in whitespace.

jdx:
Right, and worse than the finding says: read from the *text*, a one-item list holding the empty
string was indistinguishable from a cleared one, and two of them came out as `,`.

Emptiness is now asked of the value, not of its rendering. And the underlying problem is that a list
is joined with commas, which hides any item whose own text is empty — so when an item would
disappear, the whole list is written out instead:

```
exclude = []          # cleared
exclude = [""]        # one item, which is not the same thing
exclude = ["",""]     # two, which used to be `,`
exclude = [a,""]      # the item the comma form would lose
exclude = a,b         # nothing to hide, so nothing to bracket
```

A map keeps its plain form even when a value is empty, because its keys are in the text (`k=`) and
that is enough to read.

Two mutations: not writing out a list whose items would vanish, and not marking the empty list, each
killing a test — the value test directly, and the explanation test through it.

PR #860, opened 2026-08-13
feat(argv): render `-h`, byte-identical to usage-lib's

jdx:
The whole short help page, not just the line above it: the program and version, the about,
the commands list, arguments and flags grouped by heading, examples, and the hidden-item
filtering. All 211 of mise's commands come out byte for byte identical to usage-lib's, which
is the standard that decides whether an adopter's help changes.

Holding it to that found four more things a spec could say that the derive could not, each
dropped by `gen-shadow` without being counted — the same class as the three in the PR below
this one, and found the same way:

- the spec's own `about`. The root's description lives at the top level, not on the root
  command, so the shadow's root said "Undocumented" where mise says what mise is.
- `hide` on a command. mise hides eight, `asdf` and `dotfiles` among them, and a
  `Subcommands` variant had no way to say so — help listed every one of them.
- help text whose line breaks are meant literally. A doc comment's first paragraph is read
  the way Rust reads one, so a break inside it becomes a space; 37 of mise's flags and
  commands have multi-line help and every one came back with its lines run together.
  `#[usage(help = "…")]` declares it instead, rather than changing what a comment means.
- the `Undocumented` placeholder the generator wrote where a spec says nothing. It became
  *real* help text: the shadow told a reader that `--output` was "Undocumented".

Two shapes the reference has that are worth knowing about, both matched rather than
improved: a listed subcommand shows its whole path from the root (`tool-alias get <TOOL>`,
not `get <TOOL>`), and a command whose subcommands are all hidden still says `<SUBCOMMAND>`
in its usage line while printing no commands section, because usage-lib computes that line
before filtering.

---

---

> [!NOTE]
> **Low Risk**
> Changes are help rendering and test/shadow fixtures only; no parser hot-path or binding behavior.
> 
> **Overview**
> Adds **`short_help`** in `usage_argv::help`, producing the full **`-h`** page from static metadata: version line, spec **`about`**, usage line, subcommands (full path, aliases, synthetic **`help`** entry), arguments and flags grouped by **`help_heading`** with choices/env/default annotations, and examples—with **hidden** items filtered like usage-lib.
> 
> A gate test walks all **211** mise commands and compares output to **`usage::docs::cli::render_help(..., false)`**, reporting the first differing line on failure.
> 
> The **mise clap shadow** is updated so generated help matches: root **`about`**, **`hide`** on subcommands, **`#[usage(help = "...")]`** (and **`long_help`**) where doc comments would collapse line breaks, and removal of **`Undocumented`** placeholders that had become visible help text. **PLAN.md** marks **`-h`** parity done and notes **`--help`** layout plus derive wiring as next.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f43d20a226b9b1b60ee1e711311f0d2c8473ba2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## The standard

**All 211 of mise's commands render byte-identically to usage-lib** — the whole `-h` page, not
just the line at the top: program and version, about, commands list, arguments and flags grouped
by heading, examples, hidden-item filtering.

usage-lib renders through a tera template over a runtime model. usage-argv has only `&'static`
tables, so the rules are reimplemented, and the parity test over mise's real spec is what keeps
them honest. It reports the **first differing line** rather than two walls of text, which is what
made working through 211 → 74 → 70 → 37 → 3 → 0 tractable.

## Four more gaps it found

Same class as the three in #853, found the same way, and each dropped by `gen-shadow` without
being counted:

| | what happened |
|---|---|
| the spec's own `about` | the root's description lives at the top level, not on the root command, so the shadow's root said **"Undocumented"** where mise says what mise is |
| `hide` on a command | mise hides eight (`asdf`, `dotfiles`, …) and a `Subcommands` variant had no way to say so — help listed every one |
| multi-line help | a doc comment's first paragraph is read the way Rust reads one, so a line break inside it becomes a space; **37** of mise's flags and commands have multi-line help and all came back run together |
| the `Undocumented` placeholder | written where a spec says nothing, it became *real* help text — the shadow told readers `--output` was "Undocumented" |

For the third I added `#[usage(help = "…")]` rather than changing what a doc comment means: a
comment's paragraph should still flow, as clap reads it too, but help whose line breaks are
deliberate now has a way to say so. Useful to adopters beyond the shadow.

## Two reference shapes matched rather than improved

- A listed subcommand shows its **whole path from the root** — `tool-alias get <TOOL>` under
  `mise tool-alias`, not `get <TOOL>`.
- A command whose subcommands are all hidden still says `<SUBCOMMAND>` in its usage line while
  printing no commands section, because usage-lib computes that line before filtering and
  stores it.

Both are odd; both are what the reference does, and parity is the point. Worth revisiting
together later, in usage-lib.

## Next

`--help`'s wider layout — help aligned into a column, wrapped to `COLUMNS`, `[possible values:]`
on its own line — then the `--help`/`-h` wiring in the derive: `Error::Help`, `-h` short and
`--help` long, and a real `help` subcommand for any CLI that has subcommands.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -265,16 +265,14 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
 
 ### Then: what a CLI framework has to have
 
-- [ ] **Help rendering** — `--help` and `-h` from the static metadata, with headings and
-      hidden-item filtering. The **usage line** is done: `usage_argv::help::usage_line` builds
-      it from the tables, and all 211 of mise's commands come out byte-identical to usage-lib's,
-      which is the standard that matters — an adopter's help changing is a visible regression
-      even when it is one bracket. usage-lib renders through a tera template over a runtime
-      model, which this crate has no equivalent of, so the rules are reimplemented and the
-      parity test is what keeps them honest. Getting there also forced a fourth naming fix: a
-      flag is named after the form it answers to, as usage-lib names it, not after the Rust
-      field holding it — `type_` had been giving a flag called `type-`, which help printed and
-      errors reported.
+- [ ] **Help rendering** — `-h` is done and matches usage-lib byte for byte across all 211 of
+      mise's commands: the usage line, the about, the commands list, arguments and flags grouped
+      by heading, examples, hidden-item filtering. `--help`'s wider layout, which aligns help
+      into a column and wraps to `COLUMNS`, is next, then the `--help`/`-h` wiring in the derive.
+      Holding it to parity found six more things a spec could say that the derive could not —
+      value names, required collections, `var` on a count, the spec's own `about`, `hide` on a
+      command, and help text whose line breaks matter — plus a bug in usage-lib, which printed
+      everything marked `hide`.
 - [ ] **Completions, self-contained** — `<bin> completion <shell>` emits the
       script; a hidden `<bin> complete-word` serves requests from the binary's own
       embedded spec. Same dispatch shape usage-cli uses today, without requiring
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -19,7 +19,7 @@
 
 use core::fmt::Write as _;
 
-use crate::spec::{ArgMeta, CommandMeta, FlagMeta};
+use crate::spec::{ArgMeta, CommandMeta, FlagMeta, Spec};
 use crate::DoubleDash;
 
 /// How many flags or arguments are listed individually before collapsing to a placeholder.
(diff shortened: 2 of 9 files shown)


jdx:
Both right, and the same bug: `declared_help` was handed the flag's long *name* where its long *help* belongs, so `--shims` emitted `long_help = "shims"` and the shadow's regenerated spec said the flag's own name where mise's says a paragraph. Fourteen flags were affected.

Fixed in `0bccd88`, with a generator test. Worth noting what the test needed: the short help has to be *multi-line* to reach the bug at all, because one-line help is written as a doc comment and `declared_help` is only called for help a comment cannot carry. A single-line fixture left the mutation alive.

jdx:
Correct, and it is the silent-drop class this generator exists to prevent — so the fix is here, in the PR that introduced the skipping, rather than further up where I first put it.

`needs_declaring` skips the comment for *both* dialects and only the usage one was writing the text as an attribute, so every command, flag and argument whose help a comment would change lost its description from the clap shadow. clap can say all of it: `help`/`long_help` on an argument, `about`/`long_about` on a command. Both shadows now hold the same 77 declared descriptions, and the test asserts that property rather than one instance.

PR #862, opened 2026-08-13
feat(config): read a resolution as the types a struct holds

jdx:
The half of the codegen that belongs at runtime: turning a resolution into the types a settings
struct holds. `usage-config-build` will generate the struct and call this; splitting it out means
the conversion rules are tested against a real merge rather than through generated code.

## Why a typed read can fail at all

The merge already validates every value against the type its setting declares, so most of this
cannot fail. Three things keep it from being a formality — and they are why these errors carry
provenance instead of being `unwrap`s:

- **A post-merge hook** writes with `Resolved::coerced`, which is deliberately unchecked: the hook
  is where a CLI puts the rules only it knows, and validating there would make it a second merge.
  A hook that writes `-1` to a `uint` is a bug in the CLI, and the message has to say which hook —
  mise's post-merge coercions touch a dozen settings, and "expected a positive integer" alone
  would not tell you which one did it.
- **A type only the tool understands** (`any`, a union) is not coerced by the merge at all, by
  declaration. The field that holds it is still concrete, so the field type is the only thing that
  says what belongs there.
- **The field narrows further** than the declared type: `uint` is an `i64` in the merge and a `u64`
  in the struct.

## What it does

`FromValue` is **strict** — it never converts. Coercion happens once, in the merge, against the
declared type; a reader that also converted would be a second set of rules for what a value means,
which is the drift this crate exists to remove. `bool`, `i64`, `u64`, `f64`, `String`, `PathBuf`,
`Vec<T>`, `BTreeMap<String, T>`. A `set` reads as a `Vec`: the merge has already dropped the
duplicates, and a list keeps the order the files were read in — which a sorted set would throw
away for something like a `PATH`, where the order *is* the meaning.

`Fold` collects **every** failure instead of returning the first. The fleet's hand-written folds
stop at one, so a config file with three bad values takes three runs to fix:

```
jobs expected a positive integer but has `-1` (set by one job when raw)
raw expected a boolean but has `sometimes` (set by a hook)
ports expected a positive integer but has `-443` (set by a hook)
profile has no value and no default
```

`required` reports a non-`Option` field with nothing to hold, rather than unwrapping it — that
case is the CLI's own registry being wrong, and the author needs the key. It returns `None` only
when it has recorded an error, which is the documented contract generated code is written
against: the fold reports, then the struct is built.

`u64` goes through `try_from`, not `as` — a negative one becoming an enormous positive one is the
shape this bug takes in every codebase that has it.

## Verification

81 tests in the crate. Five mutations, each reverting one decision and each killing a test:
`as` in place of `try_from`, the origin dropped from the error, `finish` returning only the first
error, the missing-value check removed, and a bad list item skipped instead of reported. Clippy
clean with all features and with none.

---

> [!NOTE]
> **Low Risk**
> Additive API and wording fixes; behavior is covered by extensive unit tests in `read.rs` with low blast radius for existing merge consumers.
> 
> **Overview**
> Adds **`config/src/read`** so a merged [`Resolved`](config/src/resolve.rs) can be turned into the Rust types a settings struct uses—intended for **`usage-config-build`** generated code at runtime.
> 
> **`FromValue`** maps [`Value`](config/src/value.rs) to concrete types (`bool`, integers, `f64`, `String`, `PathBuf`, `Vec`, `BTreeMap`) without re-coercing; **`Resolved::fold`** / **`Fold`** collect every read failure (type mismatch, missing required value) with **origin** in the message, and **`Resolved::read`** covers single-setting use. **`u64`** uses **`try_from`** so negative `uint` values from unchecked **`coerced`** hooks fail visibly.
> 
> **`uint`** user-facing copy is aligned to **"a non-negative integer"** in [`Ty::describe`](config/src/ty.rs), layer warnings, and explain docs. New exports: **`Fold`**, **`FromValue`**, **`ReadError`**, **`ReadErrors`** from the crate root.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1c730338c85091e11df74aa7602d42df6b5ae38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config/src/explain.rs
+++ b/config/src/explain.rs
@@ -64,7 +64,7 @@ pub fn explain(resolved: &Resolved, key: &str) -> Option<String> {
         let _ = writeln!(out, "  {verb:<8}{}", one_line(origin.describe()));
     }
     // The spec's own spelling, not the prose an error message uses: a reader searching the docs
-    // for "a positive integer" finds nothing, and `uint` is what the author wrote.
+    // for "a non-negative integer" finds nothing, and `uint` is what the author wrote.
     let _ = writeln!(out, "  {:<8}{}", "type", meta.ty.name());
     if let Some(help) = meta.help {
         // Through the same helper as everything else: an adopter's help is a doc comment, and a
--- a/config/src/layer.rs
+++ b/config/src/layer.rs
@@ -336,7 +336,7 @@ mod tests {
         // twice, for exactly the warnings that had bothered to be specific.
         assert_eq!(
             warning.message,
-            "jobs expected a positive integer but has `lots`"
+            "jobs expected a non-negative integer but has `lots`"
         );
         assert_eq!(warning.origin, Some(origin));
 
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -73,6 +73,7 @@ pub mod explain;
 #[cfg(any(feature = "toml", feature = "json"))]
 pub mod files;
 pub mod layer;
+pub mod read;
 pub mod registry;
 pub mod resolve;
 pub mod source;
@@ -83,6 +84,7 @@ pub use explain::explain;
 #[cfg(any(feature = "toml", feature = "json"))]
 pub use files::{FileLayer, Format};
 pub use layer::{Entry, Layer, LayerCtx, LayerError, LayerOutput, Warning};
+pub use read::{Fold, FromValue, ReadError, ReadErrorKind, ReadErrors};
 pub use registry::{Lookup, Merge, PropId, PropMeta, Registry, Scope};
 pub use resolve::{resolve, Layers, Resolved};
 pub use source::{FileScope, Origin, SourceKind, Trust};
--- a/config/src/read.rs
+++ b/config/src/read.rs
@@ -1,0 +1,592 @@
+//! A resolution read as the types a settings struct holds.
+//!
+//! The merge already validates every value against the type its setting declares, so most of
+//! this cannot fail. Three things keep it from being a formality, and they are the reason the
+//! errors here carry provenance rather than panicking:
+//!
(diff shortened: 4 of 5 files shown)


jdx:
Both real.

**`uint` accepts zero, so no message should say "positive".** The finding is about the reader, but
`Ty::Uint::describe` said the same thing, and mise's `jobs = 0` (auto-detect) is exactly the value
being called invalid — fixing only `read.rs` would have left the crate with two spellings for one
type. It is `a non-negative integer` everywhere now: the type's description, the merge's warnings,
and the read errors, with the assertions that pinned the old wording updated in the same commit.

**A failure could spill across lines.** Same class as the three I fixed in `explain`, and I did not
carry the rule over to the new module: `ReadErrors` joins with newlines, and the two things
interpolated into each line are precisely the ones that can contain one — a value out of a file
(a multi-line string is perfectly ordinary in TOML) and a path. One failure over three lines hides
the failures listed after it. Both now go through `one_line`, which moved to `value.rs` in #857 so
there is one implementation of the rule rather than a copy per renderer.

88 tests. Two mutations, each dropping one `one_line` call, each killing the new test.

jdx:
Carried the empty-value fix from #857 into this module too: `mismatch` quotes the offending value,
and an emptied list is a value a user can perfectly well have arrived at — `MYCLI_EXCLUDE=` clears a
declared default. Quoted as its own text it named nothing:

```
jobs expected a non-negative integer but has ``      # before
jobs expected a non-negative integer but has `[]`    # after
```

Same helper as the explanation and the merge's warnings, so the crate has one spelling for it.
Mutation: quoting the value's own text fails the new test.

PR #863, opened 2026-08-13
fix(xtask): declare a description a comment would change

jdx:
Two more shapes where the shadow's regenerated spec did not say what the source said, both
found by rendering mise's help against usage-lib's.

A command's `help` and `long_help` can be independent: `cmd settings` says "Manage settings"
and then "Show current settings…". A doc comment's long form always *contains* its short one,
because the short form is the comment's first paragraph — so reading one back gave "Manage
settings" where the spec says "Show current settings". Declared instead, as the root's
descriptions already were.

And the long form's opening sentence often ends with a period the short form leaves off. The
generator was throwing that period away to make the two match, so "Task to run." came back as
"Task to run". The comment path is now used only when the long form opens with the short one
*exactly* and then breaks — which is the only arrangement it can reconstruct — and everything
else is declared.

Two tests changed with it. One asserted that a period-only difference "says nothing", which was
the lossy behaviour stated as an intention; it now asserts the pair is declared. The other
covered both shapes at once, and keeps the half that still exercises the comment path: an
indented example inside a long form has to keep its indentation.

The `--help` renderer these were found by is still not at parity and still held back. Each round
of this has moved the first difference further down mise's page — the remaining one is a
disagreement about the column width help wraps to.

---

---

> [!NOTE]
> **Low Risk**
> Changes are confined to xtask shadow generation and checked-in benchmark fixtures; runtime mise behavior is unchanged aside from more accurate generated CLI metadata for comparisons.
> 
> **Overview**
> Tightens the **shadow generator** (`xtask/src/shadow.rs`) so spec help round-trips without losing or mangling text when Rust doc comments cannot represent it faithfully.
> 
> **`needs_declaring`** now considers short and long help together: independent short/long pairs (e.g. `settings`), long-only help, in-paragraph newlines, long forms that continue on the next line (not only after a blank paragraph), and paragraph gaps longer than one blank line. The root CLI **about** uses the same rule so descriptions are not dropped when they fall between “comment-safe” and “declare” cases. **Period-only** differences (e.g. `Task to run` vs `Task to run.`) are emitted as explicit `help`/`long_help` instead of collapsed into a single comment.
> 
> For **clap** shadows, required variadic positionals get **`required = true`** (clap treats `Vec` args as optional otherwise). Help that must be declared is written via **`help`/`long_help`** or **`about`/`long_about`** on flags, args, and subcommands in both dialects.
> 
> **Regenerated** `benches/shadows/mise` and `mise-clap` mirror those rules (more `required = true` on variadic args, long Watchexec/`--security`/`--filter-prog` text moved into attributes, subcommand help on `plugins ls-remote` and `settings`). New/updated unit tests lock in the shapes above.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12c9ea07ffd399951f652f9e27c633d72b6f95e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Two more descriptions the round trip changed

Both found by rendering mise's help through usage-argv and diffing it against usage-lib's, which
is the only test that notices this class at all.

**A command's `help` and `long_help` can be independent.** `cmd settings` says "Manage settings"
and its long form opens "Show current settings…". A doc comment's long form always *contains*
its short one, because the short form is the comment's first paragraph — so the shadow's
regenerated spec said "Manage settings" where mise's says "Show current settings". Declared
explicitly now, as the root's descriptions already were.

**A period was being thrown away.** mise's long forms often end their opening sentence with a
period the short form leaves off, and the generator stripped it so the two would match:
`"Task to run."` came back as `"Task to run"`. The comment path is now used only where the long
form opens with the short one **exactly** and then breaks — the one arrangement it can
reconstruct — and everything else is declared.

## Two tests changed, deliberately

`a_long_form_that_only_adds_a_period_says_nothing` asserted the lossy behaviour *as an
intention*: the two forms say the same thing, so emit one comment. They do say the same thing —
but the spec does not, and a regenerated spec has to. It now asserts the pair is declared.

The other covered two shapes at once; the half that still exercises the comment path is kept,
because an indented example inside a long form must keep its indentation.

## What is still not here

The `--help` renderer. Every round of this work moves the first difference further down mise's
page — line 15 → 351 → 520 → 546 — and the count has stayed at 123 of 211 because the remaining
cause is shared: a disagreement about the column width help wraps to, which shifts where every
entry breaks. That is one more focused fix rather than a long tail, but it is not this PR, and
shipping a public `long_help` that is wrong on more than half a real CLI's pages would be worse
than shipping none.

The fixes above stand on their own: they are round-trip fidelity, which docs, manpages,
completions and the SDK generators all read.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI help output for registry security, Watchexec options, task arguments, remote plugins, and settings.
  * Preserved distinct short and long descriptions, including multiline text, punctuation, custom value names, and required or repeatable arguments.
  * Corrected help text for several commands with required variadic positional arguments.
  * Prevented duplicated, incomplete, or malformed help text in generated command usage.
  * Improved consistency of help information across supported CLI formats.

### DIFF
--- a/benches/shadows/mise-clap/src/lib.rs
+++ b/benches/shadows/mise-clap/src/lib.rs
@@ -305,7 +305,7 @@ pub struct BootstrapDotfilesAddArgs {
     #[arg(long = "yes", short = 'y')]
     pub yes: bool,
     /// Targets to add or update
-    #[arg(value_name = "TARGET", num_args = 0..)]
+    #[arg(value_name = "TARGET", required = true, num_args = 0..)]
     pub target: Vec<String>,
 }
 
@@ -801,7 +801,7 @@ pub struct BootstrapPackagesBrewUntapArgs {
     #[arg(long = "path", short = 'p', value_name = "PATH")]
     pub path: Option<String>,
     /// Tap name(s), e.g. `owner/repo`
-    #[arg(value_name = "TAPS", num_args = 0..)]
+    #[arg(value_name = "TAPS", required = true, num_args = 0..)]
     pub taps: Vec<String>,
 }
 
@@ -937,7 +937,7 @@ pub struct BootstrapPackagesUseArgs {
     #[arg(long = "yes", short = 'y')]
     pub yes: bool,
     /// Packages in `manager:package[@version]` form
-    #[arg(value_name = "PACKAGE", num_args = 0..)]
+    #[arg(value_name = "PACKAGE", required = true, num_args = 0..)]
     pub package: Vec<String>,
 }
 
@@ -1113,7 +1113,7 @@ pub struct BootstrapReposExecArgs {
     #[arg(value_name = "PATH", num_args = 0..)]
     pub path: Vec<String>,
     /// Command and arguments to run in each repo
-    #[arg(value_name = "COMMAND", num_args = 0.., last = true)]
+    #[arg(value_name = "COMMAND", required = true, num_args = 0.., last = true)]
     pub command: Vec<String>,
 }
 
@@ -1734,7 +1734,7 @@ pub struct DotfilesAddArgs {
     #[arg(long = "yes", short = 'y')]
     pub yes: bool,
     /// Targets to add or update
-    #[arg(value_name = "TARGET", num_args = 0..)]
+    #[arg(value_name = "TARGET", required = true, num_args = 0..)]
     pub target: Vec<String>,
 }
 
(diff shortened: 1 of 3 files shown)


jdx:
Right — two checks disagreed and a description could fall between them. A long form that opens with the short one but does not then break (a trailing period is enough) was too much for a comment to reconstruct and not independent enough to be declared, so the program description was dropped altogether.

`needs_declaring` decides both now. The test walks every arrangement the two forms can be in — punctuation added, text added, entirely different, and a clean paragraph break — and asserts a description survives each, rather than asserting the one case that was broken.

jdx:
Confirmed and fixed in 12c9ea0.

The check accepted any run of two or more breaks, but the comment path writes exactly one blank line, so `"Short\n\n\nRest"` came back as `"Short\n\nRest"` — the shadow saying something the spec did not. It is now asked as *exactly* two breaks, which is the same reasoning already applied to the single-break case: either side of the one shape a comment can express is declared verbatim instead.

Nothing in mise's spec has that shape, so the generated shadows are unchanged — the test is what holds it, and it fails against `>= 2`.

PR #864, opened 2026-08-13
feat(config): generate the settings registry from the spec

jdx:
The join. Every CLI in the fleet declares its settings in one file and resolves them in another,
and the two are kept in step by hand — which is why hk declares eighteen `sources.cli` bindings and
reads five, pitchfork generates five settings its own `settings get` cannot reach, and fnox's docs
describe a layer that does not exist. A spec's `config` block is read here, at build time, and what
comes out is `const`. There is no second declaration to forget.

```rust
// build.rs
usage_config_build::generate("hk.usage.kdl").expect("settings");
```

```rust
// src/settings.rs
include!(concat!(env!("OUT_DIR"), "/settings.rs"));
```

What crosses over: types (including `list<>`/`set<>`/`map<>`/`option<>` nesting, with a union or a
name usage does not know becoming `Ty::Any` — the runtime's word for "the spec has said usage
cannot decide what belongs here"), typed defaults including list defaults, merge policy, scope,
named parsers, environment variables in precedence order, custom-source bindings, `hide`,
deprecation notices and renames.

**Ids are consts.** `pub const JOBS: PropId = PropId(4)` — a `PropId` *is* the index into the
table, so reading a setting costs no lookup, and a typo in a key is a compile error instead of a
`None` at run time. The risk of emitting indices is that being wrong is silent (every read answers
about a different setting), so a test walks the whole registry and checks each key is where the
registry says it is.

## What it refuses

A build script is where strictness belongs: the alternative to refusing a declaration that cannot
mean what it says is a warning on every run of a shipped binary, for a mistake only the spec's
author can fix.

| refused | because |
|---|---|
| `renamed_to` naming a setting that is not there | the fold looks the replacement up, so values under the old key are silently never read |
| renames that form a cycle | `lookup` gives up and answers `None`, making both settings *unreachable* rather than wrong |
| an old name with a default of its own | the merge folds a rename to its target, whose default is the one that shows up |
| a `parse` nobody implements | values would be split by a rule that does not exist — one item where the author meant several |
| `map<int, …>` | keys in TOML and JSON are text; the runtime would hand the CLI string keys, quietly |
| two keys generating one const | `task.output` and `task_output` — a compile error in a file the author did not write |

All of them at once, rather than the first: fixing a registry should not be a sequence of builds.

## Verification

The generated registry for [the fixture spec](https://github.com/jdx/usage/blob/agent/config-codegen/config-build/tests/fixtures/hk.usage.kdl)
— one of everything, modelled on hk's own — is **checked in and `include!`d by the tests**, so
`cargo test` compiles it and every assertion runs against the registry that compilation produced. A
generator can be tested by comparing strings, and a string that looks like Rust is not Rust. One
test regenerates and diffs against the checked-in file, so a change to the emitter that is not
regenerated fails; the file is checked in precisely so a human reads that diff.

15 tests. Seven mutations, each disabling one refusal or altering the emitter, each killing the
right test. Clippy clean workspace-wide with `--all-features --all-targets`, `cargo fmt --check`
clean.

Stacked on #862 for `Fold`/`FromValue`, which the end-to-end test reads values through. The typed
`Settings` struct is the next PR — this one stops at the registry, which is the part the runtime
already consumes.

---

> [!NOTE]
> **Medium Risk**
> New build-time codegen path and stricter spec validation; runtime impact is limited to adopters’ build scripts, but mistakes in generated registries would affect every settings read for those CLIs.
> 
> **Overview**
> Adds **`usage-config-build`**, a build-time crate that turns a usage spec’s `config` block into `const` `usage_config::Registry` Rust (including `SETTINGS_PROPS`, `SETTINGS_REGISTRY`, and `prop::*` `PropId` constants).
> 
> **`generate` / `generate_to`** write `settings.rs`, emit `cargo::rerun-if-changed` for the spec and every included file, and skip writes when output is unchanged. **`source` / `watched`** support checked-in golden output and custom build scripts. The emitter maps types, defaults, merge/scope, parsers, envs, bindings, deprecation/renames, and **refuses** invalid specs (bad renames, cycles, unknown parsers, non-string map keys, colliding const names, etc.) with **all** errors at once.
> 
> **`usage-lib`**: `Spec` gains a skipped-serialize **`sources`** list (root + merged includes) so generators watch the right files.
> 
> **Tests**: hk-style fixture, golden `settings.rs`, integration tests that compile and resolve the generated registry, and refusal/golden-drift coverage plus a `gen` example to refresh the golden file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 540f280f287d958a629fb17871e5dd9b0623765e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a configuration build tool that generates strongly typed Rust settings registries from usage specifications.
  * Supports defaults, environment variables, aliases, parsing, merging, scopes, visibility, deprecation, and help text.
  * Generates source in memory or writes it to a configured output location.
  * Tracks included specification files for reliable rebuilds.
  * Provides clear validation and error reporting for invalid configuration definitions.

* **Tests**
  * Added comprehensive integration, generated-output, fixture, and validation coverage for configuration generation.


jdx:
Both real, and one of them is my own check with a hole in it.

**A list default on an old name was generated instead of refused.** The rename check sat *after*
`default_list` returned, so the child-node spelling of a default skipped it entirely — the one shape
of "an alias carries a default nothing reads" that got through. The check now runs before either
form returns, on `default.is_some() || !default_list.is_empty()`.

**An empty key generated `pub const : PropId = …`.** KDL takes `prop ""` perfectly happily, and I
checked: it parses, and the output does not compile — in the *adopter's* crate, in a file they did
not write and cannot fix. Refused now, along with an empty piece of a dotted key (`task..output`),
which generated a field with no name for the same reason.

Two mutations, two dead tests: restoring the `default.is_some()`-only condition fails the new
list-default test, and removing the key check fails the new empty-key one.

jdx:
Real, and a good catch on the shape of it: a key that is nothing but separators — `-` — turned every
character into an underscore and generated `pub const _: PropId = …`. That is *legal* Rust (an
anonymous constant) and unreferenceable, so the failure lands in the generated reader that names it,
and the field it produces (`pub _:`) is not legal at all. I checked both with rustc rather than
reasoning about it.

So the rule is now "every part of a key needs a letter or a digit in it", which subsumes the empty
part I refused last round — one predicate, shared with the struct generator so a key that cannot be a
name does not go on to build a tree and collect a second, stranger complaint.

Mutation: restoring the empty-part-only check fails the test, which now covers `task..output` and `-`.

jdx:
The `é` finding is already fixed — both bots re-anchored it to an older commit. `nameable` now asks
for an ASCII letter or digit, which is the alphabet `ident_of` actually builds names from, and there
is a test for `é` specifically (`67efa02`, reply above). Rust *would* take `É` as an identifier, but
what an arbitrary letter uppercases to is not something to bet a generated name on.

New in this push, from the same class on #865: a key holding a newline ended its own doc comment, and
the rest of it read as code. Keys go through `one_line` where they land in a comment now — a doc
comment is a line, and a key can hold a newline as easily as help text can.

Mutation: dropping the `one_line` fails the new test.

jdx:
This one I had left unaddressed — my sweep filters comments by "newer than my last reply", and replying
to a *later* finding marked this earlier one as handled. It was real.

`rerun-if-changed` covered only the file the build script names, while `parse_file` follows `include` —
and `include` is precisely how a CLI with many settings keeps them in a file of their own, so the file
most likely to be edited was the one not watched. The registry went stale in silence until the wrapper
file happened to change.

The parser now records what it read: `Spec::sources` is the spec's own path followed by each `include`,
recursively (`merge` extends it, so a nested include arrives through its own parse). It is
`#[serde(skip)]` — where a spec came from is not part of what it says, and `usage g json` describes the
latter. `usage-config-build` exposes it as `watched()` and prints every path.

Test: a two-file fixture where the settings live in the included file — `watched()` returns both, and
the generated registry contains the settings only that file declares. Mutation: dropping the `sources`
extend in `merge` fails it.

Also here, from #865's finding about generated names: `ident_of` was ASCII-only while the struct
generator kept any Unicode alphanumeric, so one key produced names from two alphabets. Both are ASCII
now.

jdx:
Two more; one real, one I am declining.

**Two defaults at once — real, and I had asserted the opposite in a comment.** The comment said the
parser refuses `default=1` beside a `default 80 443` node. It does not: the property assigns
`prop.default` and the child node accumulates into `default_list`, and nothing checks that both are
not set. So the emitter returned the list and dropped the scalar, silently. I checked the parser this
time rather than trusting my own comment, and it is refused here now — there is no reading of a
property that declares two defaults.

**`sources` as a private field with an accessor — declining.** Every other field on `Spec` is public:
`name`, `bin`, `cmd`, `config`, `complete`, `examples`. One private field with a getter would make this
type two things at once, and the reason offered — that a caller could modify parser-owned provenance —
applies equally to `config` and `cmd`, which callers *do* build and merge by hand (mise assembles
specs; `merge` is public). A caller who sets `sources` gets exactly what they asked for, which is the
same contract as the rest of the struct. If `Spec` grows an encapsulated core later, that is a change
to the whole type rather than to this field.

Mutation: removing the two-defaults check fails the new test.

PR #865, opened 2026-08-13
feat(config): generate the settings struct a CLI reads

jdx:
The other half of the codegen: the struct a CLI's own code reads. #864 emits the registry
resolution needs; this emits what the code needs, and between them they remove the thing the fleet
writes by hand — pitchfork's ~330 lines of key↔field match arms, written twice (once for `get`,
once for `set`), with five settings still unreachable from its own `settings get`.

```rust
let resolved = usage_config::resolve(SETTINGS_REGISTRY, layers)?;
let settings = Settings::read(&resolved)?;   // typed, and every failure at once
println!("{}", settings.task.output);
```

## The decisions worth reviewing

**Dotted keys are nested structs.** `settings.task.output` is how the setting reads in the file, so
there is no reason for the code to spell it differently. Each group gets its own struct
(`SettingsTask`), named rather than inferred.

**A field is `Option<T>` exactly when the resolution can come back with nothing** — no declared
default, or `option<T>`. So a setting with a default is the value itself, with nothing to unwrap,
and there is no run-time failure for a shape the spec permits. `Fold::required` is still what reads
those fields, and its `Missing` arm is what makes the `expect` in generated code honest: `required`
returns `None` only when it has recorded an error, and `finish` has already turned any error into a
return.

**An old name is not a field.** Every read of a renamed key folds into the setting that replaced it,
so a field for it would be a second name for one value — `config set old.key` writing somewhere
nothing reads is the pitchfork bug in miniature.

**Types**: `bool`, `i64`, `u64`, `f64`, `String`, `PathBuf`, `Vec<T>` (for `list` *and* `set` —
the merge already dropped duplicates, and order is the meaning for a `PATH`), `BTreeMap<String, T>`,
and `usage_config::Value` where the spec declined to say (`object`, a union, a name usage does not
know). `url` and `duration` read as text: what makes a string a URL is what the CLI does with it,
and the crate that owns the duration type owns its spelling — inventing one here would put a
dependency in every adopter's binary for a value some only ever print.

## Refused

- **A key that is both a setting and a group** — `python` beside `python.compile`: one field name
  with two things to be, a value and a table. The spec can say it; no struct can hold it.
- **The four keywords Rust cannot spell** (`self`, `crate`, `super`, `Self`). Every other keyword is
  fine: `type` and `match` are unremarkable names for settings, written `r#type` and `r#match` — for
  the field *and* for the local the reader binds it to, which is what `let match: Option<String>`
  taught me when the golden file stopped compiling.

## Verification

The generated file is checked in and `include!`d, so `cargo test` compiles it: every field asserted
below has to exist with that type or the test target does not build. 19 tests. Four mutations —
allowing the value-and-table collision, allowing an unspellable keyword, dropping raw identifiers,
and letting an alias become a field — each killing the right test.

`SettingsPartial` and `runtime_defaults` are deliberately not here. They need the opposite
direction — a typed value turned back into layer entries — which is its own mechanism and its own
PR, not a struct with `Option`s on it.

---

> [!NOTE]
> **Medium Risk**
> Touches generated API surface and config read paths for every CLI adopter; validation is strong but mistakes in codegen would compile and misread settings at runtime.
> 
> **Overview**
> **Extends usage-config-build** so generated `settings.rs` includes not only the registry and `prop::*` ids but also a nested **`Settings`** type (dotted keys → nested structs), **`Settings::read(&Resolved)`** that reads every field via `fold` and reports all errors at once, and field typing rules (`Option` when there is no default or the spec uses `option<T>`, skip `renamed_to` keys, `r#` for Rust keywords).
> 
> **New `config-build/src/settings.rs`** implements the tree builder, Rust type mapping (`list`/`set` → `Vec`, maps → `BTreeMap`, open types → `usage_config::Value`), and **build-time refusals** for ambiguous shapes (setting vs group under one prefix, colliding field/struct names from `-`/`_` or case, unspellable keywords, `read_*` locals to avoid shadowing `fold`).
> 
> **`usage_config`** gains **`FromValue for Value`** so generated union/`object` fields can read without inventing narrower types.
> 
> Tests and golden output cover the `match` keyword setting, struct reads from files, coercion errors, and the new refusal cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0ddb74af5d0212a2f3eeb9568af76588fa7fec9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config-build/src/emit.rs
+++ b/config-build/src/emit.rs
@@ -27,6 +27,8 @@ pub(crate) fn registry(config: &SpecConfig, name: &str) -> Result<String, Vec<St
         let _ = writeln!(body, "{}", prop_meta(key, prop, &mut problems));
     }
 
+    let settings = crate::settings::settings(&props, &idents, &mut problems);
+
     if !problems.is_empty() {
         return Err(problems);
     }
@@ -70,6 +72,7 @@ pub(crate) fn registry(config: &SpecConfig, name: &str) -> Result<String, Vec<St
         let _ = writeln!(out, "    pub const {ident}: PropId = PropId({index});");
     }
     let _ = writeln!(out, "}}");
+    out.push_str(&settings);
     Ok(out)
 }
 
--- a/config-build/src/lib.rs
+++ b/config-build/src/lib.rs
@@ -35,6 +35,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr as _;
 
 mod emit;
+mod settings;
 
 /// Read `spec` and write the registry to `$OUT_DIR/settings.rs`.
 ///
--- a/config-build/src/settings.rs
+++ b/config-build/src/settings.rs
@@ -1,0 +1,431 @@
+//! The settings struct a CLI actually reads.
+//!
+//! The registry is what resolution needs; this is what the code needs. Between them they are the
+//! whole reason the fleet's hand-written key↔field match arms exist — pitchfork writes about three
+//! hundred and thirty lines of them, twice, and five settings still cannot be reached from its own
+//! `settings get`.
+//!
+//! Dotted keys become nested structs, because `settings.task.output` is how the setting reads in
+//! the file and there is no reason for the code to spell it differently.
+
+use std::collections::BTreeMap;
+use std::fmt::Write as _;
+
+use usage::spec::config::SpecConfigProp;
+use usage::spec::config_type::{Base, SpecConfigType};
+
+/// A group of settings under a common dotted prefix, which becomes one struct.
+#[derive(Default)]
+struct Group<'a> {
+    /// Settings whose key ends here, by their last segment.
+    leaves: BTreeMap<&'a str, Leaf<'a>>,
+    /// Groups below this one, by their segment.
+    groups: BTreeMap<&'a str, Group<'a>>,
+}
+
+struct Leaf<'a> {
+    /// The whole dotted key.
+    key: &'a str,
+    /// The const in `prop`, and the local this is read into.
+    ident: &'a str,
+    prop: &'a SpecConfigProp,
+}
+
+/// The `Settings` struct, its nested structs, and the one function that reads them.
+pub(crate) fn settings(
+    props: &[(&String, &SpecConfigProp)],
+    idents: &[String],
+    problems: &mut Vec<String>,
+) -> String {
+    let mut root = Group::default();
+    for ((key, prop), ident) in props.iter().zip(idents) {
+        // An old name is not a field: every read of it folds to the setting that replaced it, so a
+        // field for it would be a second name for one value — and `config set old.key` writing to a
+        // field nothing reads is exactly the pitchfork bug.
+        if prop.renamed_to.is_some() {
+            continue;
+        }
+        // A key with a piece that cannot be a name is already refused by name, and building a tree
+        // from it invents a group called nothing that then collides with its own parent — two
+        // messages for one mistake, one of them about something the author never wrote.
+        if !crate::emit::nameable(key) {
+            continue;
+        }
+        let segments: Vec<&str> = key.split('.').collect();
+        let (last, path) = segments.split_last().expect("a key has a segment");
+        let mut group = &mut root;
+        for segment in path {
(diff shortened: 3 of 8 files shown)


jdx:
Both right, and both are one root cause: I checked the key *segments* and generated the *names*.

Two segments can differ and still translate to one field — `foo-bar` beside `foo_bar`, since a dash
and an underscore are the same character in an identifier — and two groups can differ and still
translate to one struct, since `camel` uppercases the first letter and `a` beside `A` both become
`SettingsA`. Neither was refused, and neither is caught by the `prop::` const check: `foo-bar.x` and
`foo_bar.y` make two *distinct* consts and one field. The check now runs on the emitted names, in the
two namespaces that exist — field names at each level (leaves and subgroups share it) and struct
names among subgroups.

The value-and-table case keeps its own message rather than being folded into the generic one; "`python`
and `python` both generate the field `python`" is technically true and much less use than "`python` is
a setting and a group of settings".

**A key starting with a digit** was the other half: the const path prefixed `_` and the field path
did not, so `2fa.token` generated `pub const _2FA_TOKEN` beside `pub 2fa:`. Same prefix on both now,
so the two spellings of one key agree, and `_2fa` colliding with a real `_2fa` is caught by the check
above.

Three mutations: dropping the field-collision check, dropping the struct-name check, and dropping the
digit prefix — each killing the right test. 19 tests in the crate.

jdx:
Both real, and the first one is my last fix stopping one level short.

**Struct names are not per level.** I fixed the sibling case and compared siblings — but every struct
is declared in one module and its name is built by *concatenation*, so `http.client.timeout` and
`http_client.retries` both arrive at `SettingsHttpClient` from different depths. The map is threaded
through the whole walk now, keyed on the full generated name, and the message names it:

```
`http.client` and `http_client` are both groups named `SettingsHttpClient`: rename one of them
```

**A local could shadow the reader's own fold.** A setting called `fold` generated
`let fold = fold.optional(…)`, so every read after it — and `fold.finish()` — stopped compiling; a
setting called `SELF` generated `let self`, since the keyword check compares the segment and not the
lowercased ident. Nothing about `fold` is unusual for a config file, so the fix belongs on this side:
every local is prefixed (`read_fold`), which is injective over already-unique consts, so no name a
spec can choose reaches either state. The *fields* keep the names the config file uses.

Also fixed while there: `task..output` produced two problems, the second being a group named after
something the author never wrote — an invalid key is refused by name and no longer builds a tree.

Three mutations: comparing siblings only, dropping the local prefix, and letting a bad key build a
tree — each killing the right test. 25 tests in the crate.

jdx:
Real, and reachable: `a"b` is a nameable key — it has letters in it — and the generated reader quotes
the key into an `expect` message. Interpolated raw, the quote ended that literal early and the rest of
the message read as code, in a file the adopter did not write. The key goes through the same escaper
`PropMeta` has used all along.

Doc comments are the other half and need the opposite treatment — no escapes, one line — so keys and
group names there go through `one_line`. That fix is in #864 as well, where the `prop::` consts carry
their key in a doc comment too.

Mutation: interpolating the key raw into the message fails the new test, which checks all three
places one key lands — the literal, the `PropMeta`, and the comment.

jdx:
Real, and I had missed it in the last sweep — replying to a later finding made my "unaddressed" filter
skip it.

A Rust identifier is not "any alphanumeric character": `½` is one and cannot appear in an identifier,
and a Unicode digit cannot start one. `field` and `camel` kept anything `is_alphanumeric`, while the
`prop::` consts were built from ASCII all along — so one key produced names from two alphabets, a const
the adopter could compile beside a field they could not. Both are ASCII now, which also means the const
and the field derive from the same rule, and a key that collapses to something already taken is caught
by the collision check that was already there.

```
prop "café.si½e"  ->  prop::CAF__SI_E,  settings.caf_.si_e
```

The key itself stays exactly as the spec wrote it in `PropMeta`, because there it is data rather than a
name. The test asserts that, and that every line of generated code which *declares* a name is ASCII.

Two mutations, one per function, each killing it.

PR #866, opened 2026-08-13
feat(argv): render `--help` too, byte-identical to usage-lib's

jdx:
The wider layout: help aligned into a column and wrapped to `COLUMNS`, long descriptions
preferred over short, annotations each on their own line, and the text a command puts around the
rest of the page. All 211 of mise's commands match usage-lib byte for byte, in both forms.

The last cause was the largest: `before_help`/`after_help` and their long forms did not exist in
the metadata at all, and **115 of mise's commands carry their Examples section in
`after_long_help`** — so every one of those pages was missing the part a reader came for. Now
four fields on `CommandMeta`, four attributes on the derive, written into the emitted KDL, and
carried by the generator.

Four smaller ones, each found by reading the reference's output rather than its template:

- No blank line after an entry whose help wrapped. The template asks for one and its whitespace
  trimming eats it before it reaches the page.
- The first line of an indented description is indented even when it is *empty*, and later blank
  lines are not — because the reference writes the indent literally and indents the rest with a
  filter that skips blanks.
- A long form that opens with a blank line does not open with its short form. Trimming before the
  comparison hid that, and `plugins ls-remote` says nothing on its first line.
- A text that ends with a break has a blank line at the end, which `lines()` does not report.

---

---

> [!NOTE]
> **Medium Risk**
> Large help-rendering surface area and widespread shadow regeneration; behavior is heavily tested against usage-lib but user-visible help text changes if parity assumptions were wrong.
> 
> **Overview**
> **Help parity for short and long forms.** The argv renderer now implements `long_help` (column alignment, `COLUMNS` wrapping, long descriptions, per-line annotations) and extends `short_help` with before/after text and example inheritance so output matches usage-lib byte-for-byte across mise’s 211 commands.
> 
> **Metadata and emission.** `CommandMeta` gains `before_help`, `before_long_help`, `after_help`, and `after_long_help` (root-level defaults and per-command overrides); `Spec::EMPTY`, `to_kdl`, and the derive/`gen-shadow` pipeline emit these fields. The mise shadow crate is regenerated with `after_long_help` on many commands so the gate can diff against the real spec.
> 
> **Tests and docs.** New gate tests compare every long help page to the reference plus fixtures for surrounding text, root fallbacks, examples, and KDL round-trip; `PLAN.md` marks help rendering done and notes remaining CLI wiring (`Error::Help`, `help` subcommand).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c14af91b4da875ce0887f94672f29bb97527a75b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Both forms now match, across the whole CLI

**211 of 211 pages, byte for byte, for `-h` and `--help` alike.** The long form adds the column
alignment, wrapping to `COLUMNS`, long descriptions preferred over short, annotations on their own
lines, and the text a command puts around the rest of the page.

## The cause that was worth the wait

`before_help`/`after_help` and their long forms **did not exist in the metadata at all** — and
**115 of mise's 211 commands carry their Examples section in `after_long_help`**:

```
Examples:

    $ mise plugins ls-remote
```

So more than half of mise's help pages were missing the part a reader actually came for. Four
fields on `CommandMeta`, four attributes on the derive, written into the emitted KDL, carried by
the generator.

## Four smaller ones, all found by reading output rather than the template

The template says one thing and its whitespace trimming does another, so the only reliable
reference is what the reference *prints*:

| | what the reference actually does |
|---|---|
| a wrapped entry | is **not** followed by a blank line — the template asks for one and the trimming eats it |
| an indented description | has its **first** line indented even when empty, and later blank lines not — the indent is literal, the rest is a filter that skips blanks |
| a long form opening with a blank line | does **not** open with its short form. Trimming before the comparison hid that, and `plugins ls-remote` says nothing on its first line |
| a text ending with a break | has a blank line at the end, which `lines()` does not report |

## Verification

Four parity tests over mise's real spec — usage line, `-h`, `--help`, and the root line — all
green, and the test reports the first differing line rather than two walls of text, which is what
made 123 → 116 → 1 → 0 tractable.

## Next

The wiring: `--help`/`-h` declared on every command, `Error::Help` carrying the metadata to
render, `-h` short and `--help` long as clap does it, and a real `help` subcommand for any CLI
that has subcommands.

## Summary by CodeRabbit

* **New Features**
  * Added short and long help rendering with wrapped, aligned content and detailed command listings.
  * Added support for examples and customizable text before and after help content.
  * Added fallback to root-level examples when command-specific examples are unavailable.

* **Bug Fixes**
  * Improved preservation of trailing whitespace in long help output.

* **Tests**
  * Expanded coverage for help formatting, inheritance, surrounding text, examples, and reference output parity.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -265,21 +265,14 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
 
 ### Then: what a CLI framework has to have
 
-- [ ] **Help rendering** — `-h` is done and matches usage-lib byte for byte across all 211 of
-      mise's commands: the usage line, the about, the commands list, arguments and flags grouped
-      by heading, examples, hidden-item filtering. `--help`'s wider layout, which aligns help
-      into a column and wraps to `COLUMNS`, is next, then the `--help`/`-h` wiring in the derive.
-      Holding it to parity found six more things a spec could say that the derive could not —
-      value names, required collections, `var` on a count, the spec's own `about`, `hide` on a
-      command, and help text whose line breaks matter — plus a bug in usage-lib, which printed
-      everything marked `hide`. Then three more, found while starting on `--help`: a variant's short
-      description was hiding the struct's long one, which is the shape every generated CLI has;
-      a doc comment's lines were trimmed one by one, flattening every indented example in help;
-      and a program could not describe itself twice over, since a comment's long form always
-      contains its short one where a spec keeps the two independent. `--help`'s own layout is
-      written and not yet at parity — 123 of 211 pages differ, each remaining cause a metadata
-      path where the shadow's description is not the spec's — so it is held back rather than
-      shipped wrong.
+- [x] **Help rendering** — `-h` and `--help` both match usage-lib byte for byte across all 211
+      of mise's commands: the usage line, the descriptions, commands, arguments and flags grouped
+      by heading, examples, the surrounding text, hidden-item filtering, and for the long form the
+      column alignment and wrapping to `COLUMNS`. What is left is the wiring — `--help`/`-h` on
+      every command, `Error::Help`, and a real `help` subcommand. Holding it to parity found ten
+      things a spec could say that the derive could not, plus two bugs in usage-lib's own
+      renderer; the last of them was `after_long_help`, which 115 of mise's commands use to carry
+      their Examples section, so a page without it was missing what a reader came for.
 - [ ] **Completions, self-contained** — `<bin> completion <shell>` emits the
       script; a hidden `<bin> complete-word` serves requests from the binary's own
       embedded spec. Same dispatch shape usage-cli uses today, without requiring
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -217,6 +217,12 @@ fn arg_usage(meta: &ArgMeta<'_>) -> String {
 pub fn short_help(spec: &Spec<'_>, path: &[&str], meta: &CommandMeta<'_>) -> String {
     let mut out = String::new();
 
+    // Text the command puts above everything else, and below it. The short form has only the
+    // one pair; the long form prefers the long variants.
+    if let Some(before) = meta.before_help {
+        let _ = writeln!(out, "{before}\n");
+    }
+
     // The program, then what it is for. usage-lib prints the name when the spec gives one and
     // the binary otherwise, and only when there is a version to put beside it.
     if let Some(version) = spec.version {
@@ -263,6 +269,9 @@ pub fn short_help(spec: &Spec<'_>, path: &[&str], meta: &CommandMeta<'_>) -> Str
         },
     );
     examples_section(&mut out, meta);
+    if let Some(after) = meta.after_help {
+        let _ = writeln!(out, "\n{after}");
+    }
 
     // usage-lib trims the whole document and puts back one newline, which is what keeps the
     // blank lines between sections from becoming trailing ones.
@@ -410,6 +419,10 @@ pub fn long_help(spec: &Spec<'_>, path: &[&str], meta: &CommandMeta<'_>) -> Stri
     let width = terminal_width();
     let mut out = String::new();
 
+    if let Some(before) = meta.before_long_help.or(meta.before_help) {
+        let _ = writeln!(out, "{before}\n");
+    }
+
     if let Some(version) = spec.version {
         let name = if spec.name.is_empty() {
             spec.bin.unwrap_or_default()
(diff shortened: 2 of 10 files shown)


jdx:
Both right, and both invisible to the 211-page comparison — mise carries its Examples in \`after_long_help\` and declares no \`example\` nodes at all, so that code was never reached.

**The preamble** was on \`CommandMeta\`, in the emitted KDL, and documented as text above the usage line, and neither form printed it. The short form was missing \`after_help\` too. Both render them now, the long form preferring the long variants.

**The example ordering** I checked against usage-lib rather than its template, since the two disagree about whitespace often enough that only the output is authoritative — and it prints the description first:

```
Examples:
    the quick way
    $ ex go --fast
```

Now tested against the reference on a hand-built command declaring all of it, since a fixture drawn from one real CLI cannot cover what that CLI never uses. All three mutation-checked.

jdx:
Both right.

**The whitespace-only line**: the reference's filter skips a line with *nothing* on it and still indents one holding spaces, so `trim().is_empty()` was dropping those spaces. Worth noting my first attempt at a test for it proved nothing — I put the line in a command's own long help, which its own page never renders, since a command's description appears in its *parent's* list. Moved to a flag's long help, where the block layout reads it, and the mutation fails now.

**The skipped commands**: fixed, and the asymmetry with the short-form test is the tell — a command in the shadow that the spec does not have is a difference between the two, and passing silently on it would let an extra or misnamed one through.

jdx:
Both of these were worth the check — one is fixed, one does not reproduce.

**Root help text skipped on subcommands** — real, and in a sharper form than described. usage-lib reads top-level `before_help`/`after_help` as the default for *every* page, not just the root's:

```
=== go   (declares its own before_help) : "CMD-PRE\n\nUsage: ex go\n\nSPEC-POST\n"
=== other (declares nothing)            : "SPEC-PRE\n\nUsage: ex other\n\nSPEC-POST\n"
=== other, long form                    : "SPEC-PRE-LONG\n\nUsage: ex other\n\nSPEC-POST\n"
```

A root has nowhere else to put its text — `to_kdl` writes it at the top level — so emitting it only on `ROOT_META` meant a derived CLI showed its preamble on the root page and nowhere else, while the same CLI rendered from its own emitted KDL showed it everywhere. Fixed in fcc6823: the derive emits it at both levels, and the test asserts our subcommand page equals usage-lib's page for the same command parsed back from `to_kdl`, so the two descriptions of one CLI cannot drift apart again.

**Extra blank line after trailing breaks** — does not reproduce. Compared against usage-lib for a flag long help ending in one, two and three breaks, and one with a trailing break mid-text, in both forms:

```
=== one   long=false SAME=true    === one   long=true SAME=true
=== two   long=false SAME=true    === two   long=true SAME=true
=== three long=false SAME=true    === three long=true SAME=true
=== mid   long=false SAME=true    === mid   long=true SAME=true
```

The extra newline `write_indented` appends is what makes ours match: the reference's template ends an entry the same way, and the 211-page comparison over mise's spec is what holds that.

jdx:
Both of these are the same seam, and cc2b71e closes it by removing the seam rather than patching each side.

**Root help override is discarded** — correct. The renderer preferred `meta`, `to_kdl` preferred `spec`, so a root override survived rendering and was lost on the way out. The cause was having two homes for one declaration: KDL has exactly one place for this (the top level), and the reference reads what is written there as the root's *and* as the default for every other page. The four `Spec` fields are gone; the root's metadata is the only home, the renderer falls back to `spec.root`, and `to_kdl` writes `spec.root`. The mismatch is no longer expressible.

**Spec examples skipped on subcommands** — correct, and the same rule was simply missing for examples:

```
=== go   (declares none of its own) → shows "the quick way / $ ex go --fast"  (the spec's)
=== deep (declares its own)         → shows only "$ ex go deep --own"
```

`page_examples` now applies it, and the test compares both cases against usage-lib — a page that borrows the root's, and one that keeps its own and does not also show the root's. It fails if the fallback is removed.

jdx:
Checked, and this one does not hold: usage-lib accepts **both** spellings, and `before_long_help` is the one it writes itself.

- [`lib/src/spec/mod.rs:218`](https://github.com/jdx/usage/blob/main/lib/src/spec/mod.rs#L218) — `"before_long_help" | "before_help_long" =>`, and the same for `after_*` on line 221.
- [`lib/src/spec/cmd.rs:250-258`](https://github.com/jdx/usage/blob/main/lib/src/spec/cmd.rs#L250) — both spellings again, at the command level.
- [`lib/src/spec/mod.rs:496`](https://github.com/jdx/usage/blob/main/lib/src/spec/mod.rs#L496) — usage-lib's own `to_kdl` emits `KdlNode::new("before_long_help")`, so this crate is writing what the reference writes.

`the_root_writes_its_own_surrounding_text` already holds it from the other side: it writes the KDL, parses it back with usage-lib, and asserts `parsed.after_help_long == Some("Below, at length.")` — which is only true if the node survived the round trip.

PR #867, opened 2026-08-13
feat(config): read the environment as a layer

jdx:
The layer every CLI in the fleet has, and the one every CLI writes again — mise reads 33 of its
settings through `parse_env` functions, hk's generator emits a match arm per variable, fnox has ~48
`FNOX_*` variables that live outside its registry entirely. All of them do the same thing: look up
the names a setting declares and take the first one that is set. The registry already knows those
names, so nobody should be writing this again.

```rust
let env = EnvLayer::from_process();
let resolved = resolve(SETTINGS_REGISTRY, Layers::new().then(&files).then(&env))?;
```

**The declared order decides, and the loser is not read.** A setting's variables are listed highest
first (`MISE_JOBS` beside an older `MISE_JOB`), and stopping at the first one set is what makes an
alias a second *name* for one setting rather than a second setting.

**The origin is the variable the user set.** An explanation that said "set by the environment" would
send them looking through every name the setting has.

**A bad value costs its own variable.** Refusing to start because one line of a shell profile has a
typo in it would be worse than the typo, so it is a warning naming that variable and everything else
resolves.

**The environment is injected, not reached for** — tak's pattern. A test describes an environment
without touching the process, and two tests can describe different ones at once; `from_process` is
the single place `std::env` is read. This also let the crate's headline doc example shrink, since it
was hand-writing this layer to have something to show.

**Names are case-insensitive on Windows only.** `std::env::var("PATH")` finds `Path` there, so a
case-sensitive lookup would miss a variable the user has plainly set — on one platform, which is the
worst place for that kind of difference to live.

## Verification

97 tests in the crate. Three mutations, each killing the right test: not stopping at the first name
set (both aliases contribute), swallowing a type error instead of warning, and labelling the origin
"the environment" instead of the variable. The Windows case test is `cfg(windows)`.

Stacked on #865. `scope="env"` settings — the ones mise treats as a security property, never
readable from a file — are exactly what this layer is for, and the merge already enforces that side.

---

> [!NOTE]
> **Medium Risk**
> Touches core config resolution precedence and rename/union edge cases; behavior is heavily tested but wrong env winner logic could change effective settings for upgraded users.
> 
> **Overview**
> Adds **`EnvLayer`**, a registry-driven `Layer` that maps declared `envs` to settings instead of each CLI reimplementing env lookup.
> 
> **Injection vs process reads:** `EnvLayer::new` / `from_process` (sole `std::env` use) so tests can supply env without touching the process; non-UTF-8 names/values are skipped instead of panicking like `std::env::vars`.
> 
> **Resolution rules in `load`:** first declared env name wins for aliases; origins use the variable name the user set (with Windows case-insensitive matching); parse failures warn per-variable without blocking other settings; renamed/deprecated vars fold to the target with explicit winner selection (canonical beats old names; stable order among old names) so `union` merge does not double-apply competing env vars.
> 
> **Public API:** `pub mod env` and re-export `EnvLayer`; crate doc example replaces the hand-written stub env layer with `EnvLayer::new`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7093aff76906d23fd149a84c8cb66847998bd15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config/src/env.rs
+++ b/config/src/env.rs
@@ -1,0 +1,506 @@
+//! The environment as a layer.
+//!
+//! The one layer every CLI in the fleet has, and the one every CLI writes again: mise reads 33 of
+//! its settings through `parse_env` functions, hk's generator emits a match arm per variable, fnox
+//! has about 48 `FNOX_*` variables that live outside its registry entirely. All of them are doing
+//! the same thing — look up the names a setting declares, take the first one that is set — and the
+//! registry already knows those names.
+//!
+//! The environment is *injected* rather than read at the point of use, so a test does not have to
+//! touch the process to describe one, and two tests can describe different ones at the same time.
+//! `EnvLayer::from_process` is the one place `std::env` is read.
+
+use std::collections::BTreeMap;
+use std::ffi::OsStr;
+
+use crate::layer::{Layer, LayerCtx, LayerError, LayerOutput, Warning};
+use crate::registry::PropId;
+use crate::source::{Origin, SourceKind};
+
+/// Settings read from environment variables.
+pub struct EnvLayer {
+    /// Keyed by the comparable form of each name, holding the name as it was actually set and its
+    /// value. Both are kept because the comparison and the report want different things: one needs
+    /// the names to match, the other needs to print what the user typed.
+    vars: BTreeMap<String, (String, String)>,
+}
+
+impl EnvLayer {
+    /// The variables this process was started with.
+    ///
+    /// `std::env::vars` *panics* on a name or value that is not UTF-8, which would take the CLI down
+    /// before it resolved anything — over a variable that in all likelihood belongs to something
+    /// else entirely. Read as OS strings and skipped when they will not convert: a setting whose
+    /// variable holds bytes this cannot read has no value it could have been given anyway.
+    pub fn from_process() -> Self {
+        Self::new(std::env::vars_os().filter_map(|(name, value)| readable(&name, &value)))
+    }
+
+    /// A named environment, for a test or for a CLI that has its own idea of one.
+    pub fn new(vars: impl IntoIterator<Item = (String, String)>) -> Self {
+        Self {
+            vars: vars
+                .into_iter()
+                .map(|(name, value)| (normalize(&name), (name, value)))
+                .collect(),
+        }
+    }
+
+    /// What this layer would read for `name`, if anything.
+    pub fn get(&self, name: &str) -> Option<&str> {
+        self.vars
+            .get(&normalize(name))
+            .map(|(_, value)| value.as_str())
+    }
+
+    /// The name as it is really set, which is what a report should print.
+    fn set_as(&self, name: &str) -> Option<&str> {
+        self.vars
+            .get(&normalize(name))
+            .map(|(set_as, _)| set_as.as_str())
+    }
+}
+
+/// One variable, if it is text at all.
+///
+/// A name or value that is not UTF-8 is skipped. A setting whose variable holds bytes that cannot be
+/// read has no value it could have been given, and the alternative is what `std::env::vars` does:
+/// panic, taking the CLI down before it resolves anything, over a variable that in all likelihood
+/// belongs to something else entirely.
+fn readable(name: &OsStr, value: &OsStr) -> Option<(String, String)> {
+    Some((name.to_str()?.to_string(), value.to_str()?.to_string()))
+}
+
+/// A variable's name in the form this layer compares.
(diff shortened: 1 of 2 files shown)


jdx:
Three findings, all real, and the third one is the interesting one.

**An old name could beat the name that replaced it.** Both entries fold to the same setting and the
merge takes the last writer, so with `HK_CONCURRENCY` and `HK_JOBS` both set — which is exactly what
living through a rename looks like — which one won came down to *which key sorted later in the
registry*. Aliases are now loaded first, so a setting's own variable overwrites them, and the
deprecation is still reported, because the user should stop setting the old one. The first-wins rule
inside a setting's own `envs` list was never the same question.

**`std::env::vars` panics on a name or value that is not UTF-8**, which would take a CLI down before
it resolved anything, over a variable that in all likelihood belongs to something else. It reads
`vars_os` now and skips what will not convert — a setting whose variable holds unreadable bytes has
no value it could have been given anyway. The skip is a named function with its own test (a
non-UTF-8 `OsStr`, `cfg(unix)` since that is where one can be built portably), and reverting
`from_process` to `vars()` now *fails the build*: `readable` becomes unused, which the repo's
`-D warnings` gate refuses. I checked that.

**The Windows spelling.** Fixed — the layer keeps the name as it was actually set beside the
comparable form, and the origin reports that. Worth being precise about the test: off Windows a match
requires exact equality, so the declared name and the user's are the same string by construction and
there is nothing an assertion could distinguish. The assertion therefore lives in the `cfg(windows)`
test, next to the case-insensitive lookup it belongs to, and a mutation of it is unobservable on
Linux — which is why I am telling you rather than claiming a mutation proof I did not get.

100 tests in the crate. Mutation: loading aliases last fails the new rename-precedence test.

jdx:
Right — the partition only settled aliases-versus-current, and among *two* aliases for one setting the
winner was still whichever key sorted later.

Nothing declares which old name should win when both are set, so the honest fix is to make the choice
deterministic and *visible* rather than pretend there is a right answer: the first in registry order
wins, and the other is reported.

```
HK_THREADS was not read: HK_CONCURRENCY also sets jobs (HK_THREADS)
```

The current name still beats both, which is the case from the last round. The fixture now has two old
names for `jobs`, since one alias could not exercise this at all.

Two mutations: letting the second alias overwrite the first, and making the warning say nothing
useful — both killing the new test. 102 tests in the crate.

One process note: the ref for this branch updated but GitHub kept the PR head on the previous commit
for several minutes and did not pick it up from a re-push, so I re-pushed with a fresh sha (identical
content) to force the synchronize. If you see a duplicate CI run on this PR, that is why.

jdx:
Right, and it is this layer's own oldest rule broken by its newest one: a bad value costs its own
variable and nothing else. Recorded on *presence*, `HK_CONCURRENCY=lots` discarded a perfectly good
`HK_THREADS=9` and the warning claimed the failed variable had set the setting.

The bookkeeping now happens after `ctx.entry` succeeds, so only a value that actually read speaks for
its setting. A typo in one old name leaves the next one to be read normally, and there is no "was not
read" warning, because nothing was passed over.

Mutation: moving the bookkeeping back before the read fails four tests, including the two from the
last two rounds.

jdx:
Right, and it goes deeper than the previous two rounds: my fix relied on the merge taking the last
writer, which is only true for `replace`. For `union` the deprecated variable's items ended up in the
list **as well** as the current one's, and an emptied old variable cleared the declared default on its
way past.

So the choice is made here rather than left to the merge. Candidates are gathered by the setting they
fold into, and exactly one entry is pushed per setting: the setting's own name first, then old names in
registry order, and the first that *reads* wins — so a typo in one name still does not consume the
setting. Everything else is reported:

```
HK_SKIP_STEPS was not read: HK_SKIP also sets skip (HK_SKIP_STEPS)
```

This also deleted the aliases-first partition from the last round, which was a workaround for the
ordering problem this replaces.

One behavior change worth flagging, since it is a judgement rather than a fix: an old name that is not
read no longer produces its deprecation notice, because nothing was deprecated *into use*. The message
above is the actionable one — this variable does nothing — and if the user removes `HK_SKIP` believing
the old name still works, the deprecation arrives then, which is when it tells them something.

Three mutations: pushing every candidate, preferring the alias, and letting a read value speak for
nothing — each killing three tests, including a new one for the `union` case, which one alias and one
`replace` setting could not exercise.

PR #868, opened 2026-08-13
feat(config): hold a value to the choices its setting declares

jdx:
A spec's `choice` nodes reach the docs, the JSON schema and completions — and, until this, nothing
that *resolved* a value. `PropMeta` had no choices at all, so a CLI documenting three allowed values
took a fourth in silence and failed later, somewhere that could not say why. That is the declaration
and the behaviour drifting apart, which is the thing this crate exists to stop.

```
$ HK_STASH=svn hk check
stash expected one of git, patch-file, none but has `svn` (HK_STASH)
```

**Where the check goes.** Beside the type check, in both paths a value can arrive by — text through a
parser, and a structured value out of a file. Checking one and not the other is how a rule ends up
applying to the environment and not to the file beside it, so there is a test for each.

**Collections are checked item by item.** Choices on a `list<string>` mean each item is one of them —
the rule `usage g json-schema` already follows, putting the enum on every value position rather than
on the container — and the **item** is what the message quotes, since naming the whole list leaves
the user to work out which of five is wrong.

**A refused value costs its own key**, exactly like a value of the wrong type: a bad line in a
system-wide file must not stop a CLI from starting for every user on the machine.

`Const::matches` compares without building the `Value` a choice stands for: this runs once per
declared choice for every value supplied, and a setting with choices is usually a string, where the
comparison would otherwise allocate a copy to throw away. A setting with no choices returns
immediately.

`explain` now lists what a setting will take — somebody reading an explanation after a refusal has
already scrolled past the warning.

## Verification

107 tests in the crate. Four mutations, each killing the right test: the text path unchecked, the
structured path unchecked, list items unchecked, and an empty choice list refusing everything (which
takes 30 tests with it, since it makes every ordinary setting unsettable).

Next in the stack: the generator emitting these from the spec, and refusing a declared default that
is not among them — an authoring mistake that would otherwise be a warning on every run of a shipped
binary.

### DIFF
--- a/config/src/explain.rs
+++ b/config/src/explain.rs
@@ -66,6 +66,11 @@ pub fn explain(resolved: &Resolved, key: &str) -> Option<String> {
     // The spec's own spelling, not the prose an error message uses: a reader searching the docs
     // for "a non-negative integer" finds nothing, and `uint` is what the author wrote.
     let _ = writeln!(out, "  {:<8}{}", "type", meta.ty.name());
+    // What it will take, when it says. A user reading an explanation because their value was refused
+    // needs the list here rather than in a warning they have already scrolled past.
+    if !meta.choices.is_empty() {
+        let _ = writeln!(out, "  {:<8}{}", "one of", one_line(&meta.allowed()));
+    }
     if let Some(help) = meta.help {
         // Through the same helper as everything else: an adopter's help is a doc comment, and a
         // doc comment with a second paragraph in it split one fact into several records.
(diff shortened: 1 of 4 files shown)


jdx:
Both bots found the same gap, and it is real — but it is the next commit rather than a defect here:
#869 sits directly on top of this branch and emits `choices` from the spec's `choice` nodes, with an
end-to-end test that runs a config file through the *generated* registry and watches the merge refuse
a value the spec does not allow.

They are stacked, so this never reaches a release without it. Split because they are two different
things to get wrong: what the rule is (here — where it is checked, what a collection means, what a
refusal costs) and whether the declaration reaches the registry (there). A hand-written registry —
the shape every test in this crate uses, and what a CLI writes before it adopts the generator — is
enforced by this commit alone.

PR #869, opened 2026-08-13
feat(config): carry a setting's choices into the generated registry

jdx:
The step between #868's enforcement and the spec that declares it: `choice` nodes become `Const`s in
the generated registry, so the values a CLI documents are the values it accepts. The end-to-end test
runs a config file through the *generated* registry and watches the merge refuse `stash = "svn"`
while the declared default stands — the three places that used to disagree (docs, schema, resolution)
now come from one declaration.

**A default the choices do not allow is refused here.** At run time it is seeded as the bottom layer
and then goes through the same check as every other value, so it would be a warning on every run of a
shipped binary — for a mistake only the author of the spec can fix:

```
`stash` defaults to `svn`, which is not one of the values it allows: git, none
```

A list default is held to them item by item, the way the values themselves are.

Two mutations: not emitting the choices (caught by the regenerate-and-diff test, which is how any
emitter change is caught here), and checking only the scalar default and not the list one.

Still deliberately out: generating an *enum* per choice-bearing setting. It needs variant naming,
a `FromValue` impl and its own error, and none of that is needed for the values to be enforced —
`String` plus a refusal at the boundary is the smaller thing that makes the declaration true.

---

> [!NOTE]
> **Medium Risk**
> Touches config resolution, codegen defaults, and value display/coercion across three crates; mistakes could change which values are accepted or what defaults ship, though coverage is heavy.
> 
> **Overview**
> **Choices in the generated registry** — `config-build` now emits `choices` on each `PropMeta` and depends on `usage-config` so validation uses the same `Ty::coerce` the runtime uses.
> 
> **Build-time spec validation** — The generator rejects defaults and choices the declared type cannot read, defaults outside the allowed set (including list defaults checked per-item or as a whole for scalar/`any` types), and emits defaults via `coerced_const` so seeded values match what resolution expects (e.g. `default "yes"` on `list<bool>` becomes bool literals).
> 
> **Runtime choice matching** — `PropMeta::refuses` walks collections using the **declared** type (not value shape), and `allows` coerces each choice before comparing. `Const::matches` falls back to matching `display()` text so cross-shaped literals (e.g. `choice 4` vs string `"4"`, `choice "yes"` vs bool) agree with coercion.
> 
> **Shared float rendering** — `usage-config` and `usage-lib` format whole-number floats with a decimal point (`1.0` vs `1`) so default/choice comparisons and error messages stay consistent across spec parse, codegen, and merge warnings.
> 
> **Tests** — Fixture `log_format`, golden output, end-to-end resolve for invalid `stash`, and broad refusal tests in `config-build/tests/refusals.rs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 964d72b619d41a9577c158e2b23eb3e349fbe8a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Configuration settings now expose their permitted values.
  * Added the `log_format` setting with supported values `1` and `2`.
  * Compatible scalar values are recognized across textual and numeric representations.

* **Bug Fixes**
  * Invalid scalar and list defaults are rejected with clear messages showing invalid and allowed values.
  * Invalid setting values fall back to the effective default and generate a warning.
  * Configuration validation messages now format values more clearly, including whole-number decimals.

### DIFF
--- a/config-build/Cargo.toml
+++ b/config-build/Cargo.toml
@@ -14,6 +14,10 @@ license = { workspace = true }
 # parser lives here and `usage-config` carries none of it.
 [dependencies]
 usage-lib = { workspace = true }
+# For the *rules*, not for the generated code: a default has to be held to the same choices, and read
+# by the same coercion, that the runtime will apply — and a second implementation of those rules here
+# is exactly the drift this crate removes elsewhere. It carries no dependencies of its own.
+usage-config = { workspace = true }
 
 # The generated code is compiled and run in the tests, which is the only way to know it is code
 # rather than plausible text.
--- a/config-build/src/emit.rs
+++ b/config-build/src/emit.rs
@@ -136,6 +136,10 @@ fn prop_meta(key: &str, prop: &SpecConfigProp, problems: &mut Vec<String>) -> St
         }
         let _ = writeln!(fields, "        bindings: &[{}],", pairs.join(", "));
     }
+    if !prop.choices.is_empty() {
+        let values: Vec<String> = prop.choices.iter().map(|c| value_const(&c.value)).collect();
+        let _ = writeln!(fields, "        choices: &[{}],", values.join(", "));
+    }
     if prop.hide {
         let _ = writeln!(fields, "        hide: true,");
     }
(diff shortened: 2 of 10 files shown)


jdx:
Real, and it had a twin in the generator I would not have found without it.

A spec writes `choice 4` under `type="string"` as readily as `choice "4"`, and by the time a value
reaches the check it has been coerced to the *declared* type — so the choice was an integer, the value
was the string `4`, and comparing shapes refused a value the spec plainly allows. `Const::matches` now
falls back to comparing what the two are *written* as, which is the same question the coercion already
answered: `Ty::String` turns `4` into `"4"`, and `Ty::Float` turns `1` into `1.0`, whose text is `1`
either way. Only scalars reach that comparison — `refuses` walks a list or a table item by item first
— so `a,b` the string cannot be mistaken for `[a, b]` the list.

The twin: this crate's own build-time check compared `SpecConfigValue`s by equality, so a spec with
`default="1"` beside `choice 1` was **refused outright**. I found it by adding the fixture for your
finding and watching the generator reject it. Same rule, same reason.

The fixture now has a `string` setting whose choices are written as numbers, so the golden file
carries `Const::Int` choices under a `Ty::String` prop, and the end-to-end test reads `log_format = 2`
out of a TOML file and expects no warning at all.

Two mutations: strict shapes at run time (kills the unit test and the end-to-end one) and strict
shapes at build time (kills the golden diff, since the fixture no longer generates).

jdx:
Right again, and this one took the fix past text comparison altogether.

`choice "yes"` under `type="bool"`: the value `yes` is coerced to `Bool(true)` on its way in while the
choice stays the string `yes`, so comparing what they are *written* as refused it. The choice is now
read the same way the value was — `Ty::coerce` applied to the choice — which is the same question the
coercion already answered, rather than a second rule that approximates it. Text comparison survives as
the fallback for a choice the declared type cannot read, which is what `any` needs, since nothing is
coerced there at all.

Inside a collection it is the **item's** type that reads the choice, not the container's: read as the
list it is in, `yes` becomes a one-item list and matches nothing. There is a `list<bool>` case in the
tests for that — unusual, and the only shape where the item type is the only thing making the
comparison right.

**The generator had the same weakness**, and refused that spec outright: its default-vs-choices check
compared spec literals, so `default=#true` beside `choice "yes"` was a build failure. It holds a spec
type rather than a runtime one and cannot do the reading, so it now refuses only what it can be sure
of — every choice written like the default, and none equal to it, is a typo — and leaves a difference
it cannot interpret to the coercion. I found this by probing the case from your finding through the
generator rather than assuming the two agreed.

Three mutations: the choice not read as the type, the element type replaced by `any`, and the
build-time check refusing mixed shapes. Also worth noting from writing those: the type check runs
*before* the choices, so a value that is not a boolean at all is reported as not a boolean — there is
nothing useful to say about which choice it is not.

jdx:
Correct, and it is a comment claiming what the code does not do — my least favourite kind, and the
third one this stack has caught. I wrote that the `Err` arm is "what `any` needs, since nothing is
coerced there at all", and `Ty::Any::coerce` succeeds by doing nothing, so `any` never took that arm:
a union's `choice 4` was compared, as an integer, against the string `4` a file supplied, and refused.

`allows` now tries the coerced comparison and falls back to the written form whenever that does not
settle it, rather than only when coercion errors. The test is a `Ty::Any` setting with integer choices
read from text, and the mutation — returning early on any successful coercion — kills it.

jdx:
This one corrects something I asserted rather than checked: I said a cross-shape default was "left to
the coercion at run time", and **a seeded default never goes through that check at all** — `resolve`
writes `PropMeta.default` straight into the resolution. So `default="3"` beside `choice 1` would have
become the effective value with nothing said. My narrowing created the hole; you are right.

The fix is the one I talked myself out of two rounds ago: `usage-config-build` now depends on
`usage-config` for the *rules* — not for the generated code — and reads defaults and choices through
the same `Ty::coerce` the runtime applies. That is possible because every **base** type is
`const`-constructible without a reference (only the composites borrow), and defaults and choices are
scalars, so the element base is all it needs. A second implementation of coercion here is exactly the
drift this crate exists to remove, which is why the discriminant comparison was the wrong shape of
answer.

Chasing your mutation also turned up the same asymmetry I had just fixed in the runtime, now in the
generator: a union coerces nothing, so `default="1"` beside `choice 1` was refused at the equality arm
and never reached the written-form fallback. Both now try the coerced comparison and fall back only
when it does not settle the question.

Three mutations: coerced values not compared, no written-form fallback (a union), and the list element
type replaced by `any` (a `list<bool>` with `default "yes"`, the one shape where the item's own type is
what decides).

jdx:
Right on both halves, and probing it turned up a third. `Ty::Bool` reads booleans and their spellings
and not numbers, so `choice 1` under `type="bool"` is a choice nobody can ever pick — every value the
setting is given would be refused. I checked what the three shapes actually did rather than reasoning
about them:

| declared | before | now |
|---|---|---|
| `type="bool"` with `choice 1` | refused *by accident*, blaming the default: "`c` defaults to `true`, which is not one of the values it allows: 1, 0" | `` `c` has a choice `1`, which is not a boolean `` |
| `type="bool"` with `default=1` | **accepted** — seeded straight into the resolution, read by nothing | `` `d` defaults to `1`, which is not a boolean `` |
| `type="uint"` with `choice "lots"` | **accepted** — an unreachable choice, in the docs as a value one may set | `` `e` has a choice `lots`, which is not a non-negative integer `` |

So the rule is one thing rather than three: **anything declared here that the declared type cannot
read is a value nothing can ever hold.** A default is the case nothing else catches at all — it is
seeded with no coercion and then read by whatever holds it — which is the same gap you found last
round from the other side.

The default-vs-choices comparison is now skipped when any choice is unreadable: against a choice
nothing can read, "not one of the values it allows" is a second message for one mistake, and it blames
the wrong end of it. That is what the first row above was.

Three mutations: unreadable values accepted, choices excluded from the check, and the cascade allowed
to fire.

jdx:
Real, and it is my own test case from the last round biting: I accepted `default "yes"` under
`list<bool>` because the item type reads it as `true`, and then emitted `Const::Str("yes")`. A default
is seeded straight into the resolution with no coercion, and `FromValue` is strict, so the first
`Settings::read` fails on a registry this crate had just accepted.

Defaults are emitted **coerced** now — as the value the type says they are. The fixture carries the
natural version of this (`default=1` under `type="string"`), the golden shows `Const::Str("1")`, and
the compiled test reads it back as `"1"`.

Choices are deliberately *not* put through it: they are documentation as much as values — a bool with
`choice "yes"`/`"no"` should say "one of yes, no", not "one of true, false" — and the comparison
coerces them at the moment it needs them.

One honest note on the mutation testing. The scalar path is not observably fixed by this: the spec's
own parser already reads a scalar default as the declared type, so `value_const` and `coerced_const`
agree there, and a mutation of it kills nothing. Only the list form is genuinely fixed. I applied it to
both anyway — which of them the parser happens to cover is not something this should depend on — and
said so in the doc comment rather than claiming a coverage I do not have. The list mutation does kill
its test.

jdx:
Real, and the fix belongs a level down from the comparison: a whole-number float rendered as `1`,
which is how an **integer** is written. So `choice 1.0` under a string type accepted `1` and refused
the `1.0` the spec had written, and the list of allowed values said `1` back to an author who never
typed that.

`Value::display` now writes a float with its point — `1.0`, `0.5`, and untouched for anything already
carrying a `.`/exponent or not finite. That makes the comparison agree with the spelling in the spec
and the message agree with both, rather than teaching the comparison to paper over the rendering.

One consequence worth stating plainly, since it reaches past this PR: `config get` on a whole-number
float now prints `1.0` where it printed `1`. I think that is the better answer — a value that is a
float and prints as an integer is the same misrepresentation `shown()` fixed for empty lists — but it
is a change in output, not only in validation.

It also cost a line of my own test: `Const::Int(1).matches(&Value::Float(1.0))` was true through the
old rendering and is false now, which is right. The pair that actually matters — `type="float"` with
`choice 1` — never needed it: the coercion reads the choice as a float first, and there is now a test
for that beside the string-typed one.

Mutation: dropping the point fails both the rendering test and the float-choice one.

jdx:
Right, and it is the direct consequence of the last fix: I taught the runtime that a whole-number float
is `1.0` and left this crate with its own copy of the rendering, so the two disagreed immediately —
`default="1"` beside `choice 1.0` accepted here and refused there, on a value that is then *seeded* and
never checked again.

The duplicate is gone: `display` here is `runtime_value(value).display()`, so there is one renderer for
both, the same way there is now one coercion for both. That is the general shape of this whole review
thread — every one of these was two implementations of one rule, and the fix each time was to delete
one of them rather than to teach them to agree.

Mutation: giving this crate its own renderer back fails the new test.

jdx:
Confirmed by running it rather than reading it: `type="string" default=1.0` beside `choice 1.0` — the
*same literal on both sides* — was refused, because the default is coerced to text by the **spec's**
renderer and the choice is read by the **runtime's**, and I had only taught one of them that a
whole-number float keeps its point.

usage-lib writes a float the same way now. It cannot depend on usage-config, so the rule genuinely
exists in two crates — and `usage-config-build` is the one place that can see both, so there is a test
there holding them together over a spread of values (`0.0`, `±1.0`, `0.5`, `1e3`, `1e300`, `f64::MIN`,
`f64::MAX`). Reverting either side fails it.

That is three renderings reconciled in this thread (`Value::display`, this crate's copy, and now the
spec's) and one coercion shared. The pattern in every one of them was the same: two implementations of
one rule, fixed by deleting or joining them rather than teaching them to agree.

jdx:
Real, both halves. The walk followed the *value's* shape rather than the declared type, so a setting
whose type declares no items at all — `any`, a union — accepted `[1]` for `choice 1`: it went looking
for items in something nothing said had any, and for an open type a value's shape is whatever the file
happened to write. It switches on the declared type now, and anything that is not a declared collection
is compared whole, where `Const::matches` already says a scalar choice is not a list.

The generator had the same rule and the same gap: `default 1 2` beside `choice 1` and `choice 2` under
a union passed item by item, and then the whole list is what a seeded default *is* — refused by its own
choices at run time and never checked again. Item-wise validation now applies only where the type
declares items; elsewhere the list is one value and reported as one.

Two mutations: the recursion driven by shape again, and the list-default check disabled.

jdx:
Right, and it is the same lesson one level further in: I asked the *spec's* type whether it declares
items, and `simplified` collapses a union to its first arm — so `list<uint>|uint` looked like a list,
while what this crate emits for any union is `Ty::Any`, which compares a value whole. The question is
now asked of the type that is actually emitted, and `option<T>` is looked through for the same reason
the runtime looks through it.

Two mutations: `simplified` restored (the union case), and `option` not looked through (an
`option<list<uint>>` with a list default, which should be checked item by item). Both kill the test,
which now covers `list`, `set`, `option<list>` and the union.

PR #870, opened 2026-08-13
feat(argv): answer `--help` and `-h`

jdx:
The parser recognises both itself rather than a CLI declaring them, which is the only
arrangement that does not change what help *says*: no spec declares `--help`, so a table entry
for one would put it in the rendered page and in the emitted KDL, and the page would then
disagree with the spec it came from.

Recognised *after* the command's own flags, so a CLI that declares its own `--help` or `-h`
keeps it and binds it like any other flag.

A request comes back as `Error::Help`, carrying the command it was asked about — `mise config
--help` is a question about `config`, and the parser is what knows how far the words reached.
Not printed: a library that writes to stdout on its own is one an adopter cannot embed, and a
parse that stops for help has produced no value, which is the shape every caller already
handles. clap does the same thing for the same reason. `parse()`, the convenience that reads the
process's own arguments, is the one place that prints and exits.

`-h` renders the short form and `--help` the long one, as clap has them.

The first attempt put the two flags in every command's parse table, which flatten then spliced
twice — a command that flattened another had two `-h` entries and tripped the duplicate-key
check. Recognising them in the parser instead removes the table entries, the metadata mismatch
and the collision at once.

Costs 111 instructions of 29,961, and only on a flag that matched nothing.

---

---

> [!NOTE]
> **Low Risk**
> User-facing behavior change is limited to new help handling on the parse path; embedders must handle `Error::Help` if they use `parse_from` instead of `parse()`.
> 
> **Overview**
> **Built-in help flags** are recognized in the parser (not in per-command tables), so help output and emitted KDL stay aligned with specs that never declare `--help`. `--help` and `-h` are resolved only after a command’s own flags, so CLIs that define their own `help`/`h` still bind them normally.
> 
> Parsing stops with **`Error::Help { cmd, long }`**, targeting the command in scope (e.g. `ex ls --help` → `ls`). **`usage_argv::help::find` / `render`** map that `Command` to metadata and short vs long help text. **`parse_from`** returns the error for embedders; **`parse()`** renders to stdout and exits 0.
> 
> **`usage-derive`** intercepts help flag events before `apply` and special-cases `Error::Help` in `parse()`. **PLAN.md** marks help rendering as wired; the dedicated `help` subcommand remains open. **Conformance tests** cover long/short pages, scoped help, spec omission, custom help flags, help before validation, and wrapper forwarding after `--` / `automatic` double-dash.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12f232106aa9027dc4f8327726eb896fbbb81599. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Where help is recognised, and why it matters

**In the parser, not in any command's table.** That is the only arrangement that does not change
what help *says*: no spec declares `--help`, so a table entry for one would put it in the
rendered page and in the emitted KDL, and the page would then disagree with the spec it came
from. The 211-page parity is unchanged by this PR, which is the check on that claim.

Recognised **after** a command's own flags, so a CLI that declares its own `--help` or `-h` keeps
it and binds it like any other flag.

## How it comes back

`Error::Help { cmd, long }` — carrying the command it was asked about, because `mise config
--help` is a question about `config` and the parser is what knows how far the words reached.

Not printed. A library that writes to stdout on its own is one an adopter cannot embed, and a
parse that stops for help has produced no value — the shape every caller already handles. clap
does the same thing for the same reason. `parse()`, the convenience that reads the process's own
arguments, is the one place that prints and exits.

`-h` renders the short form, `--help` the long one, as clap has them.

## The design that did not work

The first attempt put both flags in every command's parse table. `flatten` then spliced them
twice — a command that flattens another had two `-h` entries and tripped the duplicate-key check
— and the metadata had to gain an exemption from "every flag in the table needs metadata" to keep
them out of the page. Moving the knowledge into the parser removed the table entries, the
metadata mismatch, and the collision together.

## Cost

**111 instructions of 29,961**, and only on a flag that matched nothing — the check sits after the
lookup that failed. 850 ns against clap's 435 µs.

## Verification

Six tests: the two forms differ and ask for different pages; help is about the command the words
reached rather than the root; the flags appear in neither the page nor the emitted spec; a CLI
declaring its own keeps it; help answers even when a required argument is missing (which is when
someone needs it most); and the fixture still parses normally when nobody asks.

Three mutation checks — dropping the long form, dropping the short one, and checking help
*before* a command's own flags, which is what would silently steal a declared `--help`.

## Next

The `help` subcommand — `mise help config ls` — which every CLI with subcommands should have, and
which the rendered page already advertises.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -265,14 +265,14 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
 
 ### Then: what a CLI framework has to have
 
-- [x] **Help rendering** — `-h` and `--help` both match usage-lib byte for byte across all 211
-      of mise's commands: the usage line, the descriptions, commands, arguments and flags grouped
-      by heading, examples, the surrounding text, hidden-item filtering, and for the long form the
-      column alignment and wrapping to `COLUMNS`. What is left is the wiring — `--help`/`-h` on
-      every command, `Error::Help`, and a real `help` subcommand. Holding it to parity found ten
-      things a spec could say that the derive could not, plus two bugs in usage-lib's own
-      renderer; the last of them was `after_long_help`, which 115 of mise's commands use to carry
-      their Examples section, so a page without it was missing what a reader came for.
+- [x] **Help rendering** — `-h` and `--help` both match usage-lib byte for byte across all 211 of
+      mise's commands, and both are wired: the parser recognises them itself, after a command's own
+      flags so a CLI that declares its own keeps it, and a request comes back as `Error::Help`
+      carrying the command it was asked about. `parse()` renders and exits; `parse_from` returns
+      it, so a library embedding this decides for itself. Costs 111 instructions, since the check
+      is only reached by a flag that matched nothing. What is left is the `help` _subcommand_,
+      which every CLI with subcommands should have. Holding the rendering to parity found ten
+      things a spec could say that the derive could not, and two bugs in usage-lib's own renderer.
 - [ ] **Completions, self-contained** — `<bin> completion <shell>` emits the
       script; a hidden `<bin> complete-word` serves requests from the binary's own
       embedded spec. Same dispatch shape usage-cli uses today, without requiring
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -20,6 +20,7 @@
 use core::fmt::Write as _;
 
 use crate::spec::{ArgMeta, CommandMeta, Example, FlagMeta, Spec};
+use crate::Command;
 use crate::DoubleDash;
 
 /// How many flags or arguments are listed individually before collapsing to a placeholder.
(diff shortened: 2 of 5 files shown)


jdx:
The mechanism is right, the behavior is intended, and it is now pinned by a test (df327ba).

A wrapper has two ways to say "hand this on", and help interception sits before both rather than around them:

- after a typed `--`, flag interpretation has stopped, so `--help` and `-h` bind as values;
- after a `double_dash = automatic` argument takes its first value, the same is true without the caller typing a separator — which is the mode a forwarding command should be declaring.

Before either takes effect, `--help` is a question about the wrapper itself: `mise run --help` asks about `run`, and that is also what clap does. The test asserts all three.

`UnknownFlags::Value` on its own is not a forwarding declaration — it is the default for every command — so it cannot be the signal that suppresses help without taking `--help` away from most of the CLI.

jdx:
Note on the force-push: this branch and #872 had been rebuilt on a *parallel copy* of #866's commits rather than on #866 itself, so as pushed they were missing `e02c902` and the spec-fallback fix — merging them would have reverted work from the two PRs below. Both are now rebased onto #866's actual head, and the stack is one linear chain again. No content of this PR changed beyond the added test.

PR #873, opened 2026-08-14
feat(config): say what sort of thing each warning is

jdx:
Groundwork for the config conformance corpus, and useful on its own.

A warning's *wording* is for a person and is nobody's contract. What a program acts on is the kind,
and until now there wasn't one — `Warning` was a string and an origin, so every consumer that wanted
to treat deprecations differently from bad values had to match on text.

```rust
for warning in &resolved.warnings {
    match warning.kind {
        WarningKind::Deprecated => queued.push(warning),      // mise prints these once logging is up
        WarningKind::NotRead => debug!("{}", warning.message),
        _ => eprintln!("{}", warning.message),
    }
}
```

Seven kinds, one per thing this crate can report — a key nothing declares, a value the type cannot
read, a value the choices do not allow, a place that may not set the setting, a deprecation, a value
read as the setting that replaced it, and a value passed over because another name won — plus
`Other`, which is what a CLI's own layer says. A git or pkl layer complaining that it could not reach
its source is not something this crate should make it invent a kind for, so `Warning::new`/`at` keep
their signatures and `of(kind)` is chained where a kind is known.

**Why now:** the corpus has to pin *what happened* without pinning how it was worded — the same line
the argv corpus holds, where a vector names an error class and never a message. Without kinds, a
config vector would either assert exact English (which no second implementation could pass) or assert
nothing about warnings at all, which would leave the scope strips, the rename folding and the type
refusals — much of what the merge does — untested by the corpus.

## Verification

112 tests. One new test walks every kind back to the place that produces it, including the three the
*merge* adds that no layer can know on its own, and asserts a custom layer's warning stays `Other`.
The env test now pins all three kinds a rename produces at once, which is what living through one
actually looks like. Three mutations — the scope refusal, the fold, and the choices refusal each left
unclassified — kill the right tests.

Next: `corpus/config/*.json` and its runner, which is what this is for.

---

> [!NOTE]
> **Low Risk**
> Additive API change (`#[non_exhaustive]` on Warning and WarningKind); behavior is classification only—messages and resolution logic are unchanged aside from the new field.
> 
> **Overview**
> Adds **`WarningKind`** so callers can treat warnings by category instead of parsing message text. Each **`Warning`** now carries a **`kind`** field; **`Warning::new`** / **`Warning::at`** still default to **`Other`**, and known cases use **`.of(WarningKind::…)`**.
> 
> Resolution paths are tagged end-to-end: unknown keys, wrong types, and disallowed choices in **`LayerCtx`**; scope refusals, deprecations, and rename notices in **`resolve`**; env aliases that lose to another variable as **`NotRead`**. **`WarningKind`** is re-exported from the crate root.
> 
> Tests assert kinds at the producers (including merge-only warnings) and pin the three kinds produced when two old env names compete during a rename.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a1b1a671ca1a20b63e4b21b7dd34b82c9751a36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added warning classifications for unknown settings, invalid types, disallowed values, scope issues, deprecated or renamed settings, unread values, and other warnings.
  * Exposed warning categories through the public configuration interface.

* **Bug Fixes**
  * Corrected warning categorization for skipped variables and deprecated or renamed settings.

* **Tests**
  * Expanded coverage to verify warning categories across configuration and resolution scenarios.

### DIFF
--- a/config/src/env.rs
+++ b/config/src/env.rs
@@ -13,7 +13,7 @@
 use std::collections::BTreeMap;
 use std::ffi::OsStr;
 
-use crate::layer::{Layer, LayerCtx, LayerError, LayerOutput, Warning};
+use crate::layer::{Layer, LayerCtx, LayerError, LayerOutput, Warning, WarningKind};
 use crate::registry::PropId;
 use crate::source::{Origin, SourceKind};
 
@@ -133,13 +133,16 @@ impl Layer for EnvLayer {
             for (id, set_as, raw) in candidates {
                 let origin = Origin::new(SourceKind::ENV, set_as);
                 if let Some(first) = read_by {
-                    out.warn(Warning::at(
-                        format!(
-                            "{set_as} was not read: {first} also sets {}",
-                            registry.get(target).key
-                        ),
-                        origin,
-                    ));
+                    out.warn(
+                        Warning::at(
+                            format!(
+                                "{set_as} was not read: {first} also sets {}",
+                                registry.get(target).key
+                            ),
+                            origin,
+                        )
+                        .of(WarningKind::NotRead),
+                    );
                     continue;
                 }
                 match ctx.entry(id, raw, origin) {
@@ -387,6 +390,18 @@ mod tests {
                 .iter()
                 .any(|w| w.starts_with("HK_THREADS was not read: HK_CONCURRENCY also sets jobs")),
             "{warnings:?}"
+        );
+        // Three kinds at once, which is what living through a rename actually produces: the name
+        // that was passed over, and — for the one that *was* read — that it is deprecated and what
+        // it was read as. A variable passed over is its own sort of thing: nothing is wrong with the
+        // value, and a `--strict` mode that stops for a bad one should not stop for this.
+        assert_eq!(
+            resolved.warnings.iter().map(|w| w.kind).collect::<Vec<_>>(),
+            vec![
+                WarningKind::NotRead,
+                WarningKind::Deprecated,
+                WarningKind::Renamed
+            ]
         );
 
         // And the current name still beats both of them.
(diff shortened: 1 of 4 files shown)


jdx:
Both bots, same finding, and it is right: adding a public field breaks a downstream `Warning { message, origin }` literal and any exhaustive destructuring of it.

`Warning` is `#[non_exhaustive]` now. That takes the break once and closes the door: reading the
fields and matching with `..` are unaffected, and the next field this gains — it has gained one
already — costs nobody anything. `WarningKind` is `#[non_exhaustive]` too, for the same reason in the
other direction: a new kind should not break a downstream `match`.

On the version: the crates share one version across the workspace, so which release this rides in is
jdx's call rather than something this PR should decide. Worth knowing when it goes out: the break is
narrow (constructing or exhaustively destructuring a `Warning`, both of which `new`/`at` already
cover), `usage-config` is a day old, and the alternative — carrying the kind beside the warning
rather than on it — would put the two out of step exactly where a caller needs them together.

jdx:
Correcting my own note above, since I flagged a release decision that does not exist: **`usage-config`
has never been published.** The tag `v5.1.0` was cut before the crate existed — it first appears in
`fe13b0b` (#849) — and the manifest reads 5.1.0 only because `shared-version` stamps every crate in
the workspace with the same number.

So there is no downstream `Warning { .. }` literal anywhere to break, and nothing to decide about
which release this rides in. It ships with the field present and `#[non_exhaustive]` from its first
version. The bots' reasoning was sound from the manifest alone; the manifest just is not evidence of
a published crate here.

PR #874, opened 2026-08-14
feat(argv): split a command line the way the shell that typed it would

jdx:
A completion request is a line and a cursor, not an argv: the user has pressed Tab
in the middle of something the shell has not run and may never run. Four of the
five shells hand that over directly — bash's COMP_LINE/COMP_POINT, zsh's
$BUFFER/$CURSOR, fish's `commandline -cp`, PowerShell's extent plus
$cursorPosition — and nushell, whose external completer only ever sees spans,
re-quotes them into a line.

Taking the line rather than the shell's own word split is what lets `mise run "my
ta⌶` complete inside a quoted word at all: a split has already thrown away the
quote that made the space part of the word. The cost is that the splitting is ours,
which is what the tests are — every rule here fails a test when inverted.

The words come out unquoted, because they are what argv would have held had the
line been run, so the parser downstream sees what it would see in a real
invocation. Two rule sets rather than five: POSIX-shaped for bash, zsh, fish and
nu, backtick escapes for PowerShell, which is also why a Windows path survives
being completed.

A cursor in a gap gets a word made for it — `mise ⌶use` asks what can go *before*
`use`, and answering about `use` would complete the wrong word — and an empty line
is a completion request like any other rather than a special case for the caller.

Behind a `complete` feature, like `spec`: a CLI can want either without the other.

---

---

> [!NOTE]
> **Low Risk**
> New optional module behind a feature flag with extensive unit tests; no changes to default parse path or existing `spec` behavior.
> 
> **Overview**
> Adds an optional **`complete`** feature and a new **`argv::complete`** module so completion can turn a shell line plus cursor into unquoted words, **`cword`**, and **`prefix`**—the shape a parser would see on a real run, including inside quotes.
> 
> **`split(line, cursor, shell)`** implements POSIX-style rules for bash/zsh/fish/nu and PowerShell-specific backtick escapes and doubled-quote escaping. It handles cursors in gaps (inserts an empty word), after trailing spaces, on escaped characters, and clamps bad byte offsets without panicking. **`Split::walked()`** exposes only words up through the word being completed.
> 
> The **`conformance`** crate enables **`complete`** alongside **`spec`** for oracle/testing. Default builds stay unchanged when the feature is off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e245189054417f088377def2c2aea0337eb01d64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Why a line and not the shell's words

The design question here was what the hidden completion command takes. jdx chose a raw line plus a cursor offset over mirroring `usage complete-word`'s words/`--cword` interface, and it is the more primitive input in every shell that has one — the word split is something the shell derived and can only lose information.

It does give up running this implementation and usage-lib's on identical inputs. The oracle moves one layer in instead: this crate splits, and the candidate stage (next PR) can be compared against `usage complete-word` fed the words this produced. So the splitting is held by these tests and the candidates stay held by the reference.

## What each shell can supply

| shell | line | cursor |
|---|---|---|
| bash | `COMP_LINE` | `COMP_POINT` |
| zsh | `$BUFFER` | `$CURSOR` |
| fish | `commandline -cp` | implicit (line ends at the cursor) |
| PowerShell | `$commandAst.Extent.Text` | `$cursorPosition` |
| nu | — spans only | — |

nushell's external completer never sees a line, so its generated script will re-quote the spans into one. That is lossless in the direction that matters, since nushell has already done the splitting this module would do.

### DIFF
--- a/argv/Cargo.toml
+++ b/argv/Cargo.toml
@@ -18,6 +18,10 @@ license = { workspace = true }
 # Cold-path metadata and spec emission. Off by default so a CLI that wants only
 # a parser does not carry it.
 spec = []
+# Splitting a command line for completion. Off by default for the same reason, and
+# separate from `spec` because a CLI can want one without the other: completion needs
+# to split a line whether or not the binary can also emit its own spec.
+complete = []
 
 [package.metadata.release]
 shared-version = true
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -60,6 +60,11 @@ impl Shell {
     fn backtick_escapes(self) -> bool {
         matches!(self, Shell::PowerShell)
     }
+
+    /// Whether a quote inside a quoted string is written by doubling it.
+    fn doubles_quotes(self) -> bool {
+        matches!(self, Shell::PowerShell)
+    }
 }
 
 /// A command line as the shell would have passed it, plus where the cursor was.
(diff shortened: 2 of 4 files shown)


jdx:

**Cursor on escaped characters is skipped** — the loop only checked the position of characters it stopped at, and an escape consumes the next one without stopping. `mise run my\ ⌶task` therefore described the last word of the line rather than the one being typed, filtering candidates by the wrong prefix or asking about the wrong argument entirely. The check now also runs where the escape consumes — which is why it is a macro rather than a closure: it writes to the same four locals the loop does. Tested at the escaped character itself in both an unquoted word and inside double quotes.

**PowerShell quote doubling is ignored** — the module's own documentation claimed the rule and the code did not implement it; `'it''s` came back with the quote dropped. Now done, and only for PowerShell: in the POSIX-shaped shells a second quote ends the string, which the tests pin in both quote characters, since a rule applied to the wrong shell is the same bug pointing the other way.

PR #875, opened 2026-08-14
feat(config): a conformance corpus for resolution

jdx:
The argv corpus pins which token becomes which flag. This pins what a *resolution* is: given a
registry and layers that supply values, which value wins, where it is recorded as coming from, and
what the merge had to say about the ones it refused.

**46 vectors in six sections** — precedence, merging, scope, renames, types, choices — each with a
sentence saying what it establishes, because a vector nobody can read is a vector nobody will
maintain.

## A registry, not a spec

An argv vector carries a KDL spec, because parsing a command line *is* a question about a spec. A
resolution is not: it is a question about a registry — keys, types, defaults, merge policies — which
a CLI's build step produces long before anything is resolved. So a vector describes the registry
directly:

```json
{
  "id": "cli-beats-env-beats-file",
  "doc": "The declared order decides: the command line, then the environment, then a file.",
  "settings": [{ "key": "jobs", "type": "uint", "default": 4 }],
  "layers": [
    { "source": "cli", "id": "--jobs", "values": { "jobs": "8" } },
    { "source": "env", "id": "EX_JOBS", "values": { "jobs": "6" } },
    { "source": "file", "id": "hk.toml", "values": { "jobs": "2" } }
  ],
  "expect": { "values": { "jobs": 8 }, "origins": { "jobs": "--jobs" } }
}
```

An implementation in Go or TypeScript can run this without first acquiring a KDL parser, and how a
spec *becomes* a registry stays `usage-config-build`'s question, which its own golden test answers.

**Warnings are pinned by kind, never by message** — `out-of-scope`, `renamed`, `wrong-type` — which
is why #873 comes first. Wording is a quality-of-implementation concern and is expected to differ,
the same line the argv corpus holds for its error codes.

**Nothing here touches the world.** No file is opened, no environment read, no process started: a
layer is a description of what it supplies, which is all the merge ever sees of one. So no vector's
result can depend on the machine running it.

**The harness reads the type grammar itself** rather than asking the crate under test. A conformance
harness that asked the implementation what `list<string>` means would be checking that it agrees
with itself.

## What it caught while being written

`union` on a **list** keeps duplicates and only a `set` drops them — I had written the vector the
other way round. The implementation is right (that distinction is what the two types are *for*), and
there are now two vectors saying so, one per type.

## Verification

Four mutations of the runtime, each failing the vectors that describe it rather than a test written
beside it: the scope check letting everything through (4 vectors), union forgetting the lower layer
(4), the boolean spellings narrowed to `true` (3), and choices unenforced (2). Corpus
well-formedness — unique ids, a doc on every vector, no empty sections — is checked by the same test
file.

## No divergence field

The argv corpus records where usage-lib disagrees with the grammar, because that grammar predates it
and has two implementations. Resolution has one, so a disagreement is a bug to fix rather than a fact
to record — and every vector applies to any implementation, since there is no layer of a resolution
that can be left to somebody else the way a binding-only argv parser leaves `required` to the layer
above it.

Stacked on #873, which adds the warning kinds this pins.

---

> [!NOTE]
> **Low Risk**
> Test-only and corpus additions; resolution behavior is exercised but production resolver code is not changed in this diff.
> 
> **Overview**
> Adds a **config resolution conformance corpus** parallel to the existing argv JSON corpus: **46 KDL vectors** in `corpus/config/` (precedence, merging, scope, renames, types, choices) that pin registry + layered inputs → resolved values, origins, and **warning kinds** (not message text).
> 
> **`usage-conformance`** gains a `config` module (~850 lines): KDL parse/load, in-memory `Layer` stubs (no real env/files), a standalone type grammar for building registries, and `run`/`matches` against `usage-config::resolve`. Integration tests in `conformance/tests/config.rs` run every vector and enforce corpus shape (unique ids, etc.). **`load_as<T>`** generalizes corpus directory loading; argv `load` delegates to it.
> 
> Docs: `corpus/config/README.md`, link from `corpus/README.md`. Depends on **`kdl`** and **`usage-config`** in the conformance crate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6caadebcdb8707d88cb6e74802026baa3d5f875d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2106,9 +2106,11 @@ name = "usage-conformance"
 version = "0.0.0"
 dependencies = [
  "insta",
+ "kdl",
  "serde",
  "serde_json",
  "usage-argv",
+ "usage-config",
  "usage-derive",
  "usage-lib",
 ]
--- a/conformance/Cargo.toml
+++ b/conformance/Cargo.toml
@@ -12,9 +12,11 @@ authors = { workspace = true }
 license = { workspace = true }
 
 [dependencies]
+kdl = "6"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 usage-argv = { workspace = true, features = ["spec", "complete"] }
+usage-config = { workspace = true }
 usage-lib = { workspace = true }
 
 [dev-dependencies]
--- a/conformance/src/config.rs
+++ b/conformance/src/config.rs
@@ -1,0 +1,848 @@
+//! The config conformance corpus: its format, and the runner that answers it.
+//!
+//! The argv corpus pins which token becomes which flag. This one pins what a *resolution* is: given
+//! a set of settings and some layers that supply values, which value wins, where it is recorded as
+//! coming from, and what the merge has to say about the ones it refused.
+//!
+//! # Why a registry rather than a spec
+//!
+//! An argv vector carries a KDL spec, because parsing argv is a question about a spec. A resolution
+//! is not: it is a question about a *registry* — keys, types, defaults, merge policies — which a
+//! CLI's build step produces from the spec long before anything is resolved. So a vector describes
+//! the registry directly in KDL, usage's canonical interchange format. How a spec becomes a
+//! registry is `usage-config-build`'s question, and its own golden test answers it.
+//!
+//! Nothing here touches the filesystem, the process environment, or a subprocess. A file layer is a
+//! description of what a file said, which is all the merge ever sees of one.
+
+use std::collections::BTreeMap;
+use std::path::Path;
+
+use kdl::{KdlDocument, KdlNode, KdlValue};
+use serde::{Deserialize, Serialize};
+use usage_config::{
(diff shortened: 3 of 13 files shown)


jdx:
Real, and in a conformance harness it is the worst class of bug: `as_i64().unwrap_or_default()` turned
a number too large for the value model into **zero**, so the vector went on to pass or fail for a
reason that has nothing to do with what it says. A harness that quietly misreads its own corpus is
worse than one that cannot read it.

Reading a number is fallible now, and the failure reaches both roads a number arrives by — a declared
default or a choice (through `registry_of`, which already returned a `Result`) and a value a layer
supplies (through `LayerError::Unreadable`, naming the key and the layer). The run stops instead of
resolving something nobody wrote.

Test: a vector with `9223372036854775808` as a default, and the same number supplied by a layer.
Mutation: zeroing it again fails both halves.

PR #876, opened 2026-08-14
feat(argv): read the cursor's position off a real parse

jdx:
What can be typed at a cursor is decided by everything to its left: which command
the words reached, whether a flag is still waiting for its value, which positional
comes next, and whether flag interpretation is still on. `walk` reads that state
off an actual parse of the words before the cursor, so what a completion offers
and what the parser would accept are decided by the same tables — including the
shadowing rule, where a subcommand redeclaring an inherited global hides it.

A parse error is not a failure here. A line being completed is unfinished by
definition, so "the grammar runs out here" is the position being asked about
rather than a reason to discard the attempt. One error says something more: a
missing flag value means the cursor is standing in that value, which is what
decides the candidates.

`flags_possible` asks the wider question than `double_dash_seen`: past a
separator *or* past the first value of an `automatic` argument, a dash-prefixed
word is a value, and the flags there belong to whatever the wrapper forwards to.

`Split::argv` keeps the index arithmetic in one place. Two words are dropped for
two reasons — the program name, because argv does not contain it, and the word
being completed, because feeding a half-typed word in asks what can follow it when
the question is what it could be.

Three small accessors on the parser rather than a second walker: `flags_stopped`,
`pending_arg`, `flags_in_scope`. Every rule here fails a test when inverted.

---

---

> [!NOTE]
> **Low Risk**
> Completion-only API and a targeted parser state fix for EOF during variadic flags; behavior is covered by new tests and does not alter successful full-line parsing paths.
> 
> **Overview**
> Adds **shell completion positioning** in `complete`: `Position` and **`walk`** parse the words before the cursor (via `Split::argv`) and report the active command, whether flags are still allowed, any flag waiting for a value, and the next positional—using the same parser tables as a real invocation, including help-topic and variadic-flag edge cases.
> 
> Exposes **`flags_stopped`**, **`collecting`**, **`pending_arg`**, and **`flags_in_scope`** on `Parser` so completion can query state without duplicating binding logic.
> 
> Fixes variadic-flag **collecting** so reaching **end of argv** no longer clears `collecting`; only a following token that ends the run does, so completion can tell the next word still belongs to the flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abea8b716d0cbc0f1ef94166ee9e332f3a4b5d1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -15,6 +15,62 @@
 //! as argv had the line been run — the parser downstream should see exactly what it would see
 //! in a real invocation.
 
+use crate::{Arg, Command, Error, Flag, Parser};
+
+/// Where the cursor is, in the grammar rather than in the line.
+///
+/// What can be typed at a cursor is decided by everything to its left: which command the words
+/// reached, whether a flag is still waiting for its value, which positional comes next, and
+/// whether flag interpretation is still on at all. This is that state, read off a real parse
+/// of the words before the cursor rather than guessed at — so what is offered and what would
+/// be accepted are decided by the same tables.
+#[derive(Debug, Clone, Copy)]
+pub struct Position<'t> {
+    /// The command in scope: the deepest one the words selected.
+    pub cmd: &'t Command<'t>,
+    /// Whether a dash-prefixed word here would still be read as a flag.
+    ///
+    /// False past a `--`, and past the first value of an `automatic` argument — a wrapper
+    /// forwarding to another tool. There is no flag of *this* CLI to offer in either place.
+    pub flags_possible: bool,
+    /// A flag whose value the cursor is at, if the last word was one that takes a value.
+    pub awaiting_value: Option<&'t Flag<'t>>,
+    /// The positional a word here would fill, if any are left.
+    pub next_arg: Option<&'t Arg<'t>>,
+}
+
+/// Walk the words before the cursor, and report what the cursor is at.
+///
+/// Errors are not failures here. A line being completed is by definition unfinished — a flag
+/// with no value yet, a word that names nothing yet — so a parse error means "the grammar runs
+/// out here", which is exactly the position being asked about. The walk stops at the first one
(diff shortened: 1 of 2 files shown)


jdx:
Two findings, one fixed and one matching the reference.

**Help path command is discarded** (Cursor) — correct, and fixed in 4497245. `ex help config ⌶` asks which command to read about; the parser deliberately never descends into a topic, so it was still standing at the parent and offering the parent's candidates. The position now comes from the help request that already resolved the path, and says what else is possible there: nothing. A topic takes no flags and fills no argument, so neither is left in scope. Tested at both levels — the position, and the candidates that follow from it.

**Attached flag values are skipped** (greptile) — real, and shared with the reference, so it is not a drift. Both sides asked the same lines over mise's spec:

```
"mise use --jobs="          ours: []      theirs: []
"mise use --jobs=4"         ours: []      theirs: []
"mise config ls --format="  ours: []      theirs: []
```

usage-lib takes the whole token as `ctoken`, sees it start with `--`, and filters flag *names* by it — so `--jobs=` matches none and it offers nothing, exactly as here. Completing after an `=` would be an improvement to make on both sides at once rather than a difference to introduce on one; noted for the stage that adds value completions, where it belongs.

jdx:
Correct, and fixed in the head commit — the cause was one line deeper than the walk.

A variadic flag stops claiming words when it meets a token that could be something else, and the end of argv was being treated as such a token. It is not a token at all: once argv is exhausted no more events are produced either way, so clearing there decided nothing about parsing and threw away the answer to "would the next word be claimed?" — the question a completion asks and a parse never does.

So exhaustion now decides nothing, and `collecting()` reports the truth at the end of the walk. `mise use --tools node ⌶` says the next word belongs to `--tools`, and `mise use --tools node -- ⌶` says it does not, because there the run really was ended by a token.

PR #877, opened 2026-08-14
feat(argv): offer what the reference offers, from compiled tables

jdx:
The candidates for a cursor: the flags a word there could name, the subcommands it
could select, and the choices a flag or positional declares. The rules are
usage-lib's, because a CLI's completions should not change with the
implementation that answers them — a lone `-` offers both forms, `--…` offers
longs, `-…` narrows to one letter, a flag waiting for its value offers that flag's
choices and nothing else.

Held to the reference over mise's real spec: of eighteen lines across the tree,
thirteen agree exactly, including sets of forty and twenty-one candidates. The
five that do not are named in the test rather than dropped from it — two need
stages that do not exist yet (`run=` completions, the reserved `type=`
vocabulary), and two are one derive gap.

That gap is worth its own note: `flag "-p --path --file"` has a second long form
the derive cannot declare, so `gen-shadow` drops it. It is invisible in the help
comparison, which renders one long form per flag, and shows up here because a
completion offers every name a flag answers to. Next PR.

Comparing found a real bug, which is what comparing is for: hidden aliases were
being offered. The parse table cannot tell a hidden alias from a visible one —
both must be *accepted* — so which are meant to be *offered* is a question only
the metadata answers.

Descriptions come from the metadata the binary already carries. The reference
leaves a flag's empty, its own source saying `TODO: get flag description`; filling
it in is a deliberate improvement, which is why the comparison is on values.

`usage_cli::complete_candidates` is a seam, not a feature: "the same candidates"
is only a checkable claim if the reference's answer can be read as data rather
than watched going past on stdout.

---

---

> [!NOTE]
> **Medium Risk**
> Completion behavior is user-facing and must stay in sync with the reference; gaps are documented but wrong candidates or hidden-alias leaks would confuse shell users. The API surface grows (`Candidate`, richer `Position`, feature coupling `complete` → `spec`) without touching auth or data persistence.
> 
> **Overview**
> Adds **spec-driven shell completion** in `usage-argv`: `candidates()` turns a split command line plus static `Spec` metadata into `Candidate` values (flags, subcommands, declared choices, descriptions), aligned with usage-lib rules. **`Position` / `walk`** now track `--`, help topics, and in-scope flags so offerings match what the parser would accept, including hidden commands/aliases, negated long flags, restart tokens (`:::`), root **default subcommand** first-arg values, and positionals that require `--` before values.
> 
> The **`complete` feature** now enables **`spec`**. **Gate tests** compare candidate *values* from the shadow mise CLI against `usage_cli::complete_candidates` on real mise lines (documented gaps: `run=` completions, `type=config_keys`, extra long forms from shadow gen).
> 
> **`usage-cli`** exposes `complete_candidates` and makes `complete_word` public so parity tests read reference output as data instead of stdout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 652593878a2bfca6e04c8d62811bee8f5146c961. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -793,6 +793,7 @@ dependencies = [
  "shadow-mise",
  "shadow-mise-clap",
  "usage-argv",
+ "usage-cli",
  "usage-lib",
 ]
 
--- a/argv/Cargo.toml
+++ b/argv/Cargo.toml
@@ -21,7 +21,7 @@ spec = []
 # Splitting a command line for completion. Off by default for the same reason, and
 # separate from `spec` because a CLI can want one without the other: completion needs
 # to split a line whether or not the binary can also emit its own spec.
-complete = []
+complete = ["spec"]
 
 [package.metadata.release]
 shared-version = true
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -15,6 +15,7 @@
 //! as argv had the line been run — the parser downstream should see exactly what it would see
 //! in a real invocation.
 
+use crate::spec::{ArgMeta, CommandMeta, FlagMeta, Spec};
 use crate::{Arg, Command, Error, Flag, Parser};
 
 /// Where the cursor is, in the grammar rather than in the line.
@@ -24,7 +25,7 @@ use crate::{Arg, Command, Error, Flag, Parser};
 /// whether flag interpretation is still on at all. This is that state, read off a real parse
 /// of the words before the cursor rather than guessed at — so what is offered and what would
 /// be accepted are decided by the same tables.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct Position<'t> {
     /// The command in scope: the deepest one the words selected.
     pub cmd: &'t Command<'t>,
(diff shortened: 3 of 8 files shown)


jdx:
Three findings, all correct, all fixed in the head commit — and the third exposed a fourth.

**Negate flags offered without dashes** (Cursor) — the table holds a negation the way the parser matches it, with the dashes already taken off ([`model.rs:772`](https://github.com/jdx/usage/blob/main/derive/src/model.rs#L772) strips them), so offering it bare matched no `--` the user had typed and put a word on a lone `-` that no shell would accept. Candidates now put them back.

**Restart-token state is ignored** (greptile) — `mise task one ::: ⌶` starts a fresh invocation, so the cursor is back at the *first* argument rather than at the second one the words had reached. Implemented from the metadata, where `restart_token` lives.

**Default-subcommand arguments are omitted** (greptile) — `mise build` means `mise run build`, so what the fallback command's first argument accepts is a candidate at the root too. Only its first, and only at the root, which is the rule usage-lib follows.

That last one then offered those values after `help`, where only a command name belongs — so the position now records that it is a help *topic* and the fallback rule skips it.

Worth noting for anyone reading the tests: my first restart-token test proved nothing. The fixture's argument was variadic, so "back at the first" and "wherever the words reached" were the same answer, and removing the branch entirely left the test green. It now uses a command with two differently-typed arguments, and fails without the branch.

jdx:

**Required separator never offered** — an argument declared `double_dash = "required"` is not fillable until a `--` has been typed, so offering its choices was offering words the parser would reject. It now follows the reference's rule exactly ([`complete_word.rs:490`](https://github.com/jdx/usage/blob/main/cli/src/cli/complete_word.rs#L490)): the separator itself, and only when nothing has been typed to filter it away. Applied through one helper, so the restart-token branch gets it too — which was the second half of your point.

That needed `separator_seen` on the position, narrower than `flags_possible`: the latter is also false past an `automatic` argument, and an argument that requires a separator is asking for that specific token.

**Default subcommand alias not resolved** — real. `default_subcommand` names a command, and a spec may name it the way its author refers to it rather than canonically; matching only `cmd.name` silently dropped the fallback's values. It now matches any name the command answers to, and the fixture names it by an alias so the lookup is under test rather than assumed.

jdx:
Half of this matches the reference, half is a parser gap worth naming.

`restarted` only treating the *immediately* preceding word as a restart is exactly usage-lib's rule — [`complete_word.rs:113`](https://github.com/jdx/usage/blob/main/cli/src/cli/complete_word.rs#L113): `parsed.cmd.restart_token.as_ref().is_some_and(|rt| prev_token == Some(rt.as_str()))`. So `::: value ⌶` follows the pre-restart slot on both sides, and changing it here alone would make the two implementations disagree about a CLI's completions, which is the thing this stage is trying not to do.

The deeper observation is right, though: the *parser* does not implement restart tokens at all — `restart_token` is metadata, nothing in the parse tables resets on it — so `next_arg`, `separator_seen` and `flags_possible` all describe the invocation before the token. Completion inherits that. Fixing it properly means teaching the parser to restart, which changes binding as well as completion and belongs in its own PR rather than being approximated here. Filed as the next parser gap after this stack.

PR #883, opened 2026-08-14
agent/complete files

jdx:

---

> [!NOTE]
> **Medium Risk**
> Completion behavior changes for many cursor positions (closed vs open sets, separators, restarts); mistakes would show up as wrong Tab suggestions but not as runtime CLI parsing errors.
> 
> **Overview**
> **Tab completion now returns structured answers** — known `candidates` plus an optional **`files`** hint (`Any` or `Dirs`) so generated shell scripts can call native path completers instead of the CLI reading the directory.
> 
> **`complete()`** wraps the existing **`candidates()`** logic and decides when paths apply: argument/flag placeholder names like `FILE`, `PATH`, `DIR` (usage-lib’s name rule); **no** paths for dash-prefixed tokens, declared choice sets (including empty prefix matches), help topics, or positionals that still need `--`; **yes** for open positions (e.g. mistyped subcommand) and after restart tokens (`:::`), aligned with which argument the cursor is on.
> 
> **Conformance tests** assert candidate parity with the reference and treat reference directory listings as equivalent to `files: Some(Files::Any)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e8f0dd4d0413aeac424da2e32d8e8d9356296a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added shell file and directory completion for arguments and flags that accept paths.
  - Improved completion for commands, options, help topics, and arguments requiring separators.
  - Added completion support for the `edit` and `pipe` commands.

- **Bug Fixes**
  - Prevented inappropriate path suggestions for flags, closed candidate lists, and undeclared choices.
  - Improved fallback behavior for file-based arguments.
  - Ensured compiled and reference completions provide consistent results.

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -179,8 +179,19 @@ fn files_for(name: &str) -> Option<Files> {
 /// for a word nothing is known about, and the `run=` completions a spec can declare.
 pub fn complete<'a>(spec: &'a Spec<'a>, split: &Split) -> Completions<'a> {
     let position = walk(spec.root.cmd, split.argv());
+    let meta = crate::help::find(spec, position.cmd).map(|(_, meta)| meta);
     let token = split.prefix.as_str();
     let candidates = candidates(spec, split);
+
+    // Which argument the cursor is at — the same question `candidates` answers, asked once so
+    // that the two halves cannot disagree. Past a restart token it is the command's *first*
+    // argument, whatever the words before the token filled, and everything below follows from
+    // that: whether paths belong, whether the set is declared, whether a separator is owed.
+    let at_cursor = if restarted(meta, split) {
+        meta.and_then(|m| m.args.first()).map(|m| m.arg)
+    } else {
+        position.next_arg
+    };
 
     // A dash-prefixed word is a flag or nothing: no path starts with one, and the reference
     // suppresses its own listing there for the same reason.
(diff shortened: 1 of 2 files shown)


jdx:
Both of you found the same thing, which is usually a sign. Fixed in a9503fc, along with the third.

**Closed set opens on mismatch** — correct, and it was the exact failure the comment beside it claimed to prevent. The condition read the *filtered* list, which answers a different question: "nothing matched what you typed" was being treated as "there is nothing else this can be". Both questions matter and the reference asks both — was anything found (`choices.is_empty()`), and does the position declare its set (`has_explicit_choices`). A mistyped *command* still falls back to paths, because nothing declared a set there, which is the reference's behaviour too.

**Paths offered before separator** — also correct. An argument that requires `--` is not a path yet however it is named: until the separator is typed the parser rejects a path exactly as it rejects any other value. That check now comes before the by-name rule.

The conformance comparison gained the case, since the reference is the oracle for it — `mise activate zsx` over mise's real spec, where the argument declares `bash elvish fish nu xonsh zsh pwsh`. Worth noting why that matters: a test written against the *filtered* list would have passed while the rule was wrong, which is how this got in.

jdx:
Correct, and fixed in the head commit.

`needs_separator` was read from `next_arg` whether or not a flag was waiting for its value — so a command with an unfilled `double_dash`-required positional suppressed paths for the flag value being typed. Those are two different positions that happen to sit near each other: `ex --from ⌶` takes a path whatever the argument after it needs, which is also where the reference applies the rule and where it does not.

The fixture now has both on one command, so the test fails if the rule leaks across.

jdx:
Correct, and it was the sharper form of the same problem as the earlier findings: two halves of one answer disagreeing about which argument the cursor is at.

The candidates knew about the restart token and the path decision did not, so after `::: ` one answered about the command's first argument and the other about whatever the words before the token had left unfilled. Asked once now, with both halves following from the answer — which argument, then whether paths belong, whether the set is declared, whether a separator is owed.

The fixture needed rearranging to prove it: its first argument declares choices and its second takes paths, so "back at the first" and "wherever the words reached" are different *answers* rather than two routes to the same one. My first attempt had them coincide, and removing the rule entirely left the test green.

PR #885, opened 2026-08-14
feat(derive): answer a completion request from the binary itself

jdx:
`ex __complete_word__ --shell zsh --line "ex ins" --cursor 6` — the hidden command
a generated script calls, answering from the same tables the parse uses. What this
deletes for an adopter is the reason it exists: mise's shipped completions today
hard-fail unless the separate `usage` CLI is installed, dump a spec into
`$XDG_CACHE_HOME` with `mise usage`, and shell out to `usage complete-word` on
every Tab. This is one process and no spec file.

Opt in with `#[usage(bin = "…", completion)]`, not supplied like `--help`: it is a
hidden command a binary carries and a protocol a script depends on, so a CLI says
when it wants both. The script generator will be emitted under the same flag, which
makes a script calling a command the binary lacks a compile error rather than a
puzzle at the prompt. A CLI that does not ask for it carries nothing — mise's
shadow does not, and the reported cost is unchanged at 29,962 instructions.

Recognized before the parse rather than inside it. A completion is not a command
this CLI runs, and putting it in the tables would make it one: visible to the
grammar, to the help, and to the emitted spec. So `ex run __complete_word__` still
binds that word to a positional, which a test holds.

Its three arguments are read by hand for the same reason, and read forgivingly —
an unknown one is ignored and a missing cursor means the end of the line, which is
where nushell's completer leaves it. A completion that errors out is a shell that
beeps at every keystroke.

Rendering is per shell, in `argv`: bash reads values, fish and nu and PowerShell
take a description after a tab, zsh takes a third field for what to *type*, since
a candidate with a space in it has to reach the command line intact.

---

---

> [!NOTE]
> **Medium Risk**
> New opt-in completion path touches `parse()` entry and stdout/exit(0) on completion; mistakes could affect normal runs only for CLIs that enable the attribute, but wrong interception would be user-visible at the shell.
> 
> **Overview**
> **In-process tab completion** replaces shelling out to an external `usage` CLI: opt in with **`#[usage(completion)]`** on the root `#[derive(Cli)]` ( **`completion = false`** is honored via `flag_value`).
> 
> The derive emits **`completion_request`** and intercepts **`parse()`** when argv starts with **`__complete_word__`**, manually reading `--shell`, `--line`, and optional `--cursor`, then answering from **`Self::spec()`** via **`complete::split`**, **`complete`**, and new **`render`**. Completion is **not** part of the parse grammar, so normal invocations (e.g. `run __complete_word__` as a positional) stay unchanged.
> 
> **`usage-argv::complete`** gains **`render`**, **`FILES_MARKER` / `DIRS_MARKER`**, **`one_line`** (multi-line help collapsed for line protocols), and **`zsh_quote`**. **`__usage_needs_complete_feature!`** gives a clear compile error when the `complete` feature is missing.
> 
> Conformance tests in **`completion.rs`** cover candidates, per-shell output shapes, cursor handling, forgiving unknown flags, and **`completion = false`** not generating a colliding method.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cf62132d2363dd6d7e55a2f908ccbfff4e82afa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -148,6 +148,85 @@ pub struct Completions<'a> {
     pub files: Option<Files>,
 }
 
+/// The line a shell reads to mean "paths belong here too".
+///
+/// A whole line rather than a flag on the protocol, because every one of the five shells can
+/// already split output into lines and look at the last one. `\x01` opens it because no
+/// candidate this crate produces can contain a control character, so it cannot be mistaken for
+/// one.
+pub const FILES_MARKER: &str = "\u{1}files";
+/// See [`FILES_MARKER`]. Directories only.
+pub const DIRS_MARKER: &str = "\u{1}dirs";
+
+/// Write an answer the way `shell` reads it.
+///
+/// One line per candidate, in the shape the shell's own completion machinery expects — which is
+/// where the five differ. bash reads values only; fish, nu and PowerShell take a description
+/// after a tab; zsh takes a third field, the text to insert, because what it displays and what
+/// it types are not always the same string.
+///
+/// A trailing [`FILES_MARKER`] says the generated script should hand the position to the
+/// shell's own path completion afterwards.
+pub fn render(answer: &Completions<'_>, shell: Shell) -> String {
+    let mut out = String::new();
+    // Descriptions are all-or-nothing per answer: a column that appears on some rows and not
+    // others reads as missing data rather than as an absent description, which is the reason
+    // the reference decides this per answer too.
+    let described = answer.candidates.iter().any(|c| c.description.is_some());
+
+    for candidate in &answer.candidates {
+        let description = candidate.description.unwrap_or_default();
+        match shell {
+            Shell::Bash => out.push_str(&candidate.value),
+            Shell::Zsh => {
+                // Display, then description, then what to type: a candidate containing a space
+                // or a quote has to reach the command line intact.
+                out.push_str(&candidate.value);
+                out.push('\t');
+                out.push_str(description);
+                out.push('\t');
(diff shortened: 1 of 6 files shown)


jdx:
All three correct, all fixed in 277f51a.

**Descriptions break line protocol** — and not defensively: mise declares multi-line `help` on 37 of its commands and flags, so a candidate would routinely have split into several rows or invented a field. Collapsed rather than truncated at the first line, because two lines are one sentence to a reader and half of it is worse than all of it spaced. Tested for both the row count and zsh's field count.

**Completion feature dependency is implicit** — right, and the error named the symptom (`unresolved module complete`) rather than the attribute that caused it. A macro defined under both cfgs now expands to `compile_error!` with what to add and where, and the derive's docs say it too.

**Args silently ignores completion** — refused there now, with the reason: the hidden command it adds answers for the whole program, not for one of its commands. Verified the message actually fires by compiling an `Args` that declares it.

jdx:
Correct. It went through the same helper as `global` and `var` now, which reads the bare word as true and a boolean as itself — so `completion = false` means false.

Worth recording how it was verified, because the first attempt was worthless: asserting an *absence* needs something to collide with. The test declares a `completion_request` of its own on the declining CLI, and Rust refuses two inherent methods with one name — so if the attribute were still read as opting in, the crate would not build. My first version only checked that the CLI parsed, which it did either way, and the mutation sailed through it.

PR #887, opened 2026-08-14
feat(argv): generate the shell script each shell wants

jdx:
Five scripts, each a handful of lines, because the thinking is on the other side:
hand over the line and the cursor, present what comes back the way this shell
presents things, and pass the position to the shell's own path completion if the
marker appeared.

What they replace is the point. mise's current scripts hard-fail unless the
separate `usage` CLI is installed, dump a spec into `$XDG_CACHE_HOME` with `mise
usage`, prune stale spec files by age, and shell out to `usage complete-word` on
every Tab. None of that is here: one process, no spec file, no second program.

Where a shell already knows how to do something it does it — zsh's `compadd` with
descriptions aligned to the longest value, fish's `__fish_complete_path`,
PowerShell's own `CompletionCompleters`. The zsh presentation is the reference's,
including why it is `compadd` and not `_describe`: the latter groups matches sharing
a `:`-separated prefix and shows one per group, so mise's `release:create`,
`release:docs-sync` and `release:pr` would collapse into one entry.

The marker is written in each shell's own notation rather than one escape hoping to
travel — which it did not: Rust's `\u{1}` reached bash as five literal characters,
and the first test run caught the marker being offered to the user as a candidate.

Tested by running the shells. bash, zsh and fish are all driven for real against a
stand-in binary: candidates, descriptions, the files marker, directories only, a
value that needs quoting, and zsh's menu decision. zsh's builtins are stubbed so
the function can be called at all, which leaves exactly the part this crate wrote
under test. All three scripts are also parsed by their own shell, since a syntax
error in generated text reaches a user as a broken prompt. nu and PowerShell are
not installed here and are unchecked — said plainly rather than left to be assumed.

`completion_script` is emitted under the same attribute as the command it calls, so
a script naming a command the binary does not answer is a compile error instead of
a silence at the prompt.

---

---

> [!NOTE]
> **Medium Risk**
> New user-facing generated shell code and completion wiring; mistakes show up as silent Tab failures, though coverage includes real-shell tests and strict script/binary pairing.
> 
> **Overview**
> Adds **in-process shell completion scripts** that call the CLI’s hidden `__complete_word__` handler instead of the separate `usage` CLI, cached specs, or `usage complete-word`.
> 
> A new `usage_argv::script` module (behind the `complete` feature) exposes `script(bin, shell)` and emits per-shell installers for **bash, zsh, fish, nushell, and PowerShell**. Each script passes the **line truncated at the cursor in that shell’s own units**, renders tab-separated candidates (and zsh’s aligned descriptions / insert quoting), honors `\001files` / `\001dirs` markers via native path completion, and **refuses invalid binary names** that cannot be registered safely.
> 
> The derive now emits **`completion_script(shell)`** next to `completion_request`, wired to the spec’s bin name. **Integration tests** run real bash/zsh/fish against stand-in binaries; conformance checks that generated scripts call the same command the binary answers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb1286e21e9a1c2a9b8b8c0c105484548d26cb0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Three findings, and the two cursor ones were the same bug from different angles. Fixed by removing the offset from the protocol rather than converting it.

**Cursor units / non-ASCII** — correct, and worse than a units mismatch: every shell counts in its own way. Measured here rather than guessed — bash and zsh both report 15 for `"ex ünicode here"` (15 characters, 16 bytes) and slice consistently with that, while `split` was treating the number as a byte offset.

So no script passes a cursor now. Each cuts the line with its *own* offset, where the units cancel out by construction — `${COMP_LINE:0:$COMP_POINT}`, `${BUFFER[1,CURSOR]}`, fish's already-cut `commandline -cp`, PowerShell's `Substring`. `--cursor` stays in the protocol for other callers and defaults to end-of-line, which is byte-exact for a line already cut.

**PowerShell cursor ignores StartOffset** — correct, and the same fix covers it: `$cursorPosition - $extent.StartOffset`, clamped both ways, so a pipeline or a `;` no longer shifts the offset past the end of the command.

Held by a test through the real shells: `COMP_LINE='ex ünicode here'` with `COMP_POINT=6` must reach the binary as `line=[ex üni]`, asserted in bash and zsh, with a stand-in that reports what it was handed rather than answering. Passed as a number, `6` would have landed mid-character.

**Executable names lose token boundaries** — every invocation and registration is quoted now, so a name needing quotes at least runs. One place cannot be: zsh's `#compdef` is a magic comment read by `compinit` before shell quoting exists, and there is nowhere to put a quote. That is a `debug_assert!` with the reason rather than a script that looks fine and does nothing.

jdx:

**Release builds accept invalid names** — right, and the suggestion is the right call: a `debug_assert!` removes itself from exactly the build where the broken script gets shipped. It is a hard `assert!` now. The name comes from the spec its author wrote rather than from anything a user typed, so a panic is a programming error surfacing where it can be fixed, and there is a `#[should_panic]` test on it.

**PowerShell dirs filter uses Test-Path** — also right. `CompletionText` is the text to *insert*, so a directory whose name needs quoting failed the test and vanished from a dirs-only completion. It now asks PowerShell what it already said: `ResultType -eq 'ProviderContainer'`, no second path lookup and no quoting to undo.

Worth flagging what is still unverified: nu and PowerShell are not installed on this machine, so their scripts are neither run nor parsed — bash, zsh and fish are all three. Both of these findings were in the two unrun scripts, which is roughly what one would expect.

jdx:
Three findings: two fixed, one that cannot be fixed on this side and is now written down where someone would look.

**Zsh fpath autoload name mismatch** — the important one, and correct. A zsh completion is normally installed by dropping `_mise` in `$fpath`, where compinit autoloads the file and calls the function *named after it*. The script defined `_usage_complete_mise` and registered with a top-level `compdef`: fine when sourced, and silently nothing when installed the usual way — the file loads, the function compinit wants is absent, and the user gets no completions and nothing to read about why. The function is now named after the binary (dash and all, which zsh accepts where an identifier would not) and the tail is the reference's `funcstack` branch. Both installs are tested by running zsh: sourcing must register, and calling `_ex` by name must complete.

**Quote characters break generated scripts** — right, and the same check should have caught it: whitespace was refused while an apostrophe went through, and every interpolation is single-quoted, so one would close the quoting and hand the rest of the line to the shell as syntax. A name now has to be one plain shell word.

**Nu drops paths with candidates** — real, and not fixable from here. nushell's external completer can be told candidates *or* paths: returning a list means "these are the completions", returning `null` means "you do this one", and there is no option beside either for "and files too". Where an answer has both, the candidates win — they are what this CLI knows, and a path is something the user can finish typing. I have written that into the script where the decision is, rather than leaving it to be discovered; if there is a newer nushell interface for merging the two I would rather use it, and I could not find one.

Standing caveat: nu and PowerShell are not installed on this machine, so their scripts are neither run nor parsed, while bash, zsh and fish are all three. Two of these three findings were in the two unrun scripts.

jdx:
Correct, and the sanitizing turned out to be unnecessary in the first place — bash, zsh and fish all accept `-`, `.` and `+` in a function name, which I checked by running them rather than assuming:

```
bash foo-bar: ok   bash foo+bar: ok   bash foo.bar: ok
fish foo-bar: ok   fish foo+bar: ok   fish foo.bar: ok
zsh  foo+bar: ok   zsh  foo.bar: ok
```

So four of the five scripts now use the binary's name verbatim and there is nothing left to collide. nushell binds a variable, where `-` reads as subtraction, so its name is escaped rather than flattened: `foo_x2dbar` against `foo_x2bbar`. The underscore is escaped too, so no name can spell an escape by hand and collide with the character it stands for — asserted directly.

PR #890, opened 2026-08-15
feat(argv): write the `run=` a declared completer answers

jdx:
A spec is what every other consumer reads — the usage CLI, another shell's
generator, a docs page — so a CLI whose completers live in Rust still has to say
what they are. It now writes them: `complete "tool" run="mise __complete_word__
--candidates tool"`, generated from the completer rather than declared beside it,
so there is one place a completer is said to exist and no second declaration to
keep in step.

And the binary answers exactly that command, so what the KDL promises is a thing
that works rather than a string that looks plausible. A name nothing declares
answers empty rather than failing: a script generated against a newer version of a
CLI is stale, and a stale script should complete nothing rather than print into
somebody's prompt.

Found by the name a spec uses — the lowercased argument name, or a flag's value
placeholder — which is the rule the reference resolves a `complete` block by, so
both sides agree about which completer a `run=` belongs to. Deduplicated, because
six of mise's commands take a `tool` and a spec declares one block for it.

The round trip is the test: emit the KDL, parse it back with usage-lib, and check
it reads a `run=` for that argument — then ask the binary the same question and get
the answer, filtered by what was typed.

---

---

> [!NOTE]
> **Medium Risk**
> Changes completion protocol and spec output that external generators read; behavior is heavily covered by roundtrip and dispatch tests, but wrong resolution order would break mise-style multi-command `tool` completers.
> 
> **Overview**
> Declared Rust completers are now **written into emitted KDL** as `complete` blocks whose `run=` invokes the same binary with `__complete_word__ --candidates <name> --line '…'`, instead of leaving completion as a second hand-maintained declaration beside the Rust callbacks.
> 
> The binary **implements that contract** via new `for_name` resolution (aligned with usage-lib: root, then command reached on the line, then tree fallback; cursor-aware when one name maps to both a positional and a flag value) and derive support for **`--candidates`**, returning filtered candidates or an empty list when the name is unknown. Spec emission also adds **`completers_on`**, per-command `write_completers`, and shell-safe line interpolation (apostrophe escaping; no `descriptions=#true` so colon-heavy values are not split).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 968892b5b8b0de98faf0de2c7c469b39a0cc5e10. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -1265,21 +1265,39 @@ mod tests {
         longs: &["only"],
         ..Flag::VALUE
     };
+    /// A flag whose completer is *not* the argument's, so which one answered is visible.
+    fn sources(_ctx: &CompleteCtx<'_>) -> Vec<Candidate<'static>> {
+        vec![Candidate::new("upstream")]
+    }
+    static SOURCE: Flag = Flag {
+        key: 17,
+        name: "source",
+        longs: &["source"],
+        ..Flag::VALUE
+    };
     static INSTALL: Command = Command {
         name: "install",
-        flags: &[&ONLY],
+        flags: &[&ONLY, &SOURCE],
         args: &[&TOOL_ARG],
         ..Command::EMPTY
     };
     static META_INSTALL: CommandMeta = CommandMeta {
         cmd: &INSTALL,
         about: Some("Install a tool"),
-        flags: &[FlagMeta {
-            flag: &ONLY,
-            help: Some("Just this one"),
-            complete: Some(tools),
-            ..FlagMeta::EMPTY
-        }],
+        flags: &[
+            FlagMeta {
+                flag: &ONLY,
+                help: Some("Just this one"),
+                complete: Some(tools),
+                ..FlagMeta::EMPTY
+            },
+            FlagMeta {
+                flag: &SOURCE,
+                help: Some("Where from"),
+                complete: Some(sources),
+                ..FlagMeta::EMPTY
+            },
+        ],
         args: &[ArgMeta {
             arg: &TOOL_ARG,
             help: Some("Which tool"),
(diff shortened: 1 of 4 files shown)


jdx:
Three findings, all correct, all fixed in the head commit.

**Completer context never reaches binary** — right, and it would have been a *wrong* answer rather than a missing one: a completer that reads an earlier flag would compute against nothing and answer confidently. The `run=` now carries the line, interpolated by whoever runs it — which is the reference, through tera, the same way its own `run="mise alias ls {{words[PREV]}}"` works.

**Deduplication conflates distinct callbacks** — also right. A `complete` block now sits inside the command that declares it, which is where usage-lib looks for one first ([`complete_word.rs:511`](https://github.com/jdx/usage/blob/main/cli/src/cli/complete_word.rs#L511) checks `cmd.complete` after `spec.complete`), and `for_name` resolves the same way: the command the line reached, then anywhere in the tree. The test has two siblings that both take a `TOOL` and mean the installed ones against the available ones.

**Description protocol mismatches output** — correct, and the fix is to stop claiming it. `descriptions=#true` tells the reference to read a description after an unescaped colon; what this answers with is values, so a value containing a colon would have been split in half — and mise's task names are full of them.

jdx:
Two more, one fixed and one that matches the reference.

**Completer lookup reverses precedence** — correct. usage-lib reads `spec.complete.get(name).or(cmd.complete.get(name))`, spec-level *before* command-level, and the root's completers are exactly what a spec writes at the top level. So it now resolves the root's own first, then the command the line reached, then tree order as a fallback — which is what keeps two siblings meaning different things by one name. There is a test pinning the root winning, since that order is otherwise invisible.

**Unescaped apostrophes break run line** — real, and shared with the reference: its own `run="mise alias ls {{words[PREV]}}"` is unquoted, so a word with a space breaks it there. Single-quoting is strictly better than that and still not airtight, and tera has no shell-quote filter to make it so. Left as is rather than half-fixed, and worth saying plainly: a completion line containing an apostrophe will misparse on the `run=` path. The native path — the generated scripts calling the binary directly — is unaffected, and that is the path an adopter's users are on.

jdx:
Checked, and this one does not hold: a dash-prefixed token never reaches the positional branch. The order in `candidates` is flags first —

```
`-`        → both forms
`--…`      → long flags
`-…`       → short flags
restart / awaiting value / positional
```

— so `--source=⌶` is answered by `long_flags` filtered on `--source=`, which matches nothing, and the argument's completer is never consulted. Same for an attached short value.

Pinned anyway, with a fixture whose flag and argument have *different* completers so that which one answered is visible at all — because your reasoning about the walk is right even though the conclusion is not: the word being completed is excluded, so the flag is not `awaiting_value` either, and the position would look like the argument's if that branch order ever changed.

PR #892, opened 2026-08-15
feat(derive): let a field name the function that completes it

jdx:
`#[usage(arg, name = "TASK", complete = tasks)]`, where `tasks` is handed its own
command's half-parsed struct and the context:

    fn tasks(partial: &<Tasks as CommandArgs>::Partial, ctx: &CompleteCtx) -> Vec<Candidate>

That is the thing a `run=` cannot do. `mise tasks ls --complete` is a fixed command
that reads whatever config the shell happens to be standing in, not the one being
typed about; a completer that can read `--file` off the line being completed
answers `tk tasks --file other.toml ⌶` with that file's tasks. The test asserts
exactly that, and asserts the same CLI still parses those fields, since it is one
declaration read by both.

The wrapper is what makes the table entry uniform: a completer in a `&'static`
table has one signature, and the useful signature is per-command, so the generated
wrapper reparses the words that command was given — which the position now reports —
and hands over the result.

A path rather than a string, so a name that does not resolve is a compile error
where it is written rather than a completion that silently answers nothing.

One thing worth knowing for anything else generated into a user's crate: the
metadata field cannot sit behind `#[cfg(feature = "complete")]`, because a `cfg` in
generated code is read against the *user's* features, not this crate's. It compiled
and quietly dropped the field, and the completer was never called. The types moved
beside the metadata they belong to instead.

---

---

> [!NOTE]
> **Medium Risk**
> Touches parser descent tracking and completion wiring used by generated code; behavior changes for dynamic completers and global/flattened flags, though covered by new conformance tests.
> 
> **Overview**
> **Field completers** — `#[usage(complete = my_fn)]` wires Rust callbacks into completion metadata (and emitted `run=` specs). Generated wrappers reparse the **owning command's** argv slice into that type's `Partial`, then call `fn(partial, ctx) -> Vec<Candidate>`, so answers can depend on earlier flags like `--file` on the same line—unlike a fixed external `run=` command.
> 
> **Richer completion context** — `CompleteCtx` moves to `spec` with `command_words` and `command_path`; `Parser`/`Position` record where each command on the path starts. Completers on **global** or **flattened** flags resolve the declaring command via `command_for` instead of only the deepest subcommand's words.
> 
> **Spec/metadata** — `FlagMeta`/`ArgMeta::complete` is always present (not `cfg(feature = "complete")`), avoiding silent drops in user crates. Completer paths use the same module rewriting as types (`crate::`, `self::`, etc.).
> 
> Conformance tests cover file-dependent tasks, global `--profile`/`--cd`, flattened globals, `--candidates` round-trips, and completer path spelling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df950e02ff2946688f04475cd603d529b0906348. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -51,6 +51,12 @@ pub struct Position<'t> {
     pub help_topic: bool,
     /// Where the command in scope began, as an index into the words `walk` was given.
     pub command_start: usize,
+    /// Every command the words passed through, and where each one's own words begin.
+    ///
+    /// The deepest is not always the one being asked about: a global flag is declared on an
+    /// ancestor, and a completer for its value wants the words *that* command was given — the
+    /// ones before the subcommand name included.
+    pub path: Vec<(&'t Command<'t>, usize)>,
     /// The flags a word here could name: this command's own, then any ancestor's globals.
     ///
     /// Taken from the parser rather than gathered again, so what is offered is what would be
@@ -90,6 +96,7 @@ pub fn walk<'t>(root: &'t Command<'t>, words: &[String]) -> Position<'t> {
                     flags_possible: false,
                     awaiting_value: None,
                     next_arg: None,
+                    path: Vec::new(),
                     separator_seen: false,
                     command_start: 0,
                     help_topic: true,
@@ -101,6 +108,7 @@ pub fn walk<'t>(root: &'t Command<'t>, words: &[String]) -> Position<'t> {
     }
 
     Position {
+        path: parser.command_path(),
         cmd: parser.command(),
         flags_possible: !parser.flags_stopped(),
         // A variadic flag still claiming words is standing in the same place a flag waiting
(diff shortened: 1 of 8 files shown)


jdx:
Correct — and the rule already existed in this file for a field's *type*, since the generated module sits one level below where the user wrote the path. The completer path was not going through it, so `crate::…`, `self::…`, `super::…` and a leading `::` all became something that does not resolve.

One shared rule now. Tested by writing a completer under `crate::` and another under `self::` and completing through both; the mutation that always prefixes `super::` fails to compile, which is the failure you described.

Worth saying that this is a bad class of bug to get wrong: it is not a compile error in this crate, it is one in the user's, at a line they did not write.

jdx:
Correct, and the fix turned up something worth recording.

A global flag is declared on an ancestor, so the words that reached the deepest command are not the words its owner was given: `g outer --cd here leaf --profile ⌶` puts `--cd here` before the subcommand's name, and the wrapper was reparsing the leaf's words against `outer`'s tables — finding nothing, and answering as though the flag had never been typed. The walk now records where each command on the path began, and a wrapper asks for *its own* command's words.

The part worth knowing: matching the owner by address does not work. A `Subcommands` variant builds its own table entry for the command it names — it can rename it, or give it aliases the type knows nothing about — so the entry the parse walked is a copy, and `ptr::eq` is false between them:

```
g     at 0  same_as_outer=false
outer at 1  same_as_outer=false      <- the one it is
leaf  at 4  same_as_outer=false
```

The key is a hash of the type it came from and survives the copying, so that is the identity two tables of one command agree on. My first version used addresses and the test caught it.

One limitation I hit and did not fix here: a completer on a *root* field does not compile, because the root does not implement `CommandArgs` and so has no partial to name in the signature. Giving the root that impl is a change of its own — it needs a `start()` the root module does not currently emit — and I would rather do it deliberately than smuggle it into this PR.

PR #895, opened 2026-08-15
feat(argv): say what went wrong the way clap says it

jdx:

---

> [!NOTE]
> **Low Risk**
> User-facing CLI output only, behind an opt-in feature; parser hot path unchanged and changes are covered by unit and gate tests.
> 
> **Overview**
> Adds an optional **`diagnostics`** feature (on top of **`spec`**) that turns parse **`Error`** values into user-facing messages shaped like **clap 4**: coloured or plain output, usage blocks only where clap would show them, and a **`--help`** footer.
> 
> **`diagnostic::render`** walks argv to find the failing command, formats each **`Error`** variant, and builds the **Usage:** line from the same **`help::usage_line`** as **`--help`** (not clap’s wording). **`help`** now exposes **`arg_usage`** / **`flag_spelling`** internally so errors spell flags and positionals consistently with help.
> 
> **`benches/gate/tests/errors.rs`** compares first-line error text and usage-block presence against clap on real **mise** command lines; **gate** and **conformance** enable **`diagnostics`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a40c8f71c037bfff5b5864ea88e19c56a2df8b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added opt-in diagnostic rendering with plain-text and ANSI-coloured output.
  * Error messages now provide clap-compatible wording, command context, and usage information.
  * Help requests produce no diagnostic output, while value-level errors omit unnecessary usage details.

* **Tests**
  * Added compatibility coverage for invalid commands, subcommands, missing values, required arguments, usage lines, colour handling, and help requests.

### DIFF
--- a/argv/Cargo.toml
+++ b/argv/Cargo.toml
@@ -22,6 +22,9 @@ spec = []
 # separate from `spec` because a CLI can want one without the other: completion needs
 # to split a line whether or not the binary can also emit its own spec.
 complete = ["spec"]
+# What a user reads when a command line does not parse. Off by default like the rest: a CLI
+# that renders its own errors should not carry these strings, and the parser stays a parser.
+diagnostics = ["spec"]
 
 [package.metadata.release]
 shared-version = true
--- a/argv/src/diagnostic.rs
+++ b/argv/src/diagnostic.rs
@@ -95,20 +95,30 @@ impl Style {
     }
 }
 
-/// A name as a usage line writes it: `<TOOL>` when it must be filled, `[TOOL]` when it need not.
+/// A name as a usage line writes it: `<TOOL>`, `[TOOL]…`, `--jobs`.
 ///
-/// The error carries the spec's name for a thing; a user reads the form the help shows. Looked up
-/// on the command rather than guessed, and left alone when it names a flag — a flag already reads
-/// as itself.
+/// The error carries the spec's name for a thing; a user reads the form the help shows. Both come
+/// from `help` rather than being decided again here — an error and the page above it describing
+/// one argument differently is the confusing kind of inconsistency, and rewriting the rule is how
+/// that happens. A flag is spelled with its dashes, which is the whole of what `--jobs` versus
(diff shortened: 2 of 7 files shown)


jdx:
All four addressed in 9a40c8f. They were one root cause: the error text was deciding for itself how to spell an argument instead of asking the module that already knows.

**`shown()` searched only positionals** — a flag fell through to its bare spec name, so a user read `jobs` for something they have to type as `--jobs`.

**It reimplemented help's rule** rather than sharing it, and so had already drifted: `demanded` treats an argument with a default as optional, and this printed `<TOOL>` for one the help writes `[TOOL]`; a variadic lost its `…`; an argument needing a separator lost its `[-- …]`. `help::arg_usage` and `help::flag_spelling` are now `pub(crate)` and there is one rule for both. An error and the page above it describing the same argument differently is the confusing kind of inconsistency.

**`InvalidValue`, `VarTooFew` and `VarTooMany` printed the raw name** while two other variants did not — so the same argument could appear as `TOOL` in one message and `<TOOL>` in another. All variants go through `shown()` now.

**`value_bound_to` returned the last bound value.** A variadic may be given several and the check stops at the first that is not allowed, so `use node fsh zsh` reported `invalid value 'zsh'` — naming a value that was perfectly good and leaving the wrong one unmentioned. It now takes the choices and returns the first value not among them.

Each fix is load-bearing, checked by mutation: reverting `shown()` to the spec name fails `a_name_is_spelled_the_way_the_help_spells_it`; dropping the flag branch fails it too; restoring the raw name in the three variants fails it; and returning the last bound value fails `the_value_named_is_the_one_that_was_refused`.

Thanks to both bots — the first two were greptile's, the last two Bugbot's.

PR #897, opened 2026-08-15
feat(argv): suggest what was probably meant

jdx:
`--fore` → "a similar argument exists: '--force'". Jaro-Winkler above 0.7, which is
clap's rule rather than a rule of ours, so the two suggest in the same cases and
suggest the same thing. Written out rather than depended on — this crate takes no
dependencies, and the algorithm is thirty lines.

Three things measured from clap rather than assumed, each of which I had wrong
first:

Names are scored *without* their dashes. Every flag begins `--`, and Jaro-Winkler
rewards a shared prefix, so comparing the dashed forms made `--fore` look similar to
`--quiet`. The tests catch that now.

Every candidate over the bar is listed, not the best one: `mise config lss` really
is close to both `ls` and `list`, and clap says so — "some similar subcommands
exist: 'list', 'ls'". The plural is the singular with an `s`.

And they come out in *ascending* score, so the closest match is last. That is what
clap does; it reads oddly, and I have left it matching rather than fixed on one
side, since the point of this module is that an adopter's users see no change. Worth
undoing in both places if you agree it is a bug.

Suggestions come from what the parser would have accepted at that command — its own
flags and any ancestor's globals — so a tip is always something that works.

Also recorded, in the gate tests: mise's spec leaves `unknown_flags` at its default,
so `mise use --globa` is an error under clap and a *tool named `--globa`* under this
parser. That is a decision for the adopting CLI rather than a bug here — declaring
`unknown_flags=error` restores both the refusal and these suggestions — but it is
the reason no flag typo appears among the parity cases, and it should not be
discovered late.

---

---

> [!NOTE]
> **Medium Risk**
> Changes user-visible CLI error text and command-path resolution for diagnostics; behavior is heavily tested against clap and shared-subcommand fixtures, but adopters with non-default `unknown_flags` still won’t see flag typo tips until they opt into error-on-unknown-flag.
> 
> **Overview**
> Adds **clap-aligned “tip:” lines** to parse error output when the user’s token is close to a real flag, subcommand, or choice value. Similarity uses an inlined **Jaro** scorer (not Jaro-Winkler), a **> 0.7** threshold, bare flag names for matching, `=` stripped from long flags, negation spellings included, and the same multi-match / ascending-score ordering clap uses.
> 
> Diagnostics now follow the **argv route** (`path_taken` + per-parent `resolve`) instead of looking up metadata by final `Command` pointer, so shared subcommands under different parents get the right usage line and global-flag suggestions. Flag candidates are limited to **flags in scope** on that chain (own flags plus ancestor globals only).
> 
> Other message tweaks: dash-prefixed unexpected tokens on commands with subcommands read as **unexpected arguments** with flag tips; invalid-choice errors can suggest a nearby value; conflicting / double-dash errors spell names via **help** formatting. Gate and conformance tests compare tips and routing behavior to clap and cover the shared-subcommand case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db1ac64f0548dc7022e3d6bc003b7e98e92b93cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line error messages with more accurate usage and command context.
  * Suggestions now correctly account for local, global, sibling, and parent-command flags.
  * Enhanced handling of unknown arguments, attached values, negated flags, subcommands, and double-dash requirements.
  * Invalid option values now provide helpful suggestions for close matches.
  * Standardized flag names and formatting across conflict and validation errors.

* **Tests**
  * Expanded coverage for diagnostics, suggestions, subcommands, shared command definitions, and parser compatibility.

### DIFF
--- a/argv/src/diagnostic.rs
+++ b/argv/src/diagnostic.rs
@@ -115,6 +115,16 @@ fn flag_named(token: &str) -> &str {
     }
 }
 
+/// Every long spelling a flag answers to, its negation included.
+///
+/// The parser takes `--no-color` through `find_negation`, and the completions offer it, so a
+/// suggestion that leaves it out is the odd one — a near miss of a name that works gets silence.
+/// clap has no separate notion of a negation, so the two forms are two arguments there and it
+/// suggests either; matching that is the point.
+fn long_spellings<'a>(meta: &'a crate::spec::FlagMeta<'a>) -> impl Iterator<Item = &'a str> {
+    meta.flag.longs.iter().copied().chain(meta.flag.negate)
+}
+
 /// Every flag a word at this command could have named: its own, then any ancestor's globals.
 ///
 /// The same set the parser would have accepted, which is what makes a suggestion one that works.
@@ -410,9 +420,7 @@ pub fn render(
             // `--fore` came out similar to `--quiet`, which it is not. clap compares the bare
             // names for the same reason.
             let bare = typed.trim_start_matches('-');
-            let names: Vec<&str> = flags_in_scope(chain)
-                .flat_map(|meta| meta.flag.longs.iter().copied())
-                .collect();
+            let names: Vec<&str> = flags_in_scope(chain).flat_map(long_spellings).collect();
             let near: Vec<String> = nearest(bare, names.into_iter())
                 .into_iter()
                 .map(|name| format!("--{name}"))
@@ -442,9 +450,7 @@ pub fn render(
                     style.invalid(named)
                 );
                 let bare = named.trim_start_matches('-');
-                let names: Vec<&str> = flags_in_scope(chain)
-                    .flat_map(|meta| meta.flag.longs.iter().copied())
-                    .collect();
+                let names: Vec<&str> = flags_in_scope(chain).flat_map(long_spellings).collect();
                 let near: Vec<String> = nearest(bare, names.into_iter())
                     .into_iter()
                     .map(|name| format!("--{name}"))
@@ -645,6 +651,9 @@ mod tests {
         name: "force",
         longs: &["force"],
         shorts: b"f",
+        // A negation, because a flag's spellings are not only its `longs` and the parser takes
+        // this one — so a suggestion that cannot offer it is offering less than the CLI accepts.
+        negate: Some("no-force"),
         ..Flag::BOOL
     };
     static JOBS: Flag = Flag {
@@ -991,6 +1000,34 @@ mod tests {
         let message = rendered(&["use"], Error::UnknownFlag { token: b"-j=4" });
         assert!(
             message.starts_with("error: unexpected argument '-j=4' found"),
+            "{message}"
+        );
+    }
+
+    #[test]
+    fn a_negation_is_suggested_like_any_other_spelling() {
+        // `--no-force` is a name the parser accepts, through `find_negation`, and one the
+        // completions already offer. Scoring only `longs` left it out, so a near miss of a name
+        // that works got silence — and clap, which has no separate notion of a negation and
+        // sees two arguments, suggests it. Measured:
+        //
+        //     error: unexpected argument '--no-colr' found
+        //       tip: a similar argument exists: '--no-color'
+        let message = rendered(
+            &["use"],
+            Error::UnknownFlag {
+                token: b"--no-forc",
+            },
+        );
+        assert!(
+            message.contains("tip: a similar argument exists: '--no-force'"),
+            "{message}"
+        );
+
+        // And the plain form is still found, which is the half that already worked.
+        let message = rendered(&["use"], Error::UnknownFlag { token: b"--fore" });
+        assert!(
+            message.contains("tip: a similar argument exists: '--force'"),
             "{message}"
         );
     }
--- a/argv/src/diagnostic.rs
+++ b/argv/src/diagnostic.rs
@@ -120,7 +120,7 @@ fn flag_named(token: &str) -> &str {
 /// The same set the parser would have accepted, which is what makes a suggestion one that works.
 fn flags_in_scope<'a, 'c>(
     chain: &'c [&'a CommandMeta<'a>],
-) -> impl Iterator<Item = &'a crate::spec::FlagMeta<'a>> + use<'a, 'c> {
+) -> impl Iterator<Item = &'a crate::spec::FlagMeta<'a>> + 'c {
     // The command's own flags, and from each ancestor only what it declared global — the rule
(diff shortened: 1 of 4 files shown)


jdx:
Both correct, and the first one corrects something I told jdx.

**Wrong algorithm versus clap parity** — right, and clap says why in its own source:

```rust
// GH #4660: using `jaro` because `jaro_winkler` implementation in `strsim-rs` is wrong
let confidence = strsim::jaro(v, pv.as_ref());
if confidence > 0.7 {
```

Winkler's prefix bonus moves the bar — some words clear 0.7 with it and not without, and the ranking changes with it — so a module whose premise is "an adopter's users see the same tips" has to use the same algorithm. Switched to plain Jaro, with clap's comment quoted where the choice is made and the bonus three lines away if it is ever wanted on both sides. Worth noting the reason clap avoided Winkler does not apply to us — a wrong implementation in `strsim` — but parity does.

**Unrelated globals enter flag scope** — also right, including the diagnosis: `inside` was always false, so every node on the walk was treated as an ancestor. It follows the chain to the command now — own flags, plus from each real ancestor only what that ancestor declared global.

The fixture could not see either fix, which is worth recording: its sibling command was declared *after* the target, so the walk never passed through it, and every ancestor flag in it happened to be global. The sibling is declared first now and the root has a non-global flag of its own; both mutations survive against the old fixture and fail against this one.

jdx:
Real, and worse than the report says — fixed in 4d0afe4.

The parser splits on the `=` before looking a long flag's name up, so `--fore=1` names `--fore`; the error was quoting `--fore=1`, which nobody typed. And the tip is the half that would have gone quietly: `fore=1` against `force` falls under the 0.7 bar, so the suggestion vanished precisely in the attached form — which is the form you use *when the flag takes a value*, i.e. the case where mistyping it is most likely.

I ran clap 4 rather than trusting my memory of it. It says both:

```
error: unexpected argument '--fore' found

  tip: a similar argument exists: '--force'
```

so the fix restores parity on the quoted token as well as on the tip.

Long flags only, deliberately: a short cluster is refused whole — `-xy` is not `-x` with a `y` attached — and clap keeps the `=` in a short flag's value. `-j=4` still reads back as `-j=4`, asserted.

The same rule now applies to the `UnexpectedArg` arm, which is where a dash-word lands on a command that takes unknown flags as values — same spelling mistake, same message.

Mutations: leaving the value on fails the new test; truncating short flags too fails it as well.

jdx:
Confirmed, and worse than the report — fixed in 15b3202.

The premise checks out: two parents mounting one `Subcommands` type splice the same `&'static [Command]`, so `alpha shared` and `beta shared` are one command at one address. `ptr::eq` cannot tell them apart, and the tree search returns whichever comes first.

What that actually produced for `ex beta shared --betaglobl`:

```
error: unexpected argument '--betaglobl' found

Usage: ex alpha shared [--thing]
```

So not only the suggestions — **the usage line named a different command than the one on the command line**. And `--betaglobal`, the flag the user meant and the only one the parser would have taken there, went unmentioned, because alpha's globals were the ones in scope.

The command an error ends on does not identify itself; the route to it does. `path_taken` now collects each `Command` event, and `resolve` walks the metadata down that route, matching each step among *that command's* children — a parent's own child list is unambiguous even when the child is shared. `flags_in_scope` takes the resolved chain instead of searching for it.

Mutation: restoring the depth-first search fails the new test on the usage line first (`Usage: ex alpha shared` for a `beta` command line). Truncating the recorded path to its last command fails two.

**One thing I did not fix here, and it is not new:** `help::find` resolves the same way, so `ex beta shared --help` also prints `Usage: ex alpha shared`, and completions resolve a command by the same search in three places. Both predate this stack and neither is reachable from the diagnostics path, so folding them in would mean threading a route through three more call sites in two modules under a diagnostics PR. Worth its own change — happy to do it next if you'd rather it land together.

jdx:
Four more addressed in d2a8c4c, and one moved to its own PR.

**MSRV.** `use<'a, 'c>` is Rust 1.82 and `argv/Cargo.toml` declares 1.80. Written the way 1.80 spells it. Worth noting *why* it got in: nothing in CI builds at the declared MSRV, so a declaration nobody can honour compiles here — the rust-version is currently a claim rather than a check.

**`flag_spelling` under `--features spec`.** Confirmed: `warning: function 'flag_spelling' is never used`, and this workspace makes warnings errors. Gated with the code that calls it, and I checked all eight feature combinations rather than just the reported one — the missing check was the real gap, the gate is one line.

**`ConflictingFlags` and `ArgRequiresDoubleDash`.** Right, and this is the same finding #895 was about — these two were the ones I missed, sitting directly between two variants that *do* go through `shown`, so one argument could appear two ways in two messages from the same command. I ran clap 4 to confirm the target: `error: the argument '--force' cannot be used with '--jobs <JOBS>'` — dashes and value name, which is what `help::flag_spelling` produces.

**The stale Jaro-Winkler comment** in the gate test. Fixed, with the reason attached: clap uses plain `jaro` because `strsim`'s `jaro_winkler` is wrong (clap GH #4660). A comment a reader trusts that says the opposite of the code is worse than no comment.

Mutations: reverting each of the two name fixes fails the extended spelling test; ungating `flag_spelling` reproduces the warning.

**Moved out:** the `settings_layer`/`settings_given` finding on `derive/src/codegen.rs` is in already-merged code, not this stack, so it is #904 rather than folded in here. Confirmed first — `#[usage(bin = "ex", settings)]` on a root with nothing to collect fails with `cannot find function settings_given` pointing at `#[derive(Cli)]`. It is refused at the attribute now.

Rebased onto `main` (which now has #895).

jdx:
Fixed in 9ce49ae.

Confirmed: `negate` is a field of its own, not part of `longs`, so a flag's negation was never among the candidates — while `find_negation` accepts it and the completions already offer it. Silence where a working name is right there is the one thing a tip should never be.

Ran clap 4 to check the target rather than reasoning about it. clap has no separate notion of a negation, so `--color` and `--no-color` are two arguments there and it suggests either:

```
error: unexpected argument '--no-colr' found

  tip: a similar argument exists: '--no-color'
```

The fixture's `--force` now carries `negate: Some("no-force")`, and `--no-forc` suggests it. Mutation: dropping the negation from the candidate set fails the new test — and the first attempt at that mutation *survived*, because `cargo fmt` had collapsed the chain onto one line and my patch silently matched nothing. Re-ran it against the formatted source; it fails.

Rebased onto `main` (now carrying #899).

PR #900, opened 2026-08-15
feat(derive): say how an argument relates to `--`, all four ways

jdx:
The parser has understood four `double_dash` modes since it was written. The derive could say one, so it modelled the question as a bool — `double_dash_required` — and a CLI wanting either of the other two had to reach into the generated spec afterwards.

mise does exactly that. Its root is two things at once: where its global flags live, and the shorthand for `mise run`. `mise --globa build` should be a misspelled flag; `mise build --globa` should hand `--globa` to the task. clap expresses the second with `allow_hyphen_values`, which says nothing about *where* on the line a dash-word may appear, so the whole root goes permissive and the typo is silently accepted as a task argument. `double_dash="automatic"` says what clap cannot: flag parsing ends when *that argument takes its first value*. mise's spec carries it on three arguments — `run`, `asdf`, `tool-stub` — and mise patches it in by hand.

`preserve` comes along because it is the same question from the other end: `required` consumes the `--`, `preserve` keeps it as a value, and the derive could ask neither.

**The rule worth stating**, because it is easy to get backwards now that all four can be written: only `required` exempts an argument from the can-never-be-filled check behind an unbounded variadic. `automatic` stops *flags*, which is a different thing, and `preserve` changes what a `--` means without ending anything. Both are refused there exactly as a plain argument is, per mode.

The shadow generator now reports each mode against the dialect that cannot hold it — the usage shadow drops none, and the clap shadow reports the three `automatic` declarations as clap's gap, which is the honest place for them.

## Verification

Four mutations, each failing a test that named the rule:

| mutation | result |
|---|---|
| `automatic` emitted as `Optional` | FAILED (2 tests) |
| `preserve` emitted as `Optional` | FAILED (2 tests) |
| `automatic`/`preserve` exempt an argument from the variadic rule | FAILED |
| an unknown mode silently reads as `optional` | FAILED |

Workspace suite green, clippy clean with no exclusions.

## Summary by CodeRabbit

- **New Features**
  - Added expanded support for positional-argument `--` behavior, including optional, required, preserved, and automatic modes.
  - Automatic mode now hands off parsing after positional arguments begin, while preserve mode retains the separator.

- **Bug Fixes**
  - Improved variadic-argument handling when a `--` separator is present, including validation of unreachable arguments.

- **Documentation**
  - Updated usage documentation to describe all supported separator behaviors.

- **Tests**
  - Added conformance coverage for serialization, parsing, separator preservation, and variadic arguments.

---

> [!NOTE]
> **Medium Risk**
> Changes CLI parsing and compile-time validation for common mise-style wrappers; mistakes would surface as new derive errors or altered argv splitting, but behavior is heavily tested and limited to declarative metadata.
> 
> **Overview**
> **`#[usage(double_dash = …)]` now covers all four spec modes** (`optional`, `required`, `preserve`, `automatic`) instead of a single `required` boolean, and codegen emits the matching `DoubleDash` variant into specs so consumers like `usage complete-word` see the real mode without post-hoc patches.
> 
> **Variadic layout rules are tightened:** only `required` can unlock a positional behind an unbounded variadic; `preserve` variadics are treated as consuming the separator (so a later `--`-only arg is unreachable), with matching derive-time errors and `unfillable_arg` / debug_assert messaging for `#[usage(flatten)]` trees.
> 
> **mise shadow generated structs** set `double_dash = "automatic"` on `AsdfArgs`, `ToolStubArgs`, and `RunArgs` `ARGS` vectors; **xtask** emits all four modes into the usage shadow while still noting clap’s `last`-only gap for `automatic`/`preserve`.
> 
> **Conformance tests** in `double_dash.rs` cover KDL round-trip, `automatic` flag handoff, `preserve` keeping `--`, and that `automatic` does not stop an unbounded variadic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76ece68368ead3e8b6d375b2cc6f9c09f1ced433. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -85,6 +85,10 @@ fn duplicate_flag_form(cmd: &Command<'_>) -> Option<std::string::String> {
 /// because the separator ends the collecting, and a `var_max`, because a bounded variadic hands
 /// over the words past its bound. mise relies on the first on `run`, `exec` and `git`.
 ///
+/// The separator is good for one argument and only where it is still a separator. A variadic
+/// that already claimed it has spent it, and one declaring `preserve` takes it as a *value* —
+/// so in either case the argument waiting for a `--` waits forever.
+///
 /// `#[derive(Args)]` applies that rule to one struct's own arguments. It cannot apply it across
 /// a `#[usage(flatten)]`: the variadic may be on one side of the boundary and the argument that
 /// follows it on the other, and neither expansion can see the other's fields. Checked here for
@@ -98,7 +102,14 @@ fn unfillable_arg<'a>(cmd: &Command<'a>) -> Option<&'a str> {
         // one separator exists, so nothing can follow *that*.
         let stopped_by_separator = arg.double_dash == DoubleDash::Required;
         if let Some(before) = variadic {
-            if !stopped_by_separator || before.double_dash == DoubleDash::Required {
+            let separator_is_gone = matches!(
+                before.double_dash,
+                // Spent on the variadic itself.
+                DoubleDash::Required
+                // Or never a separator at all, because the variadic keeps it as a value.
+                    | DoubleDash::Preserve
+            );
+            if !stopped_by_separator || separator_is_gone {
                 return Some(arg.name);
             }
         }
@@ -573,8 +584,10 @@ impl Spec<'_> {
             "no word could ever reach the argument {:?}, because an unbounded variadic before \
              it takes every remaining one. With `flatten` this is the arrangement neither \
              expansion can see: the variadic and the argument after it were declared in \
-             different structs. Give the variadic a `var_max`, or make the later argument \
-             fillable only after a `--`.",
+             different structs. Give the variadic a `var_max` — or, if the variadic leaves the \
+             `--` alone, make the later argument fillable only after one. A variadic that \
+             already requires a separator has spent it, and one declaring `preserve` takes it \
+             as a value, so neither can be stopped that way.",
             unfillable_arg(self.root.cmd)
         );
         let mut out = String::new();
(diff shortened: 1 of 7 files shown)


jdx:
Real — fixed in 8d1064f. Reproduced before fixing:

```rust
#[usage(arg, double_dash = "preserve")]  rest: Vec<String>,
#[usage(arg, double_dash = "required")]  after: Vec<String>,
```

`ex a -- b` → `rest=["a", "--", "b"] after=[]`. It compiled, and `after` could never hold anything for any command line.

The exemption it slipped through is the one that makes mise's `run` legal: an argument fillable only after a `--` is reachable behind an unbounded variadic, because the separator ends the collecting. `preserve` takes the `--` as one of its own values, so it ends nothing and the exemption is false. Refused now with a message that says which mode caused it and what to do; a `var_max` puts the exemption back, since a bounded variadic hands over whether a separator arrives or not.

Only reachable since `preserve` became expressible in this commit, so nothing shipped with it.

Mutations: not marking a `preserve` variadic as separator-eating fails the new test; dropping the `var_max` exemption fails two.

jdx:
Real, and a good catch on the half I missed — fixed in 9fe4769.

The derive check I added closes the layout within one struct. `unfillable_arg` in `argv/src/spec.rs` is the counterpart that sees the *assembled* tables, which is where a `flatten` is caught, and it still read `required` after any non-`required` unbounded variadic as reachable. So this compiled and ran:

```rust
#[derive(Args)] struct Front { #[usage(arg, double_dash = "preserve")] rest: Vec<String> }
#[derive(Cli)]  struct Ex { #[usage(flatten)] front: Front,
                            #[usage(arg, double_dash = "required")] after: Vec<String> }
```

`ex a -- b` → `rest=["a", "--", "b"] after=[]`.

One rule now in both places: the separator is good for one argument and only where it is still a separator. A variadic that already requires one has spent it; one declaring `preserve` takes it as a value.

While confirming the fix I found the message had become self-contradictory — it advised making the later argument fillable only after a `--`, which is exactly what that argument already declares and exactly what does not work here. It now says what stops each kind, since advice that repeats what the author wrote is worse than none.

Mutations: dropping `Preserve` from the spent-separator set fails the new fixture; removing the `var_max` exemption fails it too.

PR #902, opened 2026-08-15
feat(derive): a default a collecting field can hold

jdx:
The last thing mise's 211-command spec could say that the derive could not. `mise watch --fs-events` starts out filtering to five event kinds; the derive refused a default on a `Vec` with *"not applied yet, so it would be documented and then ignored"* — true then, not now.

`default` becomes a list, because the spec's is. A `Vec` may be given several and starts out holding all of them, in the order written; every other shape has one place to put a value, so a second `default` is refused rather than silently kept — a declaration the author cannot see being ignored is what the old rule was avoiding.

**Replaced, not added to.** A default says what a flag means when nobody said anything, so `--fs-events access` means one event and not four. That matters twice: where the flag was given, and where it was *displaced* by an overriding flag and has to read as untouched — which is the path that actually reaches the code, since the declared-default path is guarded on `__given_*` and never sees a filled collection. Both are tested; the second was found by mutation, when clearing turned out to be free.

Every default is checked against the choices, not the first. A collection's second is as unusable as its first if the choices do not allow it.

## The milestone

```
generated benches/shadows/mise/src/lib.rs (211 commands, 711 flags, 128 args)
  nothing dropped: usage expressed the whole spec
```

The clap shadow still drops nine declarations — `default_subcommand`, three `double_dash = "automatic"`, two mounts, two restart tokens, and a flag's second default. That is the same list mise patches into its own spec by hand today.

## Verification

| mutation | result |
|---|---|
| a collection's defaults declared but not seeded | FAILED |
| only the first default seeded | FAILED |
| defaults appended rather than replacing | FAILED (after the override test was added — see below) |
| only the first default checked against the choices | FAILED |

The third mutation initially **survived**: the clear is unreachable on the declared-default path, so only an override on a collecting flag with defaults can see it, and nothing covered that. `a_displaced_collecting_flag_goes_back_to_its_defaults` covers it now, and the mutation fails.

Workspace suite green, clippy clean with no exclusions.

---

> [!NOTE]
> **Medium Risk**
> Changes core CLI parsing and default/env/override semantics for all fields using `default`, though behavior is heavily tested; incorrect replacement vs append logic would change real CLIs like `mise watch --fs-events`.
> 
> **Overview**
> **Collecting fields (`Vec`) can declare several `default` values** (repeated `default = "…"` / KDL `default { … }`), matching the usage spec. Non-collection fields still allow at most one default; extra defaults are a compile error.
> 
> **Runtime behavior:** absent argv, the collection starts with all declared defaults in order. Explicit `--flag` values **replace** the whole collection (not merge). Env vars on collecting fields **clear then set** one value, replacing defaults. `Option<Vec<_>>` with defaults is always `Some(…)` when defaults exist. Flags displaced by `overrides` reset to the declared default list via `reset_to_default`.
> 
> **Emitted metadata** (KDL/spec, shadow codegen) carries every default for variadic flags; `xtask` shadow translation no longer skips defaults on collecting flags.
> 
> Conformance tests cover parsing, spec emission, override displacement, env override, and `Option<Vec>` behavior. Benchmark shadows (`mise`) update `--fs-events` to the new multi-default form.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66e0643f3857bf0ebd0ec2c7bc5584fb1833b31c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Collection-based CLI options can now define and apply multiple default values.
  * Optional collection options are populated from declared defaults.
  * Environment-provided values replace collection defaults consistently.
  * Filesystem event options now support clearer multi-value default declarations.

* **Bug Fixes**
  * Defaults are correctly restored after overrides and preserve their declared order.
  * Invalid or excessive defaults now produce clearer validation errors.

* **Documentation**
  * Updated CLI option documentation to explain multiple defaults for collection options.

### DIFF
--- a/benches/mise.usage.kdl
+++ b/benches/mise.usage.kdl
@@ -5386,7 +5386,7 @@ This can also be used via the $WATCHEXEC_IGNORE_FILES environment variable.
 """#
         arg <PATH>
     }
-    flag --fs-events help="Filesystem events to filter to" var=#true default=create,remove,rename,modify,metadata {
+    flag --fs-events help="Filesystem events to filter to" var=#true {
         long_help #"""
 Filesystem events to filter to
 
@@ -5394,6 +5394,13 @@ This is a quick filter to only emit events for the given types of filesystem cha
 
 This may apply filtering at the kernel level when possible, which can be more efficient, but may be more confusing when reading the logs.
 """#
+        default {
+            create
+            remove
+            rename
+            modify
+            metadata
+        }
         arg <EVENTS> {
             choices access create remove rename modify metadata
         }
--- a/benches/shadows/mise-clap/src/lib.rs
+++ b/benches/shadows/mise-clap/src/lib.rs
@@ -5356,7 +5356,7 @@ pub struct WatchArgs {
     /// This is a quick filter to only emit events for the given types of filesystem changes. Choose from 'access', 'create', 'remove', 'rename', 'modify', 'metadata'. Multiple types can be given by repeating the option or by separating them with commas. By default, this is all types except for 'access'.
     ///
     /// This may apply filtering at the kernel level when possible, which can be more efficient, but may be more confusing when reading the logs.
-    #[arg(long = "fs-events", value_name = "EVENTS", value_parser = ::clap::builder::PossibleValuesParser::new(["access", "create", "remove", "rename", "modify", "metadata"]), default_value = "create,remove,rename,modify,metadata")]
+    #[arg(long = "fs-events", value_name = "EVENTS", value_parser = ::clap::builder::PossibleValuesParser::new(["access", "create", "remove", "rename", "modify", "metadata"]), default_value = "create")]
     pub fs_events: Vec<String>,
     /// Don't emit fs events for metadata changes
     ///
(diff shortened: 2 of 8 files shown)


jdx:
Both real — fixed in 273c9da. Reproduced before fixing, one program:

```rust
#[usage(long, var, default = "a", default = "b", env = "EX_EV")]  ev: Vec<String>,
#[usage(long, var, default = "x", default = "y")]                 opt: Option<Vec<String>>,
```

with `EX_EV=z` and an empty command line:

```
before:  ev=["a", "b", "z"]   opt=None
after:   ev=["z"]             opt=Some(["x", "y"])
```

**Bugbot's** — every other shape assigns, so the environment overrides a default; a collection pushed, so it held both, and the two extras are precisely the values the user's variable was chosen instead of. Nothing else can be in the collection at that point (argv sets `__given_*` and the fallback is guarded on it), so clearing first is exactly the assignment the other shapes already do.

**greptile's** — `None` means never given, a default is a value, and seeding one deliberately leaves `__given_*` alone so the environment can still replace it. So the answer has to come from the declaration rather than the partial; it came from the partial, and every declared default was discarded on the way out.

Both are consequences of this PR letting a `Vec` carry a default at all, so neither has shipped.

Mutations: restoring the push fails the env test; ignoring `#defaulted` fails the optional-collection test; making it unconditionally `Some` fails it too — a field that declares no default still tells "never given" from "given nothing".

PR #904, opened 2026-08-16
fix(derive): name the mistake when `settings` has nothing to collect

jdx:
`#[usage(settings)]` says this CLI resolves settings whose flags are declared *elsewhere*. With no elsewhere — nothing binding a setting, no flattened group, no subcommand — the layer was emitted while the function it calls was not:

```rust
#[derive(Cli)]
#[usage(bin = "ex", settings)]
struct Ex {
    #[usage(long)]
    plain: bool,
}
```

```
error[E0425]: cannot find function `settings_given` in this scope
 --> src/main.rs:4:10
  |
4 | #[derive(Cli)]
  |          ^^^
```

An unresolved name inside generated code, pointing at the derive, naming neither the attribute that caused it nor what to do about it.

Refused where it is written now, beside the existing check for the same attribute in the wrong *place*. Any of the three is enough to accept it, since each is a way for a flag to be somewhere this struct does not declare it.

Reported by CodeRabbit while reviewing #897; it is in merged code rather than that stack, so it is its own PR against `main`.

## Verification

| mutation | result |
|---|---|
| the refusal dropped | FAILED |
| only a bound field counts (a group or subcommand no longer enough) | FAILED |

Reproduced the E0425 first, then confirmed the same declaration is refused with a message naming the attribute. Workspace suite green, clippy clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Root-level `#[usage(settings)]` is now rejected when no settings source is available.
  * Valid configurations continue to be accepted when settings are provided through a field, flattened group, or subcommand.

---

> [!NOTE]
> **Low Risk**
> Compile-time derive validation and diagnostic span changes only; no runtime CLI behavior.
> 
> **Overview**
> Root `#[usage(settings)]` with no bound `setting`, flattened group, or subcommand used to compile and then fail with **`cannot find function settings_given`** on the derive line. **`check_position`** now rejects that case with a message that names the attribute and what to add or drop.
> 
> **`attr_span`** records where `#[usage(...)]` sits on the struct, and a **`misplaced`** helper uses it for all position-rule errors (`completion`, `settings`, `default_subcommand`, `mount`, `restart_token`, etc.) so diagnostics underline the attribute instead of the struct name.
> 
> Tests cover the new refusal and that a field with `setting`, a flatten, or a subcommand still passes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dde9e1d6ceb16665530ac16b3f3ea2e2760f9371. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/derive/src/model.rs
+++ b/derive/src/model.rs
@@ -35,6 +35,12 @@ pub struct Cli {
     /// CLI with a subcommand depend on `usage-config`. A root that binds a setting of its own has
     /// already said it, and does not need this.
     pub settings: bool,
+    /// Where `#[usage(...)]` was written on the struct, when it was.
+    ///
+    /// Every position rule in [`Cli::check_position`] is about an attribute in the wrong place,
+    /// so every one of them should point at the attribute. Spanning the struct's *name* put the
+    /// underline one line below the thing that was wrong.
+    pub attr_span: Option<proc_macro2::Span>,
     /// From the struct's doc comment: first paragraph, and the whole thing.
     pub about: Option<String>,
     pub long_about: Option<String>,
@@ -245,6 +251,11 @@ impl Cli {
             bin: None,
             completion: false,
             settings: false,
+            attr_span: input
+                .attrs
+                .iter()
+                .find(|a| a.path().is_ident("usage"))
+                .map(|a| a.path().span()),
             version: None,
             about,
             long_about,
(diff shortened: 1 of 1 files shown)


jdx:
Right, and applied to all six rather than to `settings` alone — 99fc1f3.

Every rule in `check_position` is about an attribute written where it cannot mean what it says, so every one of them should underline the attribute; all six spanned the struct's name. Fixing the reported one would have made it the odd one out, so `Cli` now records where `#[usage(...)]` was written and one helper spans them all. Before and after, from a real compile:

```
error: `settings` says this CLI resolves settings whose flags are declared elsewhere…
 --> src/main.rs:5:3
  |
5 | #[usage(bin = "ex", settings)]
  |   ^^^^^
```

and the same for `mount`, which I checked as the sibling case.

**On the requested span test — I could not write an honest one.** `Span::start()` needs `proc-macro2/span-locations`, which is off in this workspace, and without it every span in a `syn::parse_str` fixture is call-site — so a unit test cannot distinguish the attribute's span from the ident's even though rustc can. A test that passed either way would be worse than none, and enabling the feature to observe it seemed disproportionate. Verified against real compiles instead, for `settings` and for `mount`. Happy to add the dev-dependency feature if you'd rather have it pinned.

Rebased onto `main` (now carrying #899).

PR #905, opened 2026-08-16
feat(derive): say what a command does to the world

jdx:
First finding from dogfooding the derive on [communique](https://github.com/jdx/communique).

communique keeps `src/command_effects.rs` — 206 lines of a table keyed by command path, plus tests to keep the table in step with the CLI — for one reason: clap cannot express `effect`, so the classification is applied to the generated spec afterwards. mise has the same file for the same reason.

```rust
// before: a side table, and a second one keyed by (command, flag)
pub const EFFECTS: &[(&str, SpecCommandEffect)] = &[
    ("generate", Read), ("init", Write), ("sponsors", Read), ("usage", Read),
];

// after: beside the command that does the thing
#[derive(Args)]
#[usage(effect = "read")]
struct Generate {
    #[usage(long, effect = "write")]
    github_release: bool,
}
```

The flag half matters as much as the command half. `communique generate` only prints; `communique generate --github-release` writes. A table keyed by command path cannot say that, which is why communique's file carries a *second* table keyed by `(command, flag)`.

**Unsaid stays distinguishable from `read`.** A consumer treats an absent effect as "ask", so it has to survive as `None` rather than collapsing into the safe-looking answer.

**Refused on the root**, beside `mount` and `restart_token`. Bare `communique` does nothing to the world; one of its commands does. The spec writer already asserts it, so accepting it here would trade a message for a `debug_assert!` inside the writer.

## Verification

Asserted through the emitted spec rather than the in-memory tables, since a consumer reads the spec — a value the parser holds and the KDL drops is a CLI that documents different behaviour from the one that runs.

| mutation | result |
|---|---|
| a command's effect never reaches the metadata | FAILED |
| a flag's effect never reaches the metadata | FAILED |
| an unknown word reads as `read` | FAILED |
| an effect on the root is accepted | FAILED |

Workspace suite green, clippy clean, no exclusions.

## The rest of the communique gap list

Redeclared communique's whole CLI with the derive and diffed the emitted spec against its checked-in `communique.usage.kdl` (generated from clap). Everything matched except:

1. **`effect=`** — this PR.
2. **flag value-name casing** — clap defaults a flag's value placeholder to the upcased field name (`arg <CONFIG>`), the derive uses the field name verbatim (`arg <config>`). User-visible in help.
3. **`min_usage_version`** — communique prepends it by hand in `usage.rs`; the derive emits none.
4. **spec-level `usage "Usage: …"` line** — `clap_usage` emits one, the derive does not. usage-lib preserves it when present, so it round-trips.

Quoting style differs too (`cmd generate` vs `cmd "generate"`), but that is not semantic: the derive's output parses through usage-lib and re-emits identically.

## Summary by CodeRabbit

* **New Features**
  * Added support for declaring command and flag effects as `read`, `write`, or `destructive`.
  * Effect metadata is now included in generated command specifications.
  * Flag-level effects can override inherited command effects where applicable.

* **Validation**
  * Invalid effect values and unsupported declarations now produce compile-time errors.

* **Documentation**
  * Documented effect declarations, supported values, and applicable command arguments.

---

> [!NOTE]
> **Low Risk**
> Macro and spec-metadata changes with strong compile-time checks and conformance tests; no runtime auth or data-path changes.
> 
> **Overview**
> Adds **`effect = "read" | "write" | "destructive"`** on `#[derive(Args)]` structs (what running the command does) and on flags (what supplying the flag does), so effect metadata flows into **`CommandMeta`** and **`FlagMeta`** and the emitted KDL/spec instead of maintaining side tables like communique’s `command_effects.rs`.
> 
> Undeclared effects stay **`None`** so consumers can treat “unspecified” differently from **`read`**. **Codegen** wires `effect` into generated static metadata; **docs** describe the attribute.
> 
> **Compile-time validation:** invalid effect strings, **`effect` on the root `Cli`**, on positionals, or combined with **`subcommand`** (previously ignored silently) are rejected with targeted errors. **Conformance tests** assert command/flag effects round-trip through the parsed spec.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ddf863d536ada4c4fe57ad29ff1ce7e25b03cae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/conformance/tests/effect.rs
+++ b/conformance/tests/effect.rs
@@ -1,0 +1,139 @@
+//! What a command does to the world, declared beside the command.
+//!
+//! Not something clap can express, so a CLI that wants it keeps a side table keyed by command
+//! path and applies it to the generated spec afterwards. communique's `command_effects.rs` is
+//! two hundred lines of exactly that, and mise has the same file for the same reason.
+//!
+//! The spec's rule, which this only records: an invocation's effect is the highest of the
+//! command's and of every flag supplied, so a flag can raise what a command does and never
+//! lower it.
+
+use usage::Spec as LibSpec;
+use usage_derive::{Args, Cli, Subcommands};
+
+/// Generate release notes for a git tag
+#[derive(Args)]
+#[usage(effect = "read")]
+struct Generate {
+    /// Push editorialized notes to the GitHub release
+    #[usage(long, effect = "write")]
+    github_release: bool,
+    /// Generate notes without updating GitHub
+    #[usage(long)]
+    dry_run: bool,
+}
+
+/// Generate a config file in the repo root
+#[derive(Args)]
+#[usage(effect = "write")]
+struct Init {
+    /// Overwrite an existing config file
+    #[usage(long, effect = "destructive")]
+    force: bool,
+}
+
+/// Say nothing about what it does
+#[derive(Args)]
+struct Undeclared {
+    #[usage(long)]
+    plain: bool,
+}
+
+#[derive(Subcommands)]
+enum Command {
+    /// Generate release notes for a git tag
+    Generate(Box<Generate>),
+    /// Generate a config file in the repo root
+    Init(Box<Init>),
+    /// Say nothing about what it does
+    Undeclared(Box<Undeclared>),
+}
+
+/// Editorialized release notes powered by AI
+#[derive(Cli)]
+#[usage(bin = "communique", name = "communique")]
+struct Cli_ {
+    #[usage(subcommand)]
+    command: Option<Command>,
+}
+
+#[test]
+fn a_command_says_what_it_does_to_the_world() {
+    // Through the spec, because that is what a consumer reads: the point of declaring it is that
+    // something downstream — a confirmation prompt, an audit, a `--dry-run` wrapper — can tell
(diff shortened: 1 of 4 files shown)


jdx:
All three fixed in 4ddf863 — and they were one shape: declared, stored, silently dropped.

**`effect` on a positional** (flagged by all three of you). `arg_meta` has no field for one, so it compiled and vanished. Refused, with the reason stated: a positional is the thing being *acted on* rather than a choice to act, which is why the metadata has nowhere to put it.

**`effect` on a `#[usage(subcommand)]` field.** Worse than reported — that branch looked for `subcommand` and *ignored* everything else written beside it, so the declaration was never even parsed. I fixed the class rather than the instance: the field holds a set of commands, and everything the attribute can otherwise say describes a value or a flag. `long`, `default` and `global` on a subcommand field were equally quiet before this, and are refused now too.

**The docs table** said `effect` describes a flag. It also goes on an `Args`, where it says what *running* the command does — which is the half communique needed most.

Mutations: accepting an effect on a positional fails the new test; restoring the ignore-everything-else behaviour fails it for all four options.

PR #907, opened 2026-08-16
feat(derive): name a value the way clap names it, and say which usage can read the spec

jdx:
The other two gaps from the [communique](https://github.com/jdx/communique) dogfood, on top of #905.

## A value is named after its flag, shouted

clap prints `--max-tokens <MAX_TOKENS>` and `<TAG> [PREV_TAG]`; the derive printed `<max-tokens>` and `<tag>`. A visible change in `--help` for a CLI that changed nothing, which is the one thing this crate is trying not to do.

Measured from clap 4 rather than remembered — all three shapes:

```
Usage: ex [OPTIONS] <TAG> [PREV_TAG]

Options:
      --type <TYPE>
      --max-tokens <MAX_TOKENS>
  -c, --config <CONFIG>
```

Two things that fall out of those three lines:

- **From the form, not the field.** `#[usage(long = "type")] type_` renders `--type <TYPE>`, not `<TYPE_>`. The derive's existing rule already had this right and clap agrees; only the shouting was missing.
- **Underscores restored.** A flag is kebab to type and snake to read: `--max-tokens <MAX_TOKENS>`.

Set in the derive rather than in the renderers, so the metadata says what help prints. Two fallbacks would be two answers, and the spec is what docs and completions read.

## `min_usage_version`

Declarable on the root and emitted first, before anything an old `usage` would choke on — which is the whole point of the line.

**Declared rather than worked out.** Computing it means a table from every property to the version that introduced it, kept in step by hand; such a table rots into a spec claiming to be readable by a `usage` that chokes on it, which is worse than saying nothing. communique writes the line by hand today for exactly this reason.

## Not done, because it is not a gap

The spec-level `usage "Usage: …"` line that `clap_usage` emits. usage-lib parses it, carries it into the docs model, and renders it nowhere — commands compute their own via `SpecCommand::usage()`. Emitting it would be replicating a field no reader has.

## Verification

| mutation | result |
|---|---|
| a flag's value keeps the flag's own spelling | FAILED |
| a positional keeps its kebab name | FAILED |
| shouting drops the underscore restore | FAILED |
| `min_usage_version` is never written | FAILED |

**The gate's help parity over mise's 211 commands still holds** — `every_short_help_matches_the_reference`, `every_long_help_matches_the_reference` and `every_usage_line_matches_the_reference` are all green, and the checked-in shadows regenerate byte-identically. That is what says the placeholder change cost no fidelity.

Nine test expectations across the suite moved from lowercase to shouted; each is the intended change, not a workaround.

---

> [!NOTE]
> **Medium Risk**
> User-visible `--help` and emitted KDL change for every derived CLI (intentional clap parity); `min_usage_version` affects spec compatibility signaling but is opt-in on the root.
> 
> **Overview**
> Aligns **help and KDL placeholders** with clap 4: flag and positional value names are derived as **uppercase with underscores** (e.g. `--max-tokens <MAX_TOKENS>`, `<TAG>`) via a new `shout()` step in the derive, using the **flag/form name** rather than the Rust field (so `type_` → `<TYPE>`). Explicit `value_name` / `name =` still override. **Missing-required errors** now report those shouted names (e.g. `TARGET`, `FILES`).
> 
> Adds optional root **`min_usage_version`** on `#[derive(Cli)]`, plumbed through `Spec`, generated `SPEC`, and **KDL as the first property** so older `usage` readers can bail early; rejected on non-root structs like other spec-level options.
> 
> Conformance tests, snapshots, and fixtures are updated to expect shouted placeholders and the new field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdfd87a610224953fe82d05396c47d4255144480. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Both fixed in the head commit.

**Short-only placeholders** — real, and a good catch. That path names the value *before* the shouting default runs, so `-j <jobs>` sat beside `--jobs <JOBS>`: one CLI printing a placeholder two ways, and neither of them what clap prints. Measured rather than assumed:

```
Options:
  -j <JOBS>
  -m <MAX_TOKENS>
```

**`min_usage_version` on an `Args`** — parsed, stored, dropped, since only the root emits a spec. Refused where it is written, beside the other root-only options.

Mutations: unshouting the short-only path fails `a_short_only_flag_keeps_a_descriptive_placeholder`; accepting `min_usage_version` below the root fails the new rejection test.

PR #910, opened 2026-08-16
feat(derive): let `parse()` answer a failure the way a program does

jdx:
The second regression from the communique port. `parse()` rendered a failure with `{:?}`:

```
UnknownFlag { token: [45, 45, 110, 111, 112, 101] }
```

while the clap-shaped rendering sat in this crate unused. communique's `main.rs` hand-rolled 34 lines to reach it, which is not something an adopter should have to work out.

`parse()` is the entry point that **is** the process — it already printed a help page and exited. It now prints the message to stderr and exits **2**, clap's status, so a script checking for it keeps working. `parse_from` still hands the error back, for a library embedding a CLI that wants to decide.

## Why the renderer is reached through a function rather than a `#[cfg]`

Not stylistic. Whether the good rendering exists is a feature of **usage-argv in the adopter's graph**, and a `cfg` written into generated code is evaluated in the *adopter's* crate, where the feature is not theirs to see — which is exactly how a metadata field got silently dropped earlier in this project. So the cfg lives beside the thing it gates: `render_failure` is the clap-shaped message with `diagnostics` on, and the Debug form without it, which is what a parser-only build asked for.

## What it buys the adopter

With this and #909, communique's port matches clap on `--version`, `-V`, every error message tested, and every exit code:

```
########## --version
  clap : communique 1.3.1 [exit 0]
  usage: communique 1.3.1 [exit 0]
########## badcmd
  clap : error: unrecognized subcommand 'badcmd' [exit 2]
  usage: error: unrecognized subcommand 'badcmd' [exit 2]
########## --fore
  clap : error: unexpected argument '--fore' found [exit 2]
  usage: error: unexpected argument '--fore' found [exit 2]
```

and `main.rs` is 34 lines shorter than the port needed before it.

**Breaking:** `parse()` returns `Self` rather than `Result<Self, String>`. Nothing is released yet.

## Verification

Mutation: making `render_failure` fall back to the Debug form with `diagnostics` on fails the new test. Both feature configurations — `spec` alone and `spec,diagnostics` — build clean.

---

> [!NOTE]
> **High Risk**
> Breaking change to generated `parse()`’s return type and all failure/help/version exit paths; every derive adopter’s `main` must migrate from `Result` handling to the new process semantics.
> 
> **Overview**
> **`Cli::parse()` is now a process-style entry point:** it returns `Self` (breaking change from `Result<Self, String>`), prints help/version to stdout and exits 0, and on real parse failures prints a user-facing message to stderr and exits **2** (clap’s code). **`parse_from` still returns errors** for embedders that want control.
> 
> Adds **`usage_argv::render_failure`** in the argv crate with `#[cfg]` beside the gated renderer: with **`diagnostics`**, output matches clap-shaped diagnostics; without it, a minimal `Debug`-style line—so generated code does not need feature cfgs in the adopter crate.
> 
> Conformance adds **`strip_ansi`** and a test that **`render_failure`** wording matches clap for an unknown flag (e.g. `--nope`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a3201ee1abf5f28fb63e9d976ac4b3da9f0604a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -665,6 +665,33 @@ fn find_named<'t>(cmd: &'t Command<'t>, name: &[u8]) -> Option<&'t Command<'t>> 
         .find(|c| c.name.as_bytes() == name || c.aliases.iter().any(|a| a.as_bytes() == name))
 }
 
+/// What a caller should print for a parse failure, and what to exit with.
+///
+/// The one entry point a generated `parse()` reaches for, and the reason it exists here rather
+/// than in the derive: whether the good rendering is available is a *feature of this crate* in
+/// the adopter's dependency graph, and a `#[cfg]` written into generated code is evaluated in
+/// the adopter's crate, where the feature is not theirs to see. That is how a metadata field
+/// once got silently dropped; the answer is that the cfg lives beside the thing it gates.
+///
+/// With `diagnostics` on, this is the clap-shaped message. Without it, the error's `Debug`
+/// form — which is still better than nothing and is what a parser-only build asked for.
+///
+/// [`Error::Help`] and [`Error::Version`] are not failures and must be handled before this.
+#[cfg(feature = "diagnostics")]
+pub fn render_failure(spec: &spec::Spec<'_>, argv: &[&OsStr], error: &Error<'_, '_>) -> String {
+    diagnostic::render(spec, argv, error, diagnostic::Style::auto())
+}
+
+/// What a caller should print for a parse failure, without the renderer that makes it readable.
+///
+/// See the other half. A caller that wants the clap-shaped message turns on `diagnostics`;
+/// this is what a parser-only build asked for, and it still says which error it was.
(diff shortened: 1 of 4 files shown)


jdx:
Real, and a good catch — fixed in the head commit.

Reproduced before fixing: `CLICOLOR_FORCE=1 cargo test --test version` failed with

```
\x1b[1m\x1b[31merror:\x1b[0m unexpected argument '\x1b[33m--nope\x1b[0m' found
```

a message that is perfectly correct, failing a `starts_with` that could not see past the escapes. A test whose result depends on the ambient terminal is a flake waiting for the machine that has one — CI happens not to be a TTY today, which is the only reason it was green.

The message is stripped before it is read. What the test is about is the wording; the colouring has tests of its own. Green under a plain run, `CLICOLOR_FORCE=1`, and `NO_COLOR=1`.

PR #911, opened 2026-08-16
fix(help): a command's page should say what that command does

jdx:
First of the help findings from the communique dogfood, and the clearest.

`communique generate --help` printed this:

```
communique 1.3.1
Editorialized release notes powered by AI

Usage: communique generate [FLAGS] <TAG> [PREV_TAG]
```

and never once said what `generate` does. Every page carried the **root's** banner and description, so the one thing a subcommand's help is for was the one thing missing from it. clap prints the command's own description there.

After:

```
Generate release notes for a git tag

Usage: communique generate [FLAGS] <TAG> [PREV_TAG]
```

The root keeps its banner and the program's description — it has no command of its own to describe, and a program's page is where a program introduces itself.

**Changed in usage-lib and usage-argv together**, so the gate still holds them byte-identical over mise's 211 commands. The point of that gate is that two implementations agree, not that either is frozen; and doing it this way means every usage-based CLI gets the fix, not only the ones using the derive.

## Verification

| mutation | result |
|---|---|
| every page shows the program's description (the old behaviour) | FAILED |
| every page shows the banner | FAILED |

`every_short_help_matches_the_reference` and `every_long_help_matches_the_reference` still green over mise's whole spec. Workspace suite green, clippy clean.

---

## The rest of the help differences, for your call

Measured on the same page, clap vs usage. I have **not** acted on these — they are judgement calls about what is best rather than bugs, and you said you are not tied to either.

**1. Global flags are missing from subcommand pages.** `communique generate` accepts `--config`, `--verbose` and `--quiet`; its help lists none of them. clap lists all three. This is the other content loss and I think it should be fixed — a flag a user can type and cannot discover is the worst case.

**2. Long forms do not align.** clap indents so every long form starts at the same column:

```
      --github-release           Push editorialized notes to the GitHub release
  -n, --dry-run                  Generate notes without updating GitHub
```

we print both at column 2:

```
  --github-release           Push editorialized notes to the GitHub release
  -n --dry-run               Generate notes without updating GitHub
```

clap's is easier to scan. This is the one I would change next.

**3. No comma between short and long.** `-n --dry-run` vs clap's `-n, --dry-run`. The comma is near-universal.

**4. `-h, --help` is not listed** (nor `--version`, per #909). The existing reasoning is that a spec declares neither, so listing one would make the page disagree with the spec — sound for the spec, but help is for people, and clap lists both. Worth revisiting.

**5. `[FLAGS]` vs clap's `[OPTIONS]`.** Cosmetic; `OPTIONS` is the commoner convention and what clap 4 settled on.

**Where we are already better:** our wrapping. clap runs to 100+ columns and lets the terminal break lines mid-word; we wrap with a hanging indent, which is much better on a narrow terminal. I would keep that.

Say which of 1–5 you want and I will do them in order; 1 and 2 are the two I would pick.

---

> [!NOTE]
> **Low Risk**
> User-visible help text only; behavior aligns with clap and keeps usage-argv and usage-lib in sync. Clap conversion fix is correctness for derived specs, not security-sensitive.
> 
> **Overview**
> Subcommand `--help` / `-h` no longer repeats the **root** name, version, and program `about`; it leads with that command’s own short or long description (clap-style). The **root** page still shows the banner and program intro.
> 
> The same rules are applied in **usage-argv** (`short_help` / `long_help`) and **usage-lib** (Tera templates get a `root` flag from empty `full_cmd`).
> 
> **Clap → `Spec` conversion** now runs `set_subcommand_ancestors` like KDL parsing, so nested commands get correct `full_cmd` and non-empty usage strings (e.g. `go fast` instead of a blank command path).
> 
> Conformance tests cover subcommand vs root help; a unit test covers clap-derived subcommand paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0f72333d6c779fa475ab11407d2956e903361b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Improved CLI help output to clearly distinguish the root command from subcommands.
  - Root help now displays program name, version, and description.
  - Subcommand help now displays the relevant command description, including long-form text when available.
  - Nested subcommands now show accurate command paths and usage information.

- **Bug Fixes**
  - Prevented root-level metadata from appearing on subcommand help pages.
  - Improved parsing and help behavior for commands with options such as `--env`.


jdx:
Confirmed, and the cause is one level below the check — fixed in 66dd3bc.

I probed it rather than reasoning about it. For a spec built from clap:

```
full_cmd=[]  usage=""  usage()=""
```

`set_subcommand_ancestors` runs only in the KDL `FromStr` path (`lib/src/spec/mod.rs:309`); the `From<&clap::Command>` impl never called it. So **the root check is right and the data feeding it was not** — and this was already broken before this PR in a second place: `SpecCommand::usage()` joins `full_cmd`, so every clap-derived subcommand's usage line rendered with no command in it.

One pass, run where the other one is, walking all the way down:

```rust
let go = spec.cmd.subcommands.get("go").unwrap();
assert_eq!(go.full_cmd, ["go"]);
assert_eq!(go.usage, "go <SUBCOMMAND>");
let fast = go.subcommands.get("fast").unwrap();
assert_eq!(fast.full_cmd, ["go", "fast"]);
assert_eq!(fast.usage, "go fast");
```

That fixes the misclassification you flagged, for `SpecCommandBuilder`, `spec_cmd!` and the clap conversion alike, since all three land in the same `Spec`.

The gate stays green — `every_short_help_matches_the_reference` and `every_long_help_matches_the_reference` over mise's 211 commands.

PR #913, opened 2026-08-16
feat(help): list the flags a command inherits

jdx:
PR 2 of the help plan, and the content loss that started it.

`communique generate` accepts `--config`, `--verbose` and `--quiet` from its root, and its page mentioned **none of them**. A flag a user can type and cannot discover is the worst way for help to be wrong, and it was true of every subcommand page in every usage CLI.

```
Flags:
      --base-url <BASE_URL>      Base URL for the LLM API
  -o, --output <OUTPUT>          Write output to a file instead of stdout

Global flags:
  -v, --verbose                  Enable verbose logging output
  -q, --quiet                    Suppress progress output
  -c, --config <CONFIG>          Path to config file (default: communique.toml in repo root)
```

**Under a heading of their own** — this is where the design differs from clap on purpose, per your call. clap mixes them into the one list; `--config` belongs to the program rather than to `generate`, and a reader should be able to see which is which. One column across both sections, so the page reads as one table with a rule through it rather than two that happen to be adjacent.

## Shadowing, which I did not plan for and which the first version got wrong

`in_scope` chains a command's own flags before its ancestors' and takes the first match, so `mise use --raw` is *use's* `--raw` and never the root's. My first version listed both:

```
      --raw    Connect backend install command stdin/stdout/stderr directly to the terminal
...
Global flags:
      --raw    Read/write directly to stdin/stdout/stderr instead of by line
```

Two descriptions for one spelling, one of which can never apply. Decided nearest-ancestor-first and emitted root-first, which is the order a reader meets them walking down. `mise use` now lists `--raw` once, with the description that will actually be used.

**The root grows no such section**: its flags are its own, `global` or not, and there is nothing above it to inherit from.

## What made it possible

`find` walked the tree and **threw the chain away**, returning only the command — so neither renderer had ancestors to ask. It returns the chain now. That also fixes the help half of the pointer-identity bug I reported on the diagnostics PR: a `Subcommands` type mounted under two parents is one `Command` at one address, and a page rendered for the second mount showed the first one's ancestry.

In usage-lib the chain comes from `full_cmd`, which is the typed path — exact, with none of the ambiguity a search would have. (That field only became reliable for clap-derived specs in #911.)

## Verification

| mutation | result |
|---|---|
| globals not listed at all | FAILED |
| shadowing dropped | FAILED |
| non-`global` flags inherited too | FAILED |
| the root grows a global section | FAILED |

Gate green over mise's 211 commands, so both renderers moved together. Workspace suite green, clippy clean.

---

> [!NOTE]
> **Medium Risk**
> Help output changes for every subcommand page and shadowing must stay in sync between usage-argv and usage-lib; behavior is heavily tested but user-visible help text will differ widely.
> 
> **Overview**
> Subcommand **`-h` / `--help`** now lists flags inherited from ancestors (only those marked **`global`**) under a **Global flags** heading, separate from the command’s own **Flags** section, with one shared column width across both.
> 
> **`help::find`** returns the full **`CommandMeta` ancestry chain**, not just the leaf command. **`short_help`**, **`long_help`**, **`render`**, and completion use that chain so inherited globals and correct ancestry (e.g. shared subcommand mounts) are available everywhere.
> 
> Inherited entries follow **parser shadowing**: nearer commands win on duplicate spellings; hidden flags still reserve spellings; partial multi-long/short survival; negations vs longs resolved like **`long_flag`** (any long in scope before negations). **`usage-lib`** mirrors the same logic in **`inherited_flags`** and Tera templates so gate tests stay aligned with usage-argv.
> 
> Call sites and tests pass explicit chains (e.g. **`&[spec.root, config, set]`**); new **`conformance/tests/global_flags.rs`** covers listing, deduplication, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5aab55a4ac4c47c15416e91cd2bbfcb67d3d8172. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Subcommand help now displays inherited global flags in a dedicated “Global flags” section.
  - Global flag details—including aliases, descriptions, choices, and environment variables—appear consistently in short and long help.
  - Help output respects hidden flags, aliases, negations, and subcommand overrides.

- **Bug Fixes**
  - Improved metadata resolution for nested commands.
  - Help layouts now maintain consistent columns and wrapping across local and inherited flags.

- **Tests**
  - Added comprehensive coverage for inherited flags, visibility, precedence, and nested command help.

### DIFF
--- a/argv/src/complete.rs
+++ b/argv/src/complete.rs
@@ -198,7 +198,8 @@ pub fn for_name<'a>(
     // Tree order last, and only as a fallback: two sibling commands may take a `tool` and mean
     // different things by it, and the one the line reached is the one being asked about.
     let reached = walk(spec.root.cmd, ctx.command_words_start());
-    let reached_meta = crate::help::find(spec, reached.cmd).map(|(_, meta)| meta);
+    let reached_meta =
+        crate::help::find(spec, reached.cmd).and_then(|(_, chain)| chain.last().copied());
     let owner = if on(spec.root, name).is_some() {
         spec.root
     } else if let Some(meta) = reached_meta.filter(|meta| on(meta, name).is_some()) {
@@ -407,7 +408,7 @@ fn files_for(name: &str) -> Option<Files> {
 /// for a word nothing is known about, and the `run=` completions a spec can declare.
 pub fn complete<'a>(spec: &'a Spec<'a>, split: &Split) -> Completions<'a> {
     let position = walk(spec.root.cmd, split.argv());
-    let meta = crate::help::find(spec, position.cmd).map(|(_, meta)| meta);
+    let meta = crate::help::find(spec, position.cmd).and_then(|(_, chain)| chain.last().copied());
     let token = split.prefix.as_str();
     let candidates = candidates(spec, split);
 
@@ -476,7 +477,7 @@ pub fn complete<'a>(spec: &'a Spec<'a>, split: &Split) -> Completions<'a> {
 /// Just the candidates this CLI knows about, without the question of paths.
 pub fn candidates<'a>(spec: &'a Spec<'a>, split: &Split) -> Vec<Candidate<'a>> {
     let position = walk(spec.root.cmd, split.argv());
-    let meta = crate::help::find(spec, position.cmd).map(|(_, meta)| meta);
+    let meta = crate::help::find(spec, position.cmd).and_then(|(_, chain)| chain.last().copied());
     let token = split.prefix.as_str();
 
     let mut out = if position.flags_possible && token == "-" {
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -138,8 +138,24 @@ impl<'a> Shown<'a> {
         }
     }
 
-    /// What is left of a flag once `taken` has had its pick.
-    fn surviving(meta: &'a FlagMeta<'a>, taken: &[String]) -> Self {
+    /// What is left of a flag once everything nearer has had its pick.
+    ///
+    /// `taken` is the longs and shorts already claimed; `taken_negations` the negations;
+    /// `every_form` every long and short in scope at any distance, because the parser resolves
+    /// a word against all of those before it looks at a negation at all.
+    fn surviving(
+        meta: &'a FlagMeta<'a>,
+        taken: &[String],
+        taken_negations: &[String],
+        every_form: &[String],
+    ) -> Self {
+        let mine: Vec<String> = meta
+            .flag
+            .longs
+            .iter()
+            .map(|l| format!("--{l}"))
+            .chain(meta.flag.shorts.iter().map(|s| format!("-{}", *s as char)))
+            .collect();
         Shown {
             long: meta
                 .flag
@@ -153,10 +169,12 @@ impl<'a> Shown<'a> {
                 .iter()
                 .copied()
                 .find(|s| !taken.contains(&format!("-{}", *s as char))),
-            negate: meta
-                .flag
-                .negate
-                .is_some_and(|n| !taken.contains(&format!("--{n}"))),
+            negate: meta.flag.negate.is_some_and(|n| {
+                let spelling = format!("--{n}");
+                // A long anywhere in scope wins over this, this flag's own excepted.
+                !taken_negations.contains(&spelling)
+                    && (!every_form.contains(&spelling) || mine.contains(&spelling))
+            }),
         }
     }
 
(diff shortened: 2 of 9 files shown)


jdx:
Two of the three addressed; the third was **my error and I have corrected it in the commit message**.

**Hidden flags skip the shadowing check** (Bugbot) — real. This built the shadow set from the *visible* own flags while usage-lib built it from all of them, so the two renderers disagreed wherever a hidden local shared a spelling with an inherited global, and the page offered a `--raw` that binds something its reader cannot see. `hide` keeps a flag off the page; the parser still binds it, so it still shadows.

**Partial collisions hide valid aliases** (greptile) — real, and reproduced before fixing:

```
Flags:
  -v <LEVEL>  The subcommand's own -v, a different thing

parser accepts `sub --verbose`: true      ← and the page never mentioned it
```

Shadowing is per spelling now. The entry renders as `--verbose` alone — what survives is offered, what was claimed is not.

That surfaced a real case in mise, which the gate caught: the root's global is `-E --env`, and a descendant claiming `--env` leaves `-E` inherited. `-E… <ENV>` on its own gives a reader nothing to connect to the `--env` they saw elsewhere, so the declared-name prefix is judged on the forms the page is *showing* and the entry reads `env: -E… <ENV>` — which is what usage-lib already did.

## Correction

Both of you flagged that `help::find` still resolves a shared mount by address. **You are right and my commit message was wrong** — I claimed returning the chain fixed it. It does not:

```
$ help for beta's `shared`
Usage: ex alpha shared [--thing]

Global flags:
      --alphaglobal
```

Returning the chain gives the ancestors *of whichever mount the DFS found first*, which is still alpha's. The bug predates this branch — `find` always searched by pointer — but this branch makes it worse, since the wrong ancestry now shows up as wrong globals and not only a wrong usage line. Fixing it needs a route rather than a pointer, which is its own change and not this one. The commit message now says so.

## On the API break

`short_help`/`long_help` taking a slice is source-incompatible, but usage-argv has never been published — there are no downstream callers to break. Worth revisiting before the first release.

## Note on the base

This now sits on #916, which fixes two things in the column this branch builds on. GitHub shows `main` as the base because that PR is tracked in a separate stack; the commit list is the honest view, and the diff cleans up when #916 merges.

jdx:
All four addressed, and they turned out to be one root cause — so rather than patch each, "which spelling is already spoken for" is now decided in **one** place, by the parser's own rule: `in_scope` chains a command's own flags before its ancestors', nearest first, and the first match binds. A page offers a spelling only where the flag it describes is the one that would take it.

Three things the model now counts that the first version did not:

**Hidden flags on an ancestor** (CodeRabbit). They bind; `hide` only keeps them off the page. So a nearer hidden global reserves its spelling, or a farther one gets advertised while the hidden one answers. The command's *own* hidden flags were fixed a commit ago and ancestors had the identical hole.

**Negations** (Bugbot). `--no-colour` is a spelling like any other, in both directions — a descendant taking it leaves the root's positive form on offer, and a descendant whose *negation* is the root's plain long claims that.

**Every long and short, not just the first** (greptile). A global answering to `--jobs` and `--workers` loses only the one that was taken. That is why the mask became a set of surviving spellings rather than a pair of booleans.

**The test chain** (CodeRabbit) — right, and worth stating plainly: `[root, set]` for a command nested under `config` had a hole in it, so an ancestry regression would have passed. Fixed.

## On the aliases suggestion

I did not take this one, and here is why. `flag.aliases` appears in the long template's *fallback* branch and not in the `help_rendered` branch — but **usage-argv prints flag aliases in neither**. Adding them to one branch would make the two renderers disagree and break the gate. The real inconsistency is that lib prints them at all in one branch; the consistent fixes are either teaching argv to print them in both or dropping them from lib, and that is a decision about whether flag aliases should be discoverable rather than a bug in this PR. Left alone deliberately, and worth its own change.

## Verification

| mutation | result |
|---|---|
| negations dropped from the claim set | FAILED |
| only the first long survives the mask | FAILED |
| hidden ancestor globals do not reserve | FAILED |
| `Shown::surviving` ignores a claimed negation | FAILED |
| the whole flag is dropped on a partial collision | FAILED |

Two of those **survived on the first attempt** because no fixture reached them — a nearer *hidden* global, and a local negation claiming an ancestor's plain long. Both now have cases; the gate over mise's 211 commands stays green.

jdx:
Thanks — triaged all four. Three of them were read against `1c7160c`, which is a pre-rebase copy of this branch's first commit; the branch head is `5aab55a`, three commits further on. Taking them in turn:

**Shared mounts resolve wrong ancestry** (Greptile, Bugbot) — correct, and it is the bug I flagged on #897. It is not fixable here: a `&Command` cannot distinguish two mounts of one address, so the input has to change, not the search. Fixed in #928 at the top of this stack, which resolves the page by the route the argv took. Measured before and after:

| argv | before | after |
|---|---|---|
| `beta shared --help` | `ex alpha shared` / `--alphaglobal` | `ex beta shared` / `--betaglobal` |
| `help beta shared` | `ex alpha shared` / `--alphaglobal` | `ex beta shared` / `--betaglobal` |

**Partial collisions hide valid aliases** — already fixed on head. `1c7160c` had `if claims(f).any(|c| taken.contains(&c))`, all-or-nothing, which is the code you read. Head masks per spelling: `Shown::surviving(f, &taken)` keeps the forms nothing nearer has taken, and the flag is dropped only when `show.nothing()`. A descendant taking `--jobs` leaves an inherited `--workers` listed and findable.

**Hidden flags skip shadowing check** (Bugbot, CodeRabbit) — also already fixed on head, and the fix is the one both of you described. `1c7160c` seeded `taken` from `own` (visible only); head seeds it from `here.flags`, hidden included, and reserves every ancestor global whether or not it is shown, with the `continue` for `hide` placed *after* the `taken.extend`.

**Public help API breaks callers** — valid in the letter, and intended. `short_help`/`long_help` do take a chain now, and that is source-breaking. Not shimmed, because the only shim available is to pass a one-element chain — which compiles and silently renders the page with no inherited flags, i.e. reintroduces exactly the bug this PR exists to fix. A quiet wrong answer is worse than a compile error that says what changed. They also cannot be made private: the parity gate and the conformance suite call them directly to render a named command against a known chain. Recorded as breaking; the crate is pre-adoption and this stack is a run of breaking changes ahead of the first real release.

PR #914, opened 2026-08-16
feat(help): list `--help` and `--version`, which every page answers

jdx:
PR 3, the last of the four help differences. Both flags worked and neither appeared on any page: a reader looking for how to ask for help had to already know.

```
Flags:
  -o, --output <OUTPUT>  Write output to a file instead of stdout
  -h, --help             Print help
  -V, --version          Print version
```

## The rule this reverses, and the half it keeps

The old rule was that a page lists exactly what its spec declares. Per your call, the page now names these two — but the **spec still does not**, because the parser supplies them and a spec claiming otherwise would have every reader inventing a flag its CLI never wrote.

Help is written for people; the spec has its own readers, and they are not the same readers. A test asserts both halves, which is the honest way to record a reversal.

## Where they are, and are not

- **`--version` only where the parser accepts one** — a command whose table says so, which is the root of a CLI that declared a version. A subcommand's page does not offer it, because `ex other --version` is refused. Offering a flag that would be refused is worse than staying quiet.
- **Each spelling dropped where the CLI claimed it.** `-V` taken by `--verbose` leaves `--version` listed alone; the parser yields to a declaration, and the page has to say what will actually happen rather than what would be tidy.
- **Last in the command's own section**, where clap has them, carrying no `help_heading` — so a CLI that groups its flags gets them at the end of the ungrouped list rather than inside somebody else's section.

## Verification

| mutation | result |
|---|---|
| nothing supplied is listed | FAILED (3 tests) |
| `--version` offered where the parser refuses it | FAILED |
| a claimed spelling is offered anyway | FAILED |

Gate green over mise's 211 commands. Seven expected pages in `lib/tests/parse.rs` and a handful of insta snapshots gain the line — each the intended change, checked one at a time rather than blanket-accepted.

---

That completes the four differences from the plan. `communique generate -h` is now byte-for-byte clap's output apart from the word `Flags:`, plus a `Global flags:` section clap has no equivalent for — and short help, which had no column at all, has one.

---

> [!NOTE]
> **Medium Risk**
> User-visible help text changes across all commands, but behavior is tightly gated by claimed spellings and covered by conformance, insta, and parse snapshot tests.
> 
> **Overview**
> Help output now shows **`-h, --help`** and **`-V, --version`** (when applicable) at the end of each command’s unheaded flags section, matching clap-style discoverability. The **spec/KDL still does not declare** these flags—the parser supplies them—so emitted specs stay unchanged while rendered pages name what users can actually type.
> 
> Both **`usage-argv`** (`supplied` / `supplied_entries` wired through `own_and_global`) and **`lib` docs rendering** (`supplied_flags`, with `inherited_flags` returning the ancestor “taken” spellings) pick long/short/both variants from what is already claimed, including **hidden flags and negations**. **`--version`** appears only on the **root** when a version exists; subcommands and unversioned CLIs omit it. Mise’s **`disable_help`** suppresses the supplied help line on KDL-rendered pages.
> 
> Conformance and snapshot tests were flipped from “must not list” to “listed but not declared,” with new cases for claimed spellings, hidden help, and negated `--version`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6c0d410d1cb037d73a2f5a59cbd18691cef66cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -1159,7 +1159,14 @@ fn own_and_global<'a>(
     // negations, and a `--help` the page offers while something else binds it is exactly the
     // lie this whole model exists to prevent.
     let mut own = own;
-    own.extend(supplied_entries(here.cmd, &taken));
+    // Forms *and* negations: `long_flag` asks `find_negation` before it offers `--version`,
+    // so a declared negation beats a supplied flag even though it loses to a long.
+    let claimed: Vec<String> = taken
+        .iter()
+        .cloned()
+        .chain(taken_negations.iter().cloned())
+        .collect();
+    own.extend(supplied_entries(here.cmd, &claimed));
     (own, inherited)
 }
 
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -939,6 +939,135 @@ pub fn find<'a>(
     walk(&mut path, &mut chain, spec.root, cmd).then_some((path, chain))
 }
 
+/// The entries for `--help` and `--version`, which the parser supplies and no spec declares.
+///
+/// Listed because help is written for people: a reader looking for how to get help should find
+/// it on the page. This reverses the rule these two used to follow — that a page lists exactly
+/// what its spec declares — and the reason is that the spec has its own readers, and they are
+/// not the ones reading this.
+///
+/// Four spellings each, because a CLI may have claimed either form for itself. The parser
+/// yields to a declaration (`in_scope` looks a command's own flags up first), so a page that
+/// claimed otherwise would be describing a flag that never binds.
+mod supplied {
+    use crate::spec::FlagMeta;
+    use crate::Flag;
+
+    macro_rules! entry {
+        ($name:ident, $flag:ident, $key:expr, $label:expr, $longs:expr, $shorts:expr, $help:expr) => {
+            static $flag: Flag<'static> = Flag {
+                key: $key,
+                name: $label,
+                longs: $longs,
+                shorts: $shorts,
+                ..Flag::BOOL
+            };
+            pub static $name: FlagMeta<'static> = FlagMeta {
+                flag: &$flag,
+                help: Some($help),
+                ..FlagMeta::EMPTY
+            };
+        };
+    }
+
+    entry!(
+        HELP_BOTH,
+        HB,
+        crate::HELP_LONG_KEY,
+        "help",
+        &["help"],
+        b"h",
+        "Print help"
+    );
+    entry!(
+        HELP_LONG_ONLY,
+        HL,
+        crate::HELP_LONG_KEY,
+        "help",
+        &["help"],
+        b"",
+        "Print help"
+    );
+    // Named `h`, not `help`: the declared name is judged against the forms the entry shows,
+    // and a short-only entry called `help` reads as a renamed flag — it printed `help: -h`.
+    entry!(
+        HELP_SHORT_ONLY,
+        HS,
+        crate::HELP_SHORT_KEY,
+        "h",
+        &[],
+        b"h",
+        "Print help"
+    );
+    entry!(
+        VERSION_BOTH,
+        VB,
+        crate::VERSION_LONG_KEY,
+        "version",
+        &["version"],
+        b"V",
+        "Print version"
+    );
+    entry!(
+        VERSION_LONG_ONLY,
+        VL,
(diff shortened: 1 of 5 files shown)


jdx:
All three addressed, and the first two fall out of the claim model rebuilt in #913 below.

**Hidden and negated claims** (greptile, Bugbot). `supplied_entries` decided from the *visible* own and inherited lists; it now asks the same set every other decision on the page uses, which counts hidden declarations and negations. The parser yields to both, so a page offering `--help` while a hidden `--help` binds would describe an action that typing it does not perform.

**Short-only entries showed the wrong name** (Bugbot) — real, and reproduced:

```
Flags:
      --quiet / --version  ...
  help: -h                 Print help
  version: -V              Print version
```

The declared name is judged against the forms the entry *shows*, so a short-only entry called `help` reads as a renamed flag. The leftover entries are named after the form they show.

**Ungrouped section order** (Bugbot). Right, and the snapshot that accepted it was mine. `group_by_heading` sorts the unheaded group to the front and argv's `groups_section` emits it there, so pushing the supplied group put `Flags:` after the headed sections in one renderer and before them in the other. Inserted first now; the snapshot is corrected rather than re-accepted:

```
+Flags:
+  -h, --help              Print help
+
 Filtering:
-      --filter <pattern>  Only matching
```

## Verification

| mutation | result |
|---|---|
| a short-only entry keeps the long name | FAILED |
| hidden and negated claims ignored | FAILED |

Gate green over mise's 211 commands, workspace suite green, clippy clean.

jdx:
Triaged all three. Two were already fixed on head; the third turned out to be a real regression, though not on this branch — chasing it is what found it, so thank you.

**Short-only flags show wrong name** — fixed in #916. A short-only entry is named after the form it shows (`"h"`, `"V"`), so `flag_usage`'s "declared name not implied by the forms" rule no longer fires and the page prints `-h` rather than `help: -h`. Both twins.

**Ungrouped section order mismatch** — fixed on head, and the code now says so where it happens: the supplied entries are `insert(0, …)`ed rather than pushed, because `group_by_heading` sorts the unheaded group to the front and argv's `groups_section` emits it there. A CLI that heads every one of its flags would otherwise get `Flags:` after the headed sections here and before them there.

**Hidden claims still offered** — this one was live, and worse than reported. `supplied_flags` was deciding against the *display* list of inherited flags, so a hidden ancestor global binding `--help` was invisible to it while argv's `taken` counted it. Measured, subcommand page under a root declaring `#[usage(long = "help", global, hide)]`:

```
argv: ["  -h           Print help"]
lib : ["  -h, --help   Print help"]   ← advertises a spelling the hidden flag binds
```

Fixed on this branch in `93919e0`: `inherited_flags` returns its claim set alongside its display list, and `supplied_flags` decides against the claims.

The part worth flagging: that commit *was* on this branch when you reviewed it, and every branch above this one had reverted it — an earlier restack had resolved a conflict in `lib/src/docs/cli/mod.rs` by taking the old side, silently dropping it along with two other help fixes. So the divergence you describe was real again at the top of the stack even though it was fixed here. The four upper branches have been rebased onto this one; `inherited_flags` now returns the claim set on every branch in the stack, and the probe above agrees on both renderers at the top. The gate did not catch it because mise's spec has no hidden global claiming a supplied spelling.

jdx:
Second pass on this PR's review. Most of the newer comments re-report findings against pre-rebase commits — the negate double-dashes (`----no-color`) was fixed in `5aab55a`, per-spelling masking in `6ad1a4b`, and the hidden-inherited claim in `93919e0`, which now threads the claim set out of `inherited_flags` rather than rebuilding it from the visible list.

One was new and correct: **help listed when disabled**. `disable_help` makes `is_help_arg` refuse `--help`, `-h` and `-?` outright, and the supplied entries were appended without consulting it — the page advertised an action its own parser would not perform, which is the exact lie this PR is otherwise careful about. Fixed in `c6c0d41`, with a snapshot test; `--version` stays, since nothing disabled that.

No twin change in usage-argv, and no divergence either: `disable_help` is a KDL word with no equivalent in the argv tables, so no spec that crate can hold ever carries one. Called out in the commit rather than left implicit, since byte-identical output is the invariant this area runs on.

PR #918, opened 2026-08-16
fix(ci): check the version the crates promise, and promise one that is true

jdx:
The MSRV said 1.80 and nothing checked it. The first run of a job that actually tries it:

```
error[E0658]: referencing statics in constant functions is unstable
 --> argv/src/lib.rs:550:21
```

**usage-argv has not built at its declared floor for a long time.** 1.80 was set in 2024 when `LazyLock` landed and never revisited — about two years of drift. A promise nobody checks is a guess.

I came to this from the other direction: a `use<..>` capture (Rust 1.82) got into usage-argv earlier in this stack and CodeRabbit caught it by reading, not by building. That is not a check.

## Two floors, because the crates genuinely differ

| | version | why |
|---|---|---|
| usage-argv, usage-derive, usage-config | **1.91** | no KDL dependency; mise's own floor, which is the one that matters for the fleet |
| usage-lib, usage-config-build, clap_usage, usage-cli | **1.95** | `kdl` declares 1.95, and there is no arguing with a dependency |

`clap_usage` and `usage-cli` are published and declared **nothing at all**, which is the same gap without the drift.

Each crate is checked at the version *it* declares rather than at one shared guess — that is what keeps the lower floor real rather than aspirational, and usage-argv's low floor is a genuine feature of a crate with no dependencies. The job also asserts the matrix and the manifests agree, so a crate that raises its floor without touching the workflow fails instead of being quietly tested at the old one.

## Verification

Installed both toolchains and checked at the exact versions, not at stable:

```
### at 1.91          ### at 1.95
  usage-argv    ok     usage-lib           ok
  usage-derive  ok     usage-config-build  ok
  usage-config  ok     clap_usage          ok
                       usage-cli           ok
```

The matrix/manifest guard was run under `bash` locally — my shell is zsh, which does not word-split unquoted expansions, so the first version of that check silently passed on nothing.

**Worth a look:** the workspace is on edition 2021 while mise is on 2024. With a 1.91 floor, 2024 is available. Not in this PR.

---

> [!NOTE]
> **Low Risk**
> Changes are CI guards and Cargo.toml MSRV metadata only; no runtime or library behavior changes.
> 
> **Overview**
> Adds a CI **`msrv`** job that actually builds published crates at the Rust versions they declare, instead of leaving stale **1.80** promises unchecked.
> 
> **Manifest updates:** KDL-free crates (`usage-argv`, `usage-derive`, `usage-config`) move to **1.91**; KDL-backed crates (`usage-lib`, `usage-config-build`, `clap_usage`, `usage-cli`) use **1.95**. **`clap_usage`** and **`usage-cli`** gain a `rust-version` for the first time. Unpublished **`usage-conformance`** and **`xtask`** drop `rust-version` so they are not treated as MSRV promises to dependents.
> 
> The workflow runs a **two-row matrix** (1.91 vs 1.95), bidirectional checks that each row’s crate list matches every published manifest at that version, then **`cargo check --locked --all-features`** on the pinned toolchain (library surface only, not dev-deps).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b35b1ec9810557e9b61e08401c8b27c327652c92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,15 +93,39 @@ jobs:
           persist-credentials: false
       - name: the matrix and the manifests agree
         run: |
-          # Or the job checks a version nobody promised. A crate that raises its floor without
-          # touching this file would otherwise be tested at the old one and pass.
+          # Both directions, because one of them alone is how a false floor survives.
+          #
+          # Every crate this row checks declares this row's version. Or the job checks a version
+          # nobody promised: a crate that raises its floor without touching this file would
+          # otherwise be tested at the old one and pass.
           for crate in ${{ matrix.crates }}; do
-            dir="$(grep -rl "^name = \"$crate\"" ./*/Cargo.toml | head -1)"
-            declared="$(grep '^rust-version' "$dir" | sed 's/.*"\(.*\)".*/\1/')"
+            manifest="$(git ls-files '*Cargo.toml' | xargs grep -l "^name = \"$crate\"$" | head -1)"
+            declared="$(sed -n 's/^rust-version *= *"\(.*\)".*/\1/p' "$manifest" | head -1)"
             if [ "$declared" != "${{ matrix.version }}" ]; then
               echo "::error::$crate declares $declared, checked here at ${{ matrix.version }}"
               exit 1
             fi
+          done
+          # And every published crate declaring this row's version is in this row. Without this
+          # half, a new crate — or one that was simply never listed — makes a floor promise that
+          # no toolchain ever tries. `usage-conformance` and `xtask` promised 1.91 for as long as
+          # this job existed while depending on a crate that needs 1.95; nothing noticed, because
+          # nothing looked from the manifests back to the matrix.
+          #
+          # Published only: a floor is a promise to whoever depends on the crate, so a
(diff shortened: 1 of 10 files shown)


jdx:
Valid, and fixed in `d730d40`.

`usage-conformance` and `xtask` did declare 1.91 while depending on `usage-lib` (1.95), and conformance on `kdl` as well. The floor was unreachable on both.

Removed rather than raised. `rust-version` is a promise to whoever depends on the crate, and both are `publish = false` — nobody can. That also matches `gate` and the shadow crates, which already declare none.

The more useful half is why nothing caught it. The check only read in one direction: matrix → manifest, looking up each listed crate and comparing. That catches a crate that raises its floor without updating the matrix, and it cannot, by construction, catch a crate the matrix never mentions — which is what these two were. It now reads back the other way too: every *published* crate declaring this row's version must appear in this row.

Verified against both rows, and mutation-checked by restoring the original state (`rust-version = "1.91"` on `xtask`, published):

```
CAUGHT: xtask declares 1.91, but no row checks it
```

PR #923, opened 2026-08-16
feat(derive): a command that takes nothing can be written that way

jdx:
The last of the communique follow-ups. clap lets a command with no options be a bare variant; the derive wanted three things to say the same:

```rust
// before
#[derive(Args)] struct Sponsors {}
enum Command { Sponsors(Box<Sponsors>) }
// ...and at the call site, because `_` leaves the payload unread and dead_code is an error in CI:
Command::Sponsors(args) => { let cli::Sponsors {} = *args; sponsors() }

// after
enum Command { #[usage(effect = "read")] Sponsors }
Command::Sponsors => sponsors(),
```

Everything generated speaks to a struct — its tables, its metadata, its `build` — so rather than teach all of that about a variant with no struct, **the derive writes the empty struct such a variant implies** and points the variant at it. Only the construction differs.

The struct is named after the enum as well as the variant: two enums in one module may each declare a `Sponsors`, and two structs of one name is a worse error than the one being avoided — it would land in the adopter's crate, pointing at something they never wrote. A test has two such enums.

## `effect` moves with it

This is the part I did not see coming. `effect` is an `Args` attribute, because that is where a command declares itself — and a bare variant has none. Porting communique's `sponsors` over dropped `effect=read` from its spec, silently. Its in-sync test caught it, which is the second time that test has paid for itself.

So a variant carries `effect` when it holds nothing, and the written struct receives it as the same attribute, read by the same code. **Refused on a variant that does wrap a struct**: two places to say one thing is two answers waiting to disagree.

## What communique deletes

```
 src/cli.rs  | 7 +------
 src/main.rs | 8 +-------
```

## Verification

| mutation | result |
|---|---|
| a bare variant still builds with a payload | generated code fails to compile |
| the implied struct is not written | generated code fails to compile |
| the struct name omits the enum | generated code fails to compile (two enums, one name) |
| `effect` on a struct-wrapping variant is accepted | FAILED |

The third **survived at first** — my fixture had only one enum, so nothing collided. Seven conformance tests, workspace suite green, clippy clean, gate green over mise's 211 commands.

---

> [!NOTE]
> **Medium Risk**
> Touches core `Subcommands` codegen and parsing/build paths for all unit variants; mistakes could break spec/help/effect metadata or cause name collisions, though coverage is broad in new conformance tests.
> 
> **Overview**
> **Subcommands can be written as bare unit variants** (e.g. `Sponsors,`) instead of `Sponsors(Box<Sponsors>)` plus a hand-written empty `Args` struct. The derive emits a hidden `#[derive(Args)]` empty struct per unit variant so parsing, spec, help, and `build` still go through the same `CommandArgs` path; only enum construction changes (`Command::Sponsors` vs wrapping a payload).
> 
> **`#[usage(effect = "...")]` on the variant** is allowed when the variant holds nothing (forwarded onto the generated struct). Putting `effect` on a variant that wraps a struct is a **compile error**.
> 
> **Naming hardening:** implied structs use length-prefixed `__Usage{n}{Enum}{Variant}` ids so two enums in one module can both have `Sponsors`; variant ids use `unraw()` so `r#type` becomes command name `type` and macro id generation does not panic.
> 
> Adds conformance tests in `unit_variants.rs` and crate docs for the pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99f2622ead3830fac47effaa52cccd25f3b12295. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/conformance/tests/unit_variants.rs
+++ b/conformance/tests/unit_variants.rs
@@ -1,0 +1,168 @@
+//! A command that takes nothing.
+//!
+//! clap lets one be a bare variant — `Sponsors,` — and the derive needed
+//! `Sponsors(Box<Sponsors>)` plus an empty struct plus, at the call site, a destructure to
+//! keep the payload from reading as dead code. communique carries all three today.
+//!
+//! Everything generated speaks to a struct — its tables, its metadata, its `build` — so rather
+//! than teach all of that about a variant with no struct, the derive writes the empty struct
+//! such a variant implies. Only the construction differs.
+
+use usage::Spec as LibSpec;
+use usage_argv::help;
+use usage_derive::{Args, Cli, Subcommands};
+
+/// Set things up
+#[derive(Args)]
+struct Init {
+    /// Overwrite what is there
+    #[usage(long)]
+    force: bool,
+}
+
+#[derive(Subcommands)]
+enum Command {
+    /// Set things up
+    Init(Box<Init>),
+    /// Show who pays for this
+    #[usage(effect = "read")]
+    Sponsors,
+    /// Kept out of help, and still a command
+    #[usage(hide)]
+    Secret,
+    /// Answers to another name too
+    #[usage(name = "licence", alias = "license")]
+    Licence,
+}
+
+/// A second set, with a bare variant spelled the same as the first's
+///
+/// The implied structs are named after the enum as well as the variant for exactly this: two
+/// of one name in a module is a worse error than the one being avoided, and it would land in
+/// the adopter's crate pointing at a struct they never wrote.
+#[derive(Subcommands)]
+enum Other {
+    /// A different command that happens to share a name
+    Sponsors,
+}
+
+/// A second tool sharing the module
+#[derive(Cli)]
+#[usage(bin = "other")]
+struct OtherCli {
+    #[usage(subcommand)]
+    command: Option<Other>,
+}
+
+/// A tool with commands that take nothing
+#[derive(Cli)]
+#[usage(bin = "ex")]
+struct Ex {
+    #[usage(subcommand)]
+    command: Option<Command>,
+}
+
+#[test]
+fn a_bare_variant_is_a_command() {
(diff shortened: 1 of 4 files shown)


jdx:
Triaged all three; none needs a change here, but the second one surfaced a real limitation worth writing down.

**Synthesized identifiers collide and panic** — neither half holds. Raw identifiers are unraw'd before the name is built (`enum_ident.unraw()` / `variant.unraw()`), so `r#type` never reaches `syn::Ident::new` with its prefix. And the concatenation is not separator-free: it carries the enum name's length, so the pair you name goes to `__Usage3FooBarBaz` and `__Usage6FooBarBaz`. Verified by compiling a CLI with a `r#type` bare variant alongside a normal one.

**Public generated unit structs** — the premise is not right: `pub` items are legal in function scope. `pub struct`, `pub mod`, and `pub fn` in an `impl` all compile inside a `fn` body; rustc emits no error and only an unused warning.

Underneath it, though, there *is* a real failure — a `Subcommands` enum declared in a function body does not compile, because the generated `mod` refers to its types through `super`, and `super` from a module nested in a function body is the enclosing *module*, not the body:

```
error[E0425]: cannot find type `__Usage5LocalInner` in module `super`
error[E0425]: cannot find type `Local` in module `super`
```

Not a regression from this PR, and not about unit variants: the identical enum written with a wrapping variant and no bare variant anywhere fails the same way, with the same two errors. It is a pre-existing limitation of the `mod`/`super` emission. Filing it separately rather than expanding this PR's scope.

**Generated helper collides with user types** — accepted. `__Usage{len}{Enum}{Variant}` is `__`-prefixed, length-disambiguated against sibling enums, and would require an adopter to declare that exact name at module scope. Worth a note in the derive docs rather than a hygiene mechanism.

PR #925, opened 2026-08-16
feat(spec): say that one flag needs another, which nothing here could

jdx:
Stacked on #919 — review that first; this PR's diff is the second commit.

The spec could say `conflicts`, `overrides`, `required_if` and `required_unless`, and had no way to say the positive form. `requires` is the last blocker in PLAN.md's derive-v1 list.

```kdl
flag "--out <path>" requires="--format"
flag "--sign" {
  requires "--key" "--identity"   // several, and all of them are needed
}
```

```rust
#[usage(long, requires("--key", "--identity"))]
sign: bool,
```

### Why not just `required_if`

It states the same rule from the other end, and both stay. `--out` needing `--format` is `requires="--format"` on `--out`, or `required_if="--out"` on `--format`. Only the first keeps the rule beside the flag it is about, which is usually where a reader looks for it — and it is the spelling anyone arriving from clap already knows.

### Wired the whole way through

A constraint only one parser enforces is worse than none, so: the KDL node in both spellings (property for one selector, child node for several), usage-lib's check, the derive attribute with its selectors resolved at **compile time**, usage-argv's cold metadata, and the emitted KDL that docs, manpages and completions read.

An unmet requirement is reported as the **other flag missing** rather than as something wrong with the flag that named it — which is what clap says, and what a user can act on: the fix is to type `--key`, not to delete `--sign`. That also means no new `Error` variant, so the hot path's `Result` does not grow for a check that only ever fires on the cold one.

A value from the environment or a default satisfies a requirement, matching `conflicts`: the question is whether the other flag ended up with a value, not how it got one.

### The bridge does not carry it

`clap_usage` reads a `clap::Command` back through public getters, and clap 4.6 has `Arg::requires`, `requires_if`, `requires_ifs` and `requires_all` as setters with no reader, with the field `pub(crate)`. So there is nothing to read, whatever the `Arg` was built with.

`requires_cannot_come_across_from_clap` asserts that, rather than leaving the empty vector looking like a bug in the bridge — and it will fail if a future clap exposes a getter, which is when we would want to know.

`gen-shadow` counts `requires` against the clap dialect for a related reason: clap *can be written* with `requires`, so the shadow could emit it, and a spec regenerated from that command would come back without it. Counting it keeps the clap side from looking more faithful than it is.

`docs/spec/integrations/clap.md` now says this out loud, since a CLI generating its spec from clap silently loses the constraint everywhere downstream.

### Tests

- `requires_round_trips_in_both_spellings` — property and child-node forms, through `to_string()` and back
- `requires_cannot_come_across_from_clap` — the bridge limit, asserted
- `a_requirement_names_the_flag_that_is_missing` / `a_requirement_is_satisfied_by_a_short_form` — usage-lib, including a selector spelled `-k` resolving to `--key`
- `a_requirement_names_the_flag_that_was_not_given` — the derive, plus the absent-flag case that makes this different from plain required-ness
- `flag_relationships_reach_the_spec` — the emitted KDL takes the child-node spelling for two selectors, and usage-lib parses it back

`cargo test --all --all-features` and `cargo clippy --all --all-features -- -D warnings` are clean.

---

> [!NOTE]
> **Medium Risk**
> Touches post-parse validation in usage-lib and derive-generated checks; behavior is well-tested but wrong satisfaction rules (defaults/env/overrides) would change CLI acceptance for derived binaries.
> 
> **Overview**
> Adds **`requires`** so a flag can declare that other flags must be present when it is used — the positive counterpart to **`conflicts`** and the mirror of **`required_if`** on the imposing flag.
> 
> **Spec & metadata:** `requires` is on `FlagMeta` / `SpecFlag`, parsed and emitted in KDL (single property or child node for multiple selectors), with builder helpers and argv spec serialization.
> 
> **Parsing:** usage-lib validates requirements after binding (alongside conflicts), using **`selector_is_satisfied`** so defaults and env values count; failures report the **missing target flag** (`MissingFlag` / derive `MissingRequired`), not the flag that declared the rule. The derive emits post-parse checks and skips targets that have defaults.
> 
> **Derive:** `#[usage(requires = "--other")]` with compile-time selector validation; flag-only, like other inter-flag relationships.
> 
> **Clap gap:** Generated clap specs cannot round-trip `requires` (no getter); docs and `gen-shadow` document/count this instead of silently emitting it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89866b8b37ba0f2060b2d788a362067f0a58e1af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -462,6 +462,12 @@ pub struct FlagMeta<'a> {
     /// flag win, this reports it: the combination has no meaning, so honouring one
     /// side silently would hide a mistake.
     pub conflicts: &'a [&'a str],
+    /// Flags that must also be given when this one is.
+    ///
+    /// The positive form of [`conflicts`](Self::conflicts), and the mirror image of
+    /// [`required_if`](Self::required_if): the same rule written on the flag that
+    /// imposes it rather than on the flag it lands on.
+    pub requires: &'a [&'a str],
     /// Flags that make this one necessary.
     pub required_if: &'a [&'a str],
     /// Flags that make this one unnecessary.
@@ -491,6 +497,7 @@ impl FlagMeta<'_> {
         var_max: None,
         overrides: &[],
         conflicts: &[],
+        requires: &[],
         required_if: &[],
         required_unless: &[],
         help_heading: None,
@@ -911,6 +918,7 @@ fn write_flag(out: &mut String, meta: &FlagMeta<'_>, depth: usize) -> core::fmt:
     write_single_default(out, meta.default)?;
     write_single_list(out, "overrides", meta.overrides)?;
     write_single_list(out, "conflicts", meta.conflicts)?;
+    write_single_list(out, "requires", meta.requires)?;
     write_single_list(out, "required_if", meta.required_if)?;
     write_single_list(out, "required_unless", meta.required_unless)?;
 
@@ -920,6 +928,7 @@ fn write_flag(out: &mut String, meta: &FlagMeta<'_>, depth: usize) -> core::fmt:
         || meta.default.len() > 1
         || meta.overrides.len() > 1
         || meta.conflicts.len() > 1
+        || meta.requires.len() > 1
         || meta.required_if.len() > 1
         || meta.required_unless.len() > 1;
     if !has_children {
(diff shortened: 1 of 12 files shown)


jdx:
Pushed fixes for both findings.

**Defaults failing requirement checks** (@greptile-apps P1) — correct, and worse than a doc mismatch: `required`, `required_if` and `required_unless` in the same file all read a default as filling a flag, so the same flag was missing in this check and present ten lines below it.

The rule, written down so the next relationship inherits it rather than picking again:

- A **negative** rule — `conflicts` — counts only what was *given*. It has to: a flag with a default would otherwise conflict with everything and no command line would parse.
- A **positive** rule asks whether the flag it names has a *value*, and a default is a value.

usage-lib gets `selector_is_satisfied` beside `selector_is_explicit`. The derive skips emitting the check entirely when the target has a default, since it could never fail — decided at compile time, the same way `required_checks` already skips such a field.

**The lint failure** — the conformance fixture declared `--key` and `--identity` and never read them, which `clippy --all-targets` fails on. They are asserted now, which is what they were for.

PR #927, opened 2026-08-16
feat(spec)!: a group, for the rule that no single flag can state

jdx:
Third in the stack, on top of #925. Review #919 → #925 first; this PR's diff is the third commit.

```kdl
flag "--file <file>"
flag "--url <url>"
flag "--stdin"

group "input" "--file" "--url" "--stdin"                 // at most one
group "input" "--file" "--url" "--stdin" required=#true  // exactly one
group "input" "--file" "--url" "--stdin" required=#true multiple=#true  // at least one
```

Name first, then members, spelled the way every other relationship names a flag. The two properties are clap's and read the same way, so a spec generated from a clap command means by them what clap meant.

### Why a group and not more `conflicts`

Pairwise `conflicts` says "at most one of these" — once per pair. Three flags is three declarations, six is fifteen, and a flag added later has to be added to every sibling.

What `conflicts` cannot say at all is `required`. "One of these is needed" is a statement about the *set*: no rule written on an individual flag expresses it, because no individual flag is the one that must be given.

### The bridge carries this one

Unlike `requires` in #925, `Command::get_groups`, `ArgGroup::get_args` and `is_required_set` are public, so a clap CLI's groups now reach the spec. **Every spec in the fleet generated from a clap command was dropping them** — the same hole #884 closed for `conflicts`, found the same way.

`ArgGroup::is_multiple` takes `&mut self` and a `Command` only hands out `&ArgGroup`, so the group is cloned to ask. Once per group at spec-generation time, which is a build step.

### Decisions worth a look

- **Enforced across the command chain**, not just the selected command: a group may name global flags, which are declared on an ancestor.
- **A default does not count as given.** Env does. This matches `conflicts` and the flag reference's existing wording. A default that satisfied a required group would make the group unfalsifiable, and one that collided with a typed sibling would refuse a command line where the user named exactly one flag.
- **Two members is the floor**, checked where the group is written rather than silently enforcing nothing at run time. A group of one is a rule about that flag and belongs on the flag.
- **`UsageErr` gains a `MissingGroup` variant**, and the enum is marked `#[non_exhaustive]` so this is the last break of its kind — `MissingFlag`'s `Display` is `Missing required flag: --{0} <{0}>`, which mangles a group. Versions stay untouched; release-plz decides what `feat(lib)!:` implies, which does mean `lint:semver` is red on this PR by construction (see the comment below).
- **Positional members of a clap group are dropped.** The spec names relationships by flag; a selector matching nothing would read as a rule that holds and enforce even less than dropping it.

### Not in this PR

The derive cannot declare a group yet — spec first, per the canonicality rule. That is the next branch, along with usage-argv's metadata and emission.

`cargo test --all --all-features` and `cargo clippy --all --all-features -- -D warnings` are clean. New reference page at `docs/spec/reference/group.md`, linked from the sidebar and from the `conflicts` section.

---

> [!NOTE]
> **Medium Risk**
> Touches core argv validation and adds a breaking `UsageErr` API change, but behavior is covered by new tests and mirrors existing conflict/require semantics.
> 
> **Overview**
> Introduces a **`group`** KDL construct so related flags can be described as a set—mutual exclusion (`at most one`), **`required`** (`exactly one` / `at least one` with **`multiple`**), and **`multiple`** alone—instead of scaling pairwise **`conflicts`** and having no way to express “pick one of these.”
> 
> The library adds **`SpecGroup`**, wires **`group`** into command/spec KDL parse and serialize, and validates groups during **`parse`** across the full command chain (for globals on ancestors). Exclusivity uses the same “explicitly given” rule as **`conflicts`** (env counts, defaults do not); requiredness uses **`selector_is_satisfied`** (defaults count). Members dedupe by resolved flag name. Mount/merge replaces groups when root flags are replaced. Clap **`ArgGroup`** export skips bookkeeping groups (`multiple` without `required`) and groups with fewer than two flag members.
> 
> **`UsageErr`** gains **`MissingGroup`** and is marked **`#[non_exhaustive]`** (breaking for exhaustive matches). Docs add **`group.md`**, sidebar link, and a **`flag`** cross-reference. **`lint:semver`** is temporarily skipped for usage-lib below 6.x until the declared major ships.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dcfa48646a221e687a245e0129c306574945e94c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added support for grouping related flags.
  - Groups can enforce mutually exclusive options, require one or more selections, and allow multiple selections where appropriate.
  - Group definitions are supported in specifications and compatible command configurations.

- **Bug Fixes**
  - Improved validation for aliases, environment-provided values, defaults, and positional arguments in flag groups.
  - Added clearer errors when required groups are not satisfied.

- **Documentation**
  - Added comprehensive group reference documentation and navigation.
  - Updated flag guidance with recommendations for using groups.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -83,6 +83,7 @@ export default defineConfig({
               { text: "cmd", link: "/spec/reference/cmd" },
               { text: "complete", link: "/spec/reference/complete" },
               { text: "flag", link: "/spec/reference/flag" },
+              { text: "group", link: "/spec/reference/group" },
               // { text: 'env', link: '/spec/reference/env' },
               { text: "config", link: "/spec/reference/config" }
             ]
--- a/docs/spec/reference/flag.md
+++ b/docs/spec/reference/flag.md
@@ -74,6 +74,9 @@ mistake, and silently honouring one of them hides it.
 A conflict holds in either direction, so declaring it once is enough; it applies to flags
 that were actually given, not to defaults.
 
+For three or more flags that exclude each other, or for "one of these is required", a
+[`group`](/spec/reference/group) says in one node what `conflicts` says once per pair.
+
 ## `requires`
 
 The positive form: giving this flag means the flags it names must be given too. Every
--- a/docs/spec/reference/group.md
+++ b/docs/spec/reference/group.md
@@ -1,0 +1,76 @@
+# `group`
+
+A set of flags that relate to one another as a set.
+
+```kdl
+flag "--file <file>"
+flag "--url <url>"
+flag "--stdin"
+
+// at most one of these
+group "input" "--file" "--url" "--stdin"
+
+// exactly one of these: one is required, and only one is allowed
+group "input" "--file" "--url" "--stdin" required=#true
+
+// at least one of these
+group "input" "--file" "--url" "--stdin" required=#true multiple=#true
+
+// the same, written out when the members do not sit comfortably on one line
+group "input" required=#true {
+  flag "--file" "--url"
+  flag "--stdin"
+}
+```
+
+The name comes first, then the members, spelled the way every other relationship names
+a flag — `--long` or `-s`. A group lives on the command whose flags it names, and a
+group naming [global flags](/spec/reference/flag#global) belongs to the command that
+declares them.
+
+## What the two properties mean
+
+| `required` | `multiple` | meaning               |
+| ---------- | ---------- | --------------------- |
+| —          | —          | at most one of these  |
+| `#true`    | —          | exactly one of these  |
+| —          | `#true`    | nothing is enforced   |
+| `#true`    | `#true`    | at least one of these |
+
+Both default to false, so a bare group is mutual exclusion. These are clap's two
+properties read the same way, so a spec generated from a clap command means by them what
+clap meant.
+
+## Why not just `conflicts`
+
+For "at most one", [`conflicts`](/spec/reference/flag#conflicts-and-overrides) says the
+same thing — but it says it once per pair, so three flags need three declarations and six
+need fifteen, and a flag added later has to be added to every sibling.
+
+The part `conflicts` cannot say at all is `required`. "One of these is needed" is a
+statement about the set: no rule written on an individual flag expresses it, because no
+individual flag is the one that must be given.
+
+## What counts as given
+
+A member is given when it ended up with a value from the command line or from its
+[`env`](/spec/reference/flag) variable — the same rule
+[`conflicts`](/spec/reference/flag#conflicts-and-overrides) and
+[`requires`](/spec/reference/flag#requires) follow.
+
+A [`default`](/spec/reference/flag) does not count. A default that satisfied a required
+group would make the group unfalsifiable — it could never report anything — and one that
+collided with a typed sibling would refuse a command line where the user named exactly
+one flag.
+
+## Members
+
+A group needs at least two members, and naming fewer is an error where it is written
+rather than a rule that quietly enforces nothing. A group of one is a statement about
+that flag, which belongs on the flag as [`required`](/spec/reference/flag) or
+[`requires`](/spec/reference/flag#requires).
+
+Only flags can be members. clap allows a positional in a group; a spec generated from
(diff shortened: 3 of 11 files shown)


jdx:
Pushed fixes for everything raised, plus a decision that changes this PR's shape.

**The group clap invents for every struct** — `clap_derive` builds `ArgGroup::new(<struct name>).multiple(true)` for each `#[derive(Args)]` type so `flatten` works ([`args.rs`](https://docs.rs/clap_derive/4.6.4/src/clap_derive/derives/args.rs.html)). Carrying those put a `group Lint …` into usage's *own* generated spec, which is what the render check was failing on. Dropped by what they are rather than where they came from: `multiple` without `required` states no rule at all, so there is nothing to carry. `multiple` *with* `required` still comes across.

**Aliases inflating the member count** (@greptile-apps P1, @cursor Bugbot) — members are counted by the flag a selector resolves to, not by the selector, so a group naming both `-f` and `--file` holds one member and a single flag is never in conflict with itself. Deduplicated rather than refused where the group is written, since listing both spellings is redundant rather than wrong.

**Defaults** — corrected, and the correction came from the `requires` review one PR down. The two halves of a group are two kinds of rule: exclusivity counts what was *supplied*, or a defaulted member would collide with the sibling the user typed; requiredness asks whether a member ended up with a *value*, which is what plain `required` and `requires` already do. Docs say so now.

### This is now a breaking release

`UsageErr` is marked `#[non_exhaustive]` and the shared-version crates go to **6.0.0**, following the shape of #742 — the PR that broke the spec took 4.0.0 rather than leaving a release commit to discover it.

The reasoning: `MissingGroup` is the third variant added since 4.0, and each was a major for everyone matching on `UsageErr`. This enum grows whenever the spec learns to say something new, so one more break now buys none of the same kind later. Callers need a `_` arm from here on.

Happy to split the bump into its own PR if you would rather it not ride along with the feature.

jdx:
Rebased onto `main` now that #919 and #925 have merged, and the version bump is **gone** — versions are release-plz's to decide, and nothing here touches one. The `feat(lib)!:` marker is the whole declaration.

The consequence, flagged rather than worked around: **`lint:semver` on this PR is red on purpose.** `mise run lint` includes `cargo semver-checks check-release -p usage-lib`, which compares against the published crate, so any deliberate break is red until release-plz bumps after merge. The gate structurally cannot pass before the release it is waiting for.

What it reports is exactly the intended change and nothing else:

```
--- failure enum_marked_non_exhaustive: enum marked #[non_exhaustive] ---
  enum UsageErr in lib/src/error.rs:12
Summary semver requires new major version: 1 major and 0 minor checks failed
```

Every other check is green. Also fixed here since the last review round: a group no longer outlives the flags it names, so a mount that replaces a command's flags replaces its groups with them — the case that was broken was the empty one, where a non-empty check let the stale set survive precisely when there was nothing to replace it with.

jdx:
Took the suggestion on the gate rather than leaving it red.

`lint:semver` now skips while usage-lib is on 5.x and runs normally from 6.0 on. The reasoning is that a gate which *cannot* pass is one people learn to ignore, and an ignored gate stops catching the accidental breaks it exists for — whereas this one is deliberate and already declared with `feat(lib)!:`.

```toml
major=$(sed -n 's/^version = "\([0-9]*\)\..*/\1/p' lib/Cargo.toml | head -1)
if [ "$major" -lt 6 ]; then
  echo "skipped: usage-lib is $major.x, and the 6.0 break this would report is declared rather than accidental"
  exit 0
fi
cargo semver-checks check-release -p usage-lib
```

The condition clears itself, which is the part that matters: once release-plz ships 6.0.0 the break is in the published baseline, the check runs again, and nobody has to remember to come back and delete anything.

Both halves are verified locally rather than assumed — it skips at 5.1.0, and with the crate temporarily set to 6.0.0 it runs and reports `usage-lib v5.1.0 -> v6.0.0 (major change) … no semver update required`.

PR #928, opened 2026-08-16
fix(help): render the page for the mount the words reached

jdx:
A `Subcommands` type mounted under two parents is one `Command` at one
address, so `help::find` — which located a command by `core::ptr::eq` over
the whole tree — returned whichever mount came first. `ex beta shared
--help` printed `Usage: ex alpha shared`, with alpha's globals, for as long
as this crate has had help. Diagnostics fixed their half of this by
resolving the route rather than the command; help kept the pointer.

A `&Command` cannot distinguish the mounts, so it cannot be the input.
`route_to` builds the route from the argv the derive already has: it takes
the path the parser walked, and for the `help beta shared` spelling — where
the parse stops at the command that *saw* the word, short of the one being
asked about — walks the remaining words by name among the children of
wherever it has got to. By name, because the address is the thing that
cannot be trusted here. `render_at` then resolves the metadata chain the
same way, matching each step among that command's own children.

`parse()` prefers the route and falls back to `render`, so a caller that
reaches help by some other path still gets a page.

Both spellings are tested, and so is the first mount — resolving by route
must not make every page beta's.

---

---

> [!NOTE]
> **Medium Risk**
> Touches help rendering and generated `parse()` exit paths; behavior change is intentional for shared mounts, with fallback preserving old `render` when routing fails.
> 
> **Overview**
> **Fixes wrong help for shared subcommands mounted under multiple parents.** When the same `Subcommands` type is wired under two parents, both mounts share one `Command` pointer, so `help::render` / `find` picked the first mount—e.g. `ex beta shared --help` showed `Usage: ex alpha shared` and alpha’s globals.
> 
> The parser records **`help_span`**: the `argv` slice where the `help` *word* walked subcommand names (not flag values). **`help::route_to`** re-parses argv, builds the command path from `command_path` + that span, and **`help::render_at`** builds usage/metadata from that route by matching children per parent. Generated **`parse()`** uses route + `render_at` on `Error::Help`, falling back to **`render`** when the route can’t be rebuilt.
> 
> Conformance tests cover `--help`, `help beta shared`, both mounts, and **`--config <name>`** so a global flag value isn’t mistaken for a route step.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e00424ac9e2d09b02ee77d6a4ac98a9a7097a8e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -1189,8 +1189,9 @@ pub fn render(spec: &Spec<'_>, cmd: &Command<'_>, long: bool) -> Option<String> 
 /// The parse is deterministic, so walking the same argv reaches the same place.
 ///
 /// `ex help config set` asks about a command *deeper* than the parse reached, so the route is
-/// extended by matching each remaining word among the children of where it got to — the same
-/// descent the parser made, and unambiguous at every step.
+/// extended over [`Parser::help_span`](crate::Parser::help_span) — the words the parser itself
+/// resolved as a command path, which is the only reading that cannot mistake a flag's value for
+/// a command name.
 ///
 /// `None` where the command is not below this spec at all, which a caller should treat as a
 /// reason to fall back rather than a failure.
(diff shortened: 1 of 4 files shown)


jdx:
Both of you were right, and I reproduced it before fixing:

```
["--config", "alpha", "help", "beta", "shared"]   Usage: ex alpha shared   ["--alphaglobal"]
```

The arrival check did not catch it for the reason Bugbot gives: both mounts are one address, so `ptr::eq` succeeded on the wrong one. That is the same property this PR exists to work around, arriving through a different door.

Fixed in `d67156e`. The scan was the wrong source: only the parser knows which tokens were in command position. It already resolves the words after `help` itself — one subcommand at a time, breaking at the first that names none — and was throwing that walk away. It now records the `argv` range it covered, and `route_to` extends over exactly that range instead of over raw `argv`. Two indices rather than the commands themselves, so the parser goes on allocating nothing; the span is empty for `--help`, which asks about wherever the parse reached.

Tested in both directions — `--config alpha help beta shared` and `--config beta help alpha shared` — so a fix that merely preferred the later match would not pass. Mutation-checked by restoring the scan, which fails the new test with `Usage: ex alpha shared`.

PR #936, opened 2026-08-17
feat(cli): parse usage's own command line with the parser usage ships

jdx:
Stacked in this order: #937 → #939 → #946 → #945 → #947 → #949 → #951 → this PR.

`usage` is now its own first adopter. Its command structs, root, and command enums use the `usage-rs` facade instead of Clap, and `--usage-spec` prints `Cli::to_kdl()` from the same tables that parsed the command line.

## What goes away

- `clap`, `clap_usage`, and the `clap-sort` dev dependency
- `tests/clap_sort.rs`; declaration order is held by the spec
- the duplicated command-effect tables; effects now live on the commands they describe
- the empty `Sponsors` struct; unit subcommands are supported directly
- runtime checks for relationships the spec can express

## Migration shape

The declarations remain close to their former Clap layout:

- inferred shorts use `#[usage(short, long)]`
- command aliases live on their command structs, e.g. `#[usage(alias = "c", alias_hidden("complete", "completions"))]`
- positive dependencies use `requires`, including `--cache-key` → `--usage-cmd` and `--out-dir` → `--multi`
- aliases written on enum variants still merge with struct aliases for compatibility
- deliberately formatted docs keep `#[usage(verbatim_doc_comment)]`
- path hints keep `value_hint = usage_rs::ValueHint::FilePath` or `DirPath` from usage’s own runtime type
- subcommand payloads stay unboxed as they were under Clap; `Box<T>` remains an optional size optimization

The four shell commands are separate derived structs flattening a shared group. A single destination struct cannot identify which enum variant selected it, so a macro keeps their common declaration in one place.

## Parity gaps closed by the stack

- required subcommands survive KDL generation (#937)
- strict unknown-flag behavior is inherited by subcommands (#939)
- command-owned visible and hidden aliases are supported (#946)
- duplicate non-repeatable flags are rejected (#945)
- a missing required subcommand prints the available command choices (#947)
- verbatim doc-comment layout is preserved on commands, fields, and variants (#949)
- file and directory value hints reach native and emitted completions (#951)

The emitted spec still intentionally gains metadata the Clap bridge dropped, including `JDX_USAGE_BIN` on `--usage-bin`, preserved multiline long help, the manpage description, and the correct binary name `usage`.

## Verification

- `mise run render`
- `cargo test --all --all-features`
- `cargo clippy --all --all-features --all-targets -- -D warnings`
- direct binary checks for no-argument help, duplicate flags, command aliases, and positive requirements

_This PR was generated by Codex._

---

[...]

PR #937, opened 2026-08-17
feat(derive): say that a command cannot be run alone, which it knew and did not write

jdx:
Stacked on #923 — review that first; this PR's diff is the last commit. One of the two gaps #936 found by converting usage-cli itself to the derive; the other one gets its own PR.

A bare `T` subcommand field requires a subcommand and an `Option<T>` does not. The parser has always refused the invocation accordingly — that half was never missing — but `Spec::to_kdl` wrote nothing, so the emitted KDL described `usage generate` as a command a user could type on its own.

That reaches past help text: docs, manpages, completions and the SDK generators all read the emitted spec rather than the Rust type, so every one of them was offering a command that cannot run.

```kdl
cmd "settings" help="Manage settings" subcommand_required=#true {
```

## Shape

Cold metadata rather than a parse table — `subcommand_required` is not how a word binds, and the hot path never reads it. `CommandMeta` gains a `bool`, the derive fills it from the field's type, and the writer emits it **only where there is something to require**: the property on a command with no subcommands is a spec the linter reports and nothing could satisfy, so both conditions are held.

`gen-shadow` already read `subcommand_required` from a KDL and chose `T` against `Option<T>` accordingly, so mise's shadow round-trips it now instead of losing it on the way back out.

## Verification

- `conformance/tests/subcommand_required.rs` — four tests: a strict command says it, a loose one stays quiet *and* reads back as `false`, a leaf never claims it, and the parser and the spec agree about which invocations exist.
- `nesting`'s snapshot picks the change up on a fixture that was already shaped this way, which is the check that this is emission rather than a new rule.
- Workspace suite green, clippy clean.

## Not fixed here

The spec accepts `subcommand_required` only inside a `cmd` block, so a **root** that cannot run alone still has no way to say so — the same restriction as the root mount, recorded next to it in PLAN.md. The clap bridge could not say it either, so nothing regressed.

---

> [!NOTE]
> **Low Risk**
> Cold-path metadata and KDL emission only; parsing behavior was already correct, with added compile-time validation for an invalid flatten shape.
> 
> **Overview**
> **Emits `subcommand_required` in specs** so docs, completions, and generators match parser behavior: a non-optional `#[usage(subcommand)]` field sets cold metadata and writes `subcommand_required=#true` only when the command actually has subcommands; optional fields stay omitted and read back as false.
> 
> **Compile-time guard** via `flatten_checks`: a `#[usage(flatten)]` type that still declares subcommands fails const evaluation, because flatten would leave the parent unable to express required subcommands while the nested `build` still demands one.
> 
> **Tests and docs**: new `conformance/tests/subcommand_required.rs`, updated nesting snapshot (`settings` gains the flag), and PLAN.md marks the derive/spec gap done while noting root-level `subcommand_required` remains unsupported in KDL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ca7f41e074529cf0ec100aebdff11bba1bd66ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Command specifications now identify when a subcommand is required.
  * Help, documentation, completions, and generated specifications consistently reflect required versus optional subcommands.
  * Nested required-subcommand behavior is supported.

* **Bug Fixes**
  * Improved parsing and specification output for commands with required or optional subcommands.

* **Documentation**
  * Documented the remaining limitation that root-level required subcommands cannot currently be represented.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -257,6 +257,22 @@ tasks --usage"`, so task names are meant to come from running that. usage-argv d
       `cmd` block, so a CLI whose _top-level_ subcommands are discovered by running
       something cannot say so. Worth deciding whether that is a gap or a deliberate
       restriction.
+- [x] **`subcommand_required`, which the derive knew and did not say** — a bare `T`
+      subcommand field requires a subcommand and an `Option<T>` does not, and the parser
+      has always refused the invocation accordingly. `Spec::to_kdl` wrote neither, so the
+      emitted KDL described a group command as one a user could run alone — and help, docs,
+      completions and the SDK generators all read that rather than the type. Cold metadata,
+      since it is not how a word binds. Found by converting usage-cli itself to the derive
+      and diffing the spec it prints against the clap bridge's. One shape could have made the
+      answer a lie — a `#[usage(flatten)]` group declaring subcommands of its own, which
+      flatten leaves behind while the group's `build` still demands one — and is a compile
+      error now, asserted during const evaluation in the parent's expansion, where the group
+      is only a type.
+- [ ] **`subcommand_required` on the root command** — the same restriction as the root
+      mount, and found beside it: the spec accepts the property only inside a `cmd` block,
+      so a CLI whose _root_ cannot be run alone has no way to say so. The clap bridge could
+      not say it either, so nothing regressed — but a bare `T` subcommand field on the root
+      is now a thing the derive knows and the spec has nowhere to put.
 
 - [x] **Three things a spec could say that the derive could not** — a flag's value name
       (`--tool <TOOL>` came back as `--tool <tool>`, since the flag's own name was all there
--- a/argv/src/spec.rs
+++ b/argv/src/spec.rs
@@ -386,6 +386,15 @@ pub struct CommandMeta<'a> {
     /// A token that starts a fresh invocation of this command, such as mise's
     /// `:::`.
     pub restart_token: Option<&'a str>,
+    /// Whether this command cannot be run on its own: naming it and stopping is an
+    /// error, and one of its subcommands has to follow.
+    ///
+    /// Cold metadata rather than a parse table, because it is not how a word binds —
+    /// the derive already refuses the invocation from the type, a bare `T` subcommand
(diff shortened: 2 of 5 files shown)


jdx:
Addressed the P1, and it was real — worse than reported, in fact.

A `#[usage(flatten)]` group that declares subcommands of its own does not merely report `subcommand_required=false`: the group's subcommands never reach the parent's table at all, so the command is unreachable in both directions. Reduced to a probe on the branch before this fix:

```
cmd "parent" {
}

bare parent parses: false
```

The spec says `parent` is a runnable leaf; the parser rejects it; and no word could satisfy it, because `leaf` is in no table. That is flatten's own hole rather than something this PR introduced — this PR would just have printed a confident `false` over it.

**Refused at compile time**, the way `Option<T>` flatten already is: joining two commands' subcommand sets needs a rule for which enum a word selects from, and nothing in the fleet asks for one. `flatten_checks` asserts the group's `COMMAND` has no subcommands, during **const evaluation in the parent's expansion** — the same trick `find_subcommand` uses for `default_subcommand`, and what lets one expansion check a fact about another it cannot see:

```
error[E0080]: evaluation panicked: a flattened group cannot declare subcommands: flatten
joins flags and arguments into the parent's tables and leaves subcommands behind, so the
command would require one that no word could select. Declare the `subcommand` field on the
command's own struct instead.
  --> src/main.rs:10:10
   |
10 | #[derive(Args)]
   |          ^^^^
```

It points at the parent's derive, which is where the flatten is written.

The legitimate shape is untouched — `conformance/tests/flatten.rs` has a command with both a `subcommand` field and a flattened group, and mise's ten flattens are all flags and arguments. Workspace suite green, clippy clean under `--all-targets`, gate green.

There is no compile-fail harness in this repo, so the refusal is verified by hand and documented at the top of `conformance/tests/subcommand_required.rs` and in PLAN.md, next to the entry it protects.

PR #972, opened 2026-08-17
test(gate): hold every jdx CLI to the reference, not only mise

jdx:
The parity gate read one spec, so it was blind to everything mise's CLI
happens not to do — and mise misses more than it looks like it would. Three
bugs were living in that blind spot, and none of them is exotic:

| | what mise does | what hid it |
|---|---|---|
| version banner | declares no `version` | nothing to print, so both sides agreed |
| optional flag value | has none | `<X>` vs `[X]` never came up |
| description ending in a break | none in `Commands:` | the stray blank never appeared |

Between them they changed the front page of five of the seven CLIs. Each was
found by running the comparison by hand over the fleet; this is that run, made
a test, so the next one is found by CI instead.

Seven specs, 492 commands, both help forms each. The specs are checked in for
the reason `benches/mise.usage.kdl` is: the comparison needs a fixed input and
a reviewer needs to be able to read the diff. `gen-shadow` regenerates all
seven shadows, so the CI no-diff check covers them and a change to the derive's
vocabulary has to be committed rather than discovered later.

It is a large diff — 16k lines — and almost all of it is generated: 6.9k of
vendored spec and 9k of emitted tables. The hand-written part is
`benches/gate/tests/fleet.rs` and six lines of `mise.toml`.

Mutation-checked against the bugs it exists for. Dropping `version` again:

    hk — 2 page(s) differ:  fnox — 2  pitchfork — 2  aube — 2

and un-rendering the optional bracket:

    pitchfork — 6 page(s) differ:  aube — 2

Adding a CLI here is how a new adopter's shape gets covered; see
`benches/fleet/README.md`.

---

---

> [!NOTE]
> **Low Risk**
> Test and fixture expansion with generated shadows; no production CLI or auth paths changed.
> 
> **Overview**
> Extends the parity gate so **usage-argv shadow help** is checked against **usage-lib** for seven jdx CLIs, not only mise. Checked-in `benches/fleet/*.usage.kdl` snapshots (from each tool’s `usage` subcommand) are the fixed inputs; `benches/gate/tests/fleet.rs` walks every command path and compares short and long help.
> 
> Adds six generated `shadow-*` workspace crates (hk, fnox, pitchfork, aube, tak, communique) plus bench members, wired into the gate as dev-dependencies. `benches/fleet/README.md` documents why the fleet exists—mise-only coverage missed version banners, optional flag value brackets, and trailing-break formatting that showed up on other CLIs.
> 
> The diff is mostly generated spec and shadow tables; the behavioral change is CI enforcing renderer agreement across ~492 commands and both help forms.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff91ff1902a09030c34505b10159c78ddf21a08d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Valid — fixed in `ff91ff1`, and it addresses a concern I raised on the PR myself.

Only `tests/fleet.rs` names the six fleet shadows, so a release or perf build of the gate was compiling generated crates it never calls. Moved to `[dev-dependencies]`, the same reasoning this file already records for `usage-lib` and `usage-cli`. `shadow-mise` and `shadow-mise-clap` stay in `[dependencies]` because the benchmarks are the thing that runs them.

Separately, thank you for `2f718f6` — comparing the command sets in both directions is better than what I wrote. My version walked the shadow's tree and silently skipped anything it couldn't find in the spec, so a command present in the spec but missing from the shadow passed quietly. That commit also needed a `cargo fmt`, which is what the `test` job was failing on; that's fixed and the job is green.

PR #973, opened 2026-08-17
test(spec): a rendering corpus, for the shapes the fleet cannot cover

jdx:
Rebased onto main, which changed this PR's premise twice over — see the note at the bottom. What is left is the corpus.

## What this adds

`corpus/render/` does for rendering what `corpus/` does for parsing: 41 vectors pairing a spec with the text it must produce, in the same language-neutral JSON shape, each carrying the same two-way `reference` label.

- **`01-flag-values.json`** — every pairing of the flag's brackets against its value's, defaults on the flag against defaults on the value, the variadic and repeatable ellipses
- **`02-usage-line.json`** — positional brackets, `[-- COMMAND]…`, the collapse thresholds and hidden entries not counting towards them, and the `<SUBCOMMAND>`-with-no-Commands-section oddity mise reaches on `direnv`
- **`03-sections.json`** — supplied `--help`/`--version`, annotations in both layouts, headings, inherited globals, aliases, negations, the short column, long-help wrapping, and the examples fallback

`conformance/tests/render.rs` runs them against both Rust implementations. `render-oracle` mirrors `oracle`: `--json` prints what each rendered in `expect`'s own shape, so a page goes in as a measurement rather than a transcription.

## Why, now that the gate covers seven CLIs

#972 widened the parity gate from mise alone to the whole fleet, and found three bugs doing it. That is the argument for this corpus, not against it: a fixture asks only about the vocabulary its CLIs happen to use.

Measured rather than asserted. Across all **809 value-taking flags** in mise and the fleet, three of the four flag/value bracket pairings appear and one appears nowhere:

| pairing | in the fleet |
| --- | --- |
| `[--tool <TOOL>]` optional flag, required value | 796 |
| `<--v <n>>` both required | 8 |
| `[--opt [n]]` both optional | 5 — all in pitchfork and aube |
| `<--jobs [n]>` required flag, optional value | **0** |
| a value carrying a `default`, which relaxes its brackets by another route | **0** |

The third row is what #969 fixed, and those five instances are the only reason it was visible. The last two rows have nothing but these vectors holding them.

## A divergence it found

`an-explicit-synopsis-replaces-the-root-line` is the corpus's first recorded `reference` divergence: **usage-lib's help renderer ignores a declared top-level `usage` synopsis and writes the generated line; usage-argv honours it**, which is what #965 added it for. usage-lib's *manpage* renderer honours it too (`lib/src/docs/manpage/renderer.rs:198`), so it is the help renderer that is the odd one out.

Six of the fleet's seven CLIs declare one, and the gate still cannot see this — because **`xtask gen-shadow` does not carry the node into the shadow it generates**, so every shadow's `Spec::usage` is `None` and both sides render the generated line. This corpus builds usage-argv's tables from KDL directly, so it sees the real behaviour.

Recorded rather than fixed: the expectation is right and the reference needs changing, and the two-way label check means whoever fixes it is told to delete the label. @jdx — your #965 feature, and the fix looks like a one-liner in `lib/src/docs/cli/mod.rs` if you want it here instead. The `gen-shadow` drop is a separate issue (it touches 16k lines of generated shadow).

## Out of usage-argv's scope

`disable_help` is KDL-only with no derive spelling, so no spec the derive produces carries one — which is why `lib/src/docs/cli/mod.rs` says the two renderers cannot disagree about it. This harness breaks that premise by building tables from KDL, so a vector declaring it is answered by the reference alone and skipped, the way the argv corpus skips a post-binding vector. The count of exemptions is asserted.

## Review fixes folded in

Both from Cursor Bugbot, both real, both mine rather than usage-argv's:

1. **Top-level examples dropped** — `example` at the top level hangs off `Spec` and leaves `spec.cmd.examples` empty, so every page lost its Examples section. @jdx fixed it in c53756e and pinned the root page; this adds the fallback cases, which is the half `page_examples` actually implements.
2. **`..EMPTY` in `flag_meta`/`arg_meta`** contradicting the module doc. `complete` genuinely cannot come from a spec (a Rust function vs. a shell command) and now says so explicitly; `complete_type` can, and now does — and writing its test found the mapping broken for the same reason as the examples, **third field in this file that lives on `Spec` rather than `spec.cmd`**.

That pattern is worth naming: `examples`, `usage` and `complete` are all written at the top level, all hang off `Spec`, and all three were silently dropped by a builder that read only the command. A usage-lib helper for "what the top level declares that belongs to the root" would stop it recurring — separate issue if wanted.

## Rebase note

This branch opened with a fix for the optional flag value. #969 landed the same rule independently as `FlagMeta::value_optional` — opposite polarity to the field here — with the derive attribute, the KDL emitter, and a conformance test. That commit is dropped and the builder speaks main's spelling; all 41 vectors pass against it, which is an independent confirmation of that fix.

## Checks

`cargo test --all --all-features`, clippy `-D warnings`, `cargo fmt`, `prettier` — clean.

## Summary by CodeRabbit

* **New Features**
  * Added a rendering conformance tool with human-readable and JSON reporting.
  * Added support for comparing command usage and help output across implementations.
  * Added rendering coverage for flags, arguments, aliases, defaults, sections, and edge cases.

* **Documentation**
  * Documented the rendering corpus format, expected results, supported cases, and testing workflow.
  * Clarified the roles of the argv, configuration, and rendering corpora.

* **Tests**
  * Added comprehensive rendering vectors and validation for implementation agreement and known differences.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to conformance tooling, corpus JSON, and test binaries; production parsers are only exercised via existing render APIs, with a small shared refactor in the argv harness path.
> 
> **Overview**
> Adds a **rendering conformance layer** parallel to the argv corpus: JSON vectors under `corpus/render/` pin usage lines and optional full `-h` / `--help` pages, with the same two-way `reference` labels when usage-lib disagrees.
> 
> **Harness:** `conformance/src/render.rs` runs each vector through usage-lib and usage-argv; `conformance/tests/render.rs` enforces agreement, reference-label honesty, and a fixed count of usage-argv out-of-scope cases (e.g. `disable_help`). **`render-oracle`** (human + `--json` output) supports authoring expectations from measured output.
> 
> **Refactor:** Spec → usage-argv table construction moves into shared **`conformance/src/tables.rs`** (parse tables + help metadata, including top-level `Spec` fields like examples, `usage`, and completer lookup). The argv harness now calls `tables::build` instead of inlined leaking logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 233384d9042f29bc0bf9a25b0b0747539eb4bc6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/conformance/Cargo.toml
+++ b/conformance/Cargo.toml
@@ -28,3 +28,7 @@ usage-derive = { workspace = true }
 [[bin]]
 name = "oracle"
 path = "src/bin/oracle.rs"
+
+[[bin]]
+name = "render-oracle"
+path = "src/bin/render-oracle.rs"
--- a/conformance/src/argv.rs
+++ b/conformance/src/argv.rs
@@ -58,7 +58,15 @@ pub fn run(vector: &Vector) -> Outcome {
     // hold it. Everything below inherits it, which the parser now does itself rather than
     // this flattening it on the way in — a second implementation of the same rule, and the
     // one that hid the parser not having it.
-    let root = tables::build(&spec.cmd, spec.unknown_flags.map(convert_unknown_flags)).cmd;
+    //
+    // No completers, because this harness binds a line and no completion metadata changes what
+    // binds. `build_spec` resolves a spec's own, for the rendering corpus.
+    let root = tables::build(
+        &spec.cmd,
+        spec.unknown_flags.map(convert_unknown_flags),
+        &[],
+    )
+    .cmd;
     // `default_subcommand` is a property of the spec rather than of a command, so it is
     // resolved once, here, against the root's own subcommands. A name that answers to
     // nothing is left as None: the spec is what it is, and a vector that expects routing
--- a/conformance/src/argv.rs
+++ b/conformance/src/argv.rs
@@ -1,13 +1,8 @@
 //! Running the corpus against [`usage_argv`], the compiled parser.
 //!
 //! usage-argv reads `static` tables that a derive macro is meant to emit. Nothing
-//! emits them yet, so this module builds them from a [`Spec`] instead, which also
-//! makes the corpus usable as usage-argv's test suite from the first commit.
-//!
-//! The tables are leaked. They must outlive the parse and be `'static`-shaped,
-//! and a test process that builds a handful of small tables and exits is the one
-//! place where leaking is the simplest correct answer. Generated code has no such
-//! problem: its tables really are `static`.
+//! emits them here, so [`crate::tables`] builds them from a [`Spec`] instead, which
+//! also makes the corpus usable as usage-argv's test suite from the first commit.
 //!
 //! # Scope
 //!
@@ -21,10 +16,9 @@ use std::collections::BTreeMap;
 use std::ffi::OsStr;
 
 use usage::{Spec, SpecCommand};
-use usage_argv::{
-    Arg, Command, DoubleDash, Error, Event, Flag, Parser, UnknownFlags as ArgvUnknownFlags,
-};
+use usage_argv::{Command, Error, Event, Parser};
 
+use crate::tables::{self, convert_unknown_flags, leak};
 use crate::{ErrorCode, Expect, Layer, Parsed, Value, Vector};
 
 /// What usage-argv did with a vector.
@@ -64,7 +58,7 @@ pub fn run(vector: &Vector) -> Outcome {
     // hold it. Everything below inherits it, which the parser now does itself rather than
     // this flattening it on the way in — a second implementation of the same rule, and the
     // one that hid the parser not having it.
-    let root = build(&spec.cmd, spec.unknown_flags.map(convert_unknown_flags));
+    let root = tables::build(&spec.cmd, spec.unknown_flags.map(convert_unknown_flags)).cmd;
     // `default_subcommand` is a property of the spec rather than of a command, so it is
     // resolved once, here, against the root's own subcommands. A name that answers to
     // nothing is left as None: the spec is what it is, and a vector that expects routing
@@ -177,14 +171,6 @@ fn code(err: Error<'_, '_>) -> ErrorCode {
     }
 }
 
-/// The spec's spelling of the setting, in the parser's terms.
-fn convert_unknown_flags(mode: usage::UnknownFlags) -> ArgvUnknownFlags {
-    match mode {
-        usage::UnknownFlags::Value => ArgvUnknownFlags::Value,
-        usage::UnknownFlags::Error => ArgvUnknownFlags::Error,
-    }
-}
-
 /// Which flags accumulate rather than replace.
 enum Multi {
     Count,
@@ -223,109 +209,4 @@ fn out_of_scope(vector: &Vector) -> Option<&'static str> {
              owns the target struct",
         ),
     }
-}
-
-/// Build leaked tables mirroring a spec command.
-///
-/// `unknown_flags` is carried through as the spec states it — `None` where a command says
-/// nothing — because the parser inherits it. The root takes the spec-level setting, since
-/// that is the command a spec's own property describes.
-fn build(
-    cmd: &SpecCommand,
-    root_unknown_flags: Option<ArgvUnknownFlags>,
-) -> &'static Command<'static> {
-    let unknown_flags = cmd
-        .unknown_flags
-        .map(convert_unknown_flags)
-        .or(root_unknown_flags);
-    let flags: Vec<&'static Flag<'static>> = cmd
-        .flags
-        .iter()
-        .map(|f| -> &'static Flag<'static> {
-            let longs: Vec<&'static str> = f.long.iter().map(|l| leak(l)).collect();
-            let shorts: Vec<u8> = f.short.iter().map(|c| *c as u8).collect();
-            Box::leak(Box::new(Flag {
-                key: 0,
-                name: leak(&f.name),
-                longs: Box::leak(longs.into_boxed_slice()),
-                shorts: Box::leak(shorts.into_boxed_slice()),
-                // usage-lib stores the negation with its dashes; the table wants
-                // the bare name.
-                negate: f.negate.as_ref().map(|n| leak(n.trim_start_matches('-'))),
-                takes_value: f.arg.is_some(),
-                // Only a variadic *argument* is greedy. A `var` flag with a
-                // single-value argument is repeatable instead: one value per
-                // occurrence, which the parser gets by not collecting.
-                variadic: f.arg.as_ref().is_some_and(|a| a.var),
-                // The bound on one occurrence's values, which is the argument's. A
-                // repeatable flag's own `var_max` counts occurrences and is checked after
-                // the parse, so it does not belong in this table.
-                var_max: f
-                    .arg
-                    .as_ref()
-                    .filter(|a| a.var)
-                    .and_then(|a| a.var_max)
-                    // Saturating rather than truncating: `4294967296 as u32` is zero,
-                    // which would read as "stop at once" rather than "no real limit".
-                    .map(|max| u32::try_from(max).unwrap_or(u32::MAX)),
-                global: f.global,
-            }))
-        })
-        .collect();
-
-    let args: Vec<&'static Arg<'static>> = cmd
-        .args
-        .iter()
-        .map(|a| -> &'static Arg<'static> {
-            Box::leak(Box::new(Arg {
-                key: 0,
-                name: leak(&a.name),
-                var: a.var,
-                var_max: a
-                    .var_max
-                    .filter(|_| a.var)
-                    .map(|max| u32::try_from(max).unwrap_or(u32::MAX)),
-                double_dash: match a.double_dash {
-                    usage::SpecDoubleDashChoices::Required => DoubleDash::Required,
-                    usage::SpecDoubleDashChoices::Preserve => DoubleDash::Preserve,
-                    usage::SpecDoubleDashChoices::Automatic => DoubleDash::Automatic,
-                    _ => DoubleDash::Optional,
-                },
-            }))
-        })
-        .collect();
-
-    let subcommands: Vec<&'static Command<'static>> = cmd
-        .subcommands
-        .values()
-        // A subcommand states its own or says nothing; there is no spec-level setting to
-        // hand it, since the root has already taken that.
-        .map(|sub| build(sub, None))
-        .collect();
-
-    let aliases: Vec<&'static str> = cmd
-        .aliases
-        .iter()
-        .chain(cmd.hidden_aliases.iter())
-        .map(|a| leak(a))
-        .collect();
-
-    Box::leak(Box::new(Command {
-        name: leak(&cmd.name),
-        aliases: Box::leak(aliases.into_boxed_slice()),
-        flags: Box::leak(flags.into_boxed_slice()),
-        args: Box::leak(args.into_boxed_slice()),
-        subcommands: Box::leak(subcommands.into_boxed_slice()),
-        // Filled in by the caller for the root, which is the only place a spec declares one.
-        default_subcommand: None,
-        unknown_flags,
-        // The corpus describes argv parsing; `--version` is a question the *caller* answers,
-        // so no vector turns on it and the harness leaves it off.
-        version: false,
-        key: 0,
-    }))
-}
-
-fn leak(s: &str) -> &'static str {
-    Box::leak(s.to_string().into_boxed_str())
 }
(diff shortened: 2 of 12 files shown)


jdx:
## Bugbot: top-level examples dropped — confirmed and fixed

Real, and mine rather than usage-argv's. `example` at the top level parses onto `Spec::examples` and leaves `spec.cmd.examples` empty, so the fold in `build_spec` copied the help texts and lost the examples — costing every page rendered through the corpus its Examples section, the one usage-lib still wrote from the same spec. usage-argv itself renders them correctly when the metadata carries them, which the gate fixture already proves.

@jdx's c53756e fixed it and pinned the root's own page. 1ed7f1a adds the two cases that were left, since the interesting half of the rule is the fallback:

- `a-specs-examples-reach-a-page-that-has-none` — a subcommand page showing the spec's, which is what `page_examples` actually implements and what a builder reading only `spec.cmd.examples` breaks
- `a-page-with-its-own-examples-does-not-also-show-the-specs` — and the long form putting a description *before* its command

## One more, found while checking whether anything else was dropped

`disable_help` is the opposite problem, and worth a look. usage-lib reads it and drops the supplied `--help` entry. usage-argv has no equivalent — and [`lib/src/docs/cli/mod.rs`](https://github.com/jdx/usage/blob/main/lib/src/docs/cli/mod.rs) says why the two can't disagree about it:

> usage-argv has no equivalent: `disable_help` is a KDL-only word, so no spec that crate can hold ever carries one, and the two renderers cannot disagree about it.

This harness breaks that premise, because it builds usage-argv's tables from KDL rather than from a Rust type. A vector declaring it makes them disagree — argv offers a `--help` the spec turned off.

I've treated it the way the argv corpus treats a post-binding vector: answered by the reference alone, skipped for usage-argv with the reason attached, and the count of exemptions asserted so the set can't grow unnoticed. Recording it as a `reference` divergence would have been wrong — the divergence isn't usage-lib's, and nothing usage-argv rendered would be right, because the question never reaches it.

**Worth deciding separately:** whether `disable_help` should get a derive spelling. It's a real asymmetry, it just wasn't observable before this corpus made it so. Happy to open an issue rather than leave it in a test comment.

Also carried `min_usage_version`, which the builder was dropping. Nothing renders it, so nothing caught it.

`cargo test --all --all-features`, clippy `-D warnings`, fmt and prettier all clean on the rebase.

jdx:
## Bugbot round 2: `..EMPTY` in `flag_meta` / `arg_meta` — correct, fixed in 5249c3a

The module doc I added in cf58be3 claims nothing in `tables.rs` ends in `..EMPTY`, and those two did. Fair catch on a claim I'd just made.

The two fields being defaulted:

- **`complete`** genuinely cannot come from a spec — it's a Rust function the binary calls, where a spec says `run=`, a shell command. The emitted KDL turns the former into the latter, not the reverse. Now written out as `None` with the reason attached, so the exhaustiveness is real rather than nearly real.
- **`complete_type`** can, and now does.

Writing the test for `complete_type` found the mapping didn't work: top-level `complete` nodes hang off `Spec` and leave `spec.cmd.complete` empty — **the same shape as the examples bug, third time in this file**. Now filled from the spec's map where the command's own has nothing, so a `complete` inside a `cmd` block still wins for that command.

Both are covered by a unit test rather than a corpus vector, because neither reaches a page. `corpus/render` catches a dropped field by the difference it makes to rendered text — which is what caught the examples and what could never have caught these.

## The pattern worth naming

Three fields now (`examples`, `usage`, `complete`) live on `Spec` rather than on `spec.cmd` because they're written at the top level, and all three were silently dropped by a builder that only read the command. If `usage::Spec` grew a helper for "everything the top level declares that belongs to the root", this class of bug would stop recurring. Happy to open an issue — it's a usage-lib API question, not this PR's.

## Also on this branch since the last comment

Rebased onto main, which added `Spec::usage` in #965. The literals here are exhaustive, so it broke the build — kept that way deliberately and documented, because carrying the field took one line and turned up something worth knowing:

**usage-argv honours a declared `usage` synopsis on the root page; usage-lib's help renderer ignores it** (its manpage renderer honours it, `lib/src/docs/manpage/renderer.rs:198`). So `an-explicit-synopsis-replaces-the-root-line` is the corpus's first recorded `reference` divergence, expecting usage-argv's page with a note pointing at the file that would need to change. Not fixing usage-lib here — recording it is what the label is for, and the two-way check means whoever fixes it gets told to delete the label.

@jdx worth a look, since it's your #965 feature and the fix is a one-liner in `lib/src/docs/cli/mod.rs` if you want it in this PR instead.

All checks green on cf58be3 including the perf gate.

jdx:
Heads up: #969 merged an hour or so before this and put the same bit on `FlagMeta` with the opposite polarity, which is why this is now `DIRTY`. My merge caused the conflict, so flagging the reconciliation rather than leaving you to find it.

`main` currently has, from #969:

- `FlagMeta::value_optional: bool`, `EMPTY` false
- `#[usage(value_optional)]` on the derive, refused on anything that is not a flag with a value
- `placeholder(name, variadic, optional)` in the KDL emitter, plus ` required=#false` on the emitted `arg`
- `conformance/tests/optional_flag_value.rs`
- `xtask`: `value_optional` in the usage dialect, and a `skipped.note` for clap

**Your version is better and should win**, on three counts:

1. It covers four pairings; mine covers two. I never handled a required flag whose *value* is defaulted — `<--jobs [n]>` — because I read the bit as "is the value optional" rather than folding the value's own default in. `value_required: arg.is_none_or(|a| a.required && a.default.is_empty())` is the statement I should have written.
2. `EMPTY = true` is the honest default. Mine needed every construction site to opt in to "required", which is the common case.
3. **No derive attribute, and you are right about why.** I added `#[usage(value_optional)]` and shouldn't have. For a spec that already says `required=#false` the bit is fidelity; for a derive-based CLI it is an invitation to advertise a form the parser refuses, since nothing in usage-argv accepts a bare `--bump`. Your note — "emitting one would advertise a form the parser refuses" — is the argument against the attribute I wrote.

So the rebase is a replacement, not a merge: drop `value_optional` and the derive attribute wholesale, keep `value_required`. Two things from #969 worth carrying over rather than deleting:

- The clap dialect must **note** this rather than express it. I first emitted `num_args = 0..=1` and Bugbot correctly pushed back — it makes the clap shadow accept a line the usage shadow rejects, so the perf pair stops measuring one grammar. `main` has `skipped.note("a flag's value being optional in help only")`; that reasoning still applies to `value_required` and the note should survive under whatever name fits.
- `conformance/tests/optional_flag_value.rs` has a round-trip case — square brackets alone come back `required=#true`, since usage-lib reads the attribute as well as the name. Worth keeping as a corpus vector if it isn't already among your 36.

On your "why a corpus came with it": strongly agree, and #972 (also just merged) is the same conclusion reached from the other end. It holds all seven jdx CLIs against usage-lib rather than mise alone, which is what caught the missing version banner and pitchfork's `[BUMP]` — both invisible to a mise-only fixture for exactly the reason you give. But it is Rust-only: it compares compiled `&'static` tables, so it cannot cover the Go renderer in #974/#975. Your `corpus/render/` can. If it would help, the seven specs are vendored at `benches/fleet/*.usage.kdl` and would make a reasonable source of render vectors — same fixtures, three renderers.

Happy to do the rebase myself if you'd rather not untangle someone else's merge; say the word and I'll take it.

jdx:
## Rebased onto main — and main changed the premise twice

**#969 landed the optional flag value independently.** Same rule, opposite polarity: `FlagMeta::value_optional` where this branch carried `value_required`, plus the derive attribute, the KDL emitter, and `conformance/tests/optional_flag_value.rs`. That commit is dropped from this branch and the table builder now speaks main's spelling. All 41 vectors pass against it unchanged, which is an independent confirmation of that fix rather than a duplicate of it.

**#972 widened the parity gate to seven CLIs**, so "one CLI cannot cover this" is no longer the argument. The argument is better now and measured. Across all **809 value-taking flags** in mise and the fleet:

| pairing | in the fleet |
| --- | --- |
| `[--tool <TOOL>]` | 796 |
| `<--v <n>>` | 8 |
| `[--opt [n]]` | 5 — all in pitchfork and aube |
| `<--jobs [n]>` | **0** |
| a value carrying a `default` | **0** |

Those five instances are the entire reason #969's bug was visible. Two rows still sit at zero.

## What the corpus found that the widened gate still cannot

**usage-lib's help renderer ignores a declared top-level `usage` synopsis; usage-argv honours it.** Verified on current main, with and without subcommands:

```
spec: usage "Usage: ex [OPTIONS] <COMMAND>"  (generated line would be: ex [FLAGS] <SUBCOMMAND>)
usage-lib : Usage: ex [FLAGS] <SUBCOMMAND>
usage-argv: Usage: ex [OPTIONS] <COMMAND>
```

usage-lib's *manpage* renderer honours it (`lib/src/docs/manpage/renderer.rs:198`), so the help renderer is the odd one out.

**Six of the fleet's seven CLIs declare one** — mise and hk included — and `fleet.rs` is green anyway, because `xtask gen-shadow` never carries the node into the shadow. Every shadow's `Spec::usage` is `None`, so both sides render the generated line and the divergence is invisible. This corpus builds usage-argv's tables from KDL directly, which is why it sees it.

Recorded as the corpus's first `reference` divergence rather than fixed: the expectation is right and the reference needs changing. The two-way label check means whoever fixes usage-lib gets told to delete the label.

@jdx two follow-ups, neither this PR's business, say the word and I'll open issues or fold either in:
- the usage-lib help fix, which looks like a one-liner in `lib/src/docs/cli/mod.rs`
- `gen-shadow` carrying `usage`, which touches ~16k lines of generated shadow and would need the above first, or the gate goes red

## Also

Three fields now (`examples`, `usage`, `complete`) are written at a spec's top level, hang off `Spec` rather than `spec.cmd`, and were each silently dropped by a builder that read only the command — the examples one being Bugbot's first finding here, `complete_type` being the third and found by writing its test. A usage-lib helper for "what the top level declares that belongs to the root" would stop this recurring.

Title and body updated to match what the PR now is. All checks were green before the rebase; re-running.

jdx:
## Review round 3: the completer lookup was wrong three ways, and fnox proves each one

Both bots landed on the same field from different sides — Bugbot on case and on subcommands, CodeRabbit on subcommands and precedence. All of it is real, and the three faults compound: **fnox's completers resolved to nothing at all** through this builder.

Its spec writes, once, at the top level:

```kdl
// Complete secret keys - this will automatically match args named "key"
complete "key" run="fnox list --complete 2>/dev/null || true"
```

and means the `arg <KEY>` of `get`, `rm`, `set`, and a dozen more. The reference (`cli/src/cli/complete_word.rs:468,513`) resolves that with `spec.complete.get(&arg.name.to_lowercase()).or(cmd.complete.get(...))`. Three rules in one line, and the builder had a different answer for each:

| the reference | the builder had |
| --- | --- |
| key lowercased, because `SpecComplete::parse` stores the node's name lowercased | the name as written, so `<KEY>` never met `key` |
| consulted for whichever command is being completed | folded onto the root's flags and args only |
| the spec's nodes **before** the command's | the command's first |

Fixed by threading the spec's completers down through `build` in the reference's order of preference, so the first match wins, and lowercasing the key.

**One more the findings didn't name.** The lookup tried the *flag's* name first and the value's second. The reference never keys a completer by a flag's name — it completes a flag by handing the flag's `SpecArg` to the same code that completes a positional (`complete_word.rs:183`), so the key is always the value's. A spec that declares both `complete "out"` and `complete "file"` for `--out <FILE>` got `out`'s type here and `file`'s there. Now the value's name, with the flag's kept only as the fallback for a flag that takes no value — which is exactly the `value_name.unwrap_or(flag.name)` that `Spec::to_kdl` writes back, so the round trip closes.

Unit tests rather than corpus vectors, and worth saying why: `complete_type` reaches no page. `corpus/render` catches a dropped field by the difference it makes to rendered text — which is how the missing examples surfaced and why it could never have caught any of this. This is the second finding in this file of that shape.

## Non-ASCII short flags — correct, refused rather than truncated

`'é' as u8` is 0xE9. A UTF-8 line spells `-é` as 0xC3 0xA9, so the table described a flag nobody could type. usage-argv holds a short as one byte and has no representation for anything else, so the builder now asserts rather than mirroring a lie:

```
a short flag must be ASCII for usage-argv's tables, and `-é` is not
```

Deliberately not the saturating treatment `var_max` gets: there is no nearest honest value for a character the table cannot hold.

## The oracle's `--json`, corrected

Fair, and the instruction as written did not work. A row is `{"id", "usage-lib", "usage-argv"}`; only each renderer's object is an `expect` as a vector writes one. The doc comment says that now, and `corpus/render/README.md` shows the extraction:

```sh
cargo run -q -p usage-conformance --bin render-oracle -- --json flag-optional-value-optional \
  | jq '.[0]."usage-lib"'
```

Verified against the real vector rather than written from memory.

---

`cargo test --all --all-features`, clippy `-D warnings`, fmt and prettier all clean. 41 vectors unchanged — none of this reaches a page, which was the point.

PR #1179, opened 2026-08-21
feat(cli): add usage explain

jdx:
`docs/spec/argv.md` opens by saying it exists to define "which token binds to which flag or argument". Nothing in the toolchain would show you that for a given command line — the parser knew and threw it away, so a value that was typed, a value from `$MYCLI_TOKEN` and a value from `default=` were indistinguishable once parsed.

That gap has a cost. mise's hand-written argv scanner silently ignores `mise --env=production` while `mise --env production` works ([jdx/mise discussion #8883](https://github.com/jdx/mise/discussions/8883)); PLAN.md's fleet survey lists eight more places where a CLI re-derives binding knowledge by hand.

```
$ usage explain -f examples/explain.usage.kdl \
      -e MYCLI_COLOR=never -e MYCLI_PROFILE=prod \
      -- mycli -j8 --env=prod build a b -- --raw

mycli -j8 --env=prod build a b -- --raw
command  mycli build

tokens
  [0]  mycli       program
  [1]  -j8         flag -j, value of jobs = "8", attached
  [2]  --env=prod  flag --env, value of env = "prod", attached
  [3]  build       subcommand build
  [4]  a           arg target = "a"
  [5]  b           refused, extra only accepts words after `--`
  [6]  --          separator
  [7]  --raw       arg extra = "--raw"

values
  flag  --jobs       8      argv [1]
  flag  --env        prod   argv [2]
  flag  --color      never  env MYCLI_COLOR
  flag  --profile    prod   env MYCLI_PROFILE
  flag  --strict     true   default_if --profile when="prod"
  arg   <target>     a      argv [4]
  arg   [-- extra]…  --raw  argv [7]

shadowed
  flag  --jobs   default 1     lost to argv [1]
  flag  --color  default auto  lost to env MYCLI_COLOR

errors
  Argument <extra> can only be set after a `--` separator
```

## Three commits

**`refactor(parse): carry phase-1 bindings on the word`** — no behaviour change, and its value is being reviewable as such. `prefix_bindings` was a `VecDeque` popped in step with `input`; two queues staying aligned is an invariant nothing checks, and it was delicate enough to need explaining at three call sites. It moves onto the word. The substitution is behaviour-preserving by construction: `prefix_bindings.pop_front().flatten()` returned `None` both for "phase 1 pushed `None`" and "phase 1 never reached this word", and the only consumer that distinguishes anything wants exactly that collapsed answer.

**`feat(parse)!: record where each value came from`** — the token trace and the value origins. Breaking: `ParseOutput` gains four fields and `#[non_exhaustive]`. Nothing outside the crate constructs one, and the semver gate is off below 6.x by design (`mise.toml:84`).

**`feat(cli): add usage explain`** — the command, a fixture spec, and the docs pointer on the grammar page.

## Design notes

- **Two tables, deliberately.** A table keyed by token cannot show a value that came from nowhere in argv; a table keyed by declaration cannot show a token that bound to nothing.
- **Origins are per occurrence, not per value or per binding.** A `delimiter` turns one token into several values, and `try_bind_default_missing` on a `var` flag appends to a list that may already hold argv values — so `--color=red --color` genuinely has two origins.
- **Synthesized words fold onto the token they came from.** `-sj8` is one word the caller wrote that names two flags and a value; its re-queued tails do not appear as tokens nobody typed.
- **Exit 0 even when the explained line fails.** The report succeeded. Exiting nonzero would kill the tool under `set -e`, in the case it exists for. Where the parse cannot continue at all, the binding phase is asked on its own and the report says the fallbacks did not run.
- **`ValueOrigin::Env` names the variable.** A flag may list `env`, `env_fallback` and `deprecated_env`; "from the environment" does not say which declaration fired or which to delete.

## Found while writing it

`double_dash="automatic"` on `argv` ends `usage explain`'s own flag parsing at the program name, but a later `--` is still honoured as a separator — which is what a78564c0 settled on purpose. So an explained line carrying its own `--` needs the leading separator. Documented on the field and pinned by a test rather than papered over.

## Not here, deliberately

- **Corpus vectors for token attribution.** Attribution is grammar-observable and `usage-argv` already has an `Event` stream that could be compared against it, but extending the corpus format obligates every implementation to answer it — a separate decision. `corpus/07-env-and-defaults.json`'s post-binding vectors are the specification this renders, so the option stays open.
- **Spelling suggestions** for unknown flags: no string-distance dependency in the workspace, and that is its own feature.

## Verification

`cargo test --all --all-features` (124 targets), `cargo test -p usage-conformance` — the refactor's real review — `cargo clippy --all --all-features -- -D warnings`, `mise run lint`, and `mise run render` leaving a clean tree.

Wall clock is **not** measured: this machine was at load average 34 and `lib/benches/parse.rs` moved ±40% on identical code, so any number would be noise. `usage-argv` is untouched, so the gated instruction counts in `benches/gate` cannot have moved. The change costs one `Vec<String>` clone per bound flag value and one push per recorded role, on `usage-lib`'s interpreter rather than the compiled parser.

---

> [!NOTE]
> **Medium Risk**
> Touches the core argv parser and expands public `ParseOutput` (breaking, `#[non_exhaustive]`). Intended parse behavior is preserved, but the binding-phase refactor is large and easy to get wrong on edge cases (bundles, mounts, separators).
> 
> **Overview**
> Adds **`usage explain`**: given a spec and a command line, it reports what each argv token bound to, where non-argv values came from (env, default, `default_if`, `default_missing`), shadowed defaults, and overrides. Text or JSON. Exits 0 even when the explained line fails, so the report is usable under `set -e`. Mounts are never spawned; empty injected answers are reused from lint.
> 
> The parser now keeps that provenance instead of throwing it away. `ParseOutput` is `#[non_exhaustive]` and gains `tokens`, `flag_origins`, `arg_origins`, and `overridden_flags`. New `Parser::explain` / `explain_refused` collect bindings without bailing on the first error. Phase-1 flag ownership moves onto each word (`Token::binding`) instead of a parallel `prefix_bindings` queue.
> 
> Docs, manpage, Fig completions, and a fixture spec (`examples/explain.usage.kdl`) cover the command; the grammar page’s example is snapshot-tested against real output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e928d2dbc6b2f9c7675582d204d3e7a1cb85e64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added `usage explain` to show how command-line tokens are interpreted, including values from defaults and environment settings.
  - Supports specifications from files, stdin, or inline text, with text or JSON output, view selection, and repeatable environment overrides.
  - Reports parsing issues alongside partial results while completing successfully.
  - Added `--install` and `--force` options to completion generation.

- **Documentation**
  - Added CLI reference documentation, usage examples, completion support, and guidance on argument provenance and deprecation warnings.


jdx:
Rebased onto `main` (`9e5c3898`) — the three-way conflicts were in `lib/src/parse.rs` and `cli/src/cli/mod.rs`, all from main moving under the branch rather than from disagreement:

- The two `env_names()` sites the branch had rewritten to capture *which* variable fired are now main's `first_set_env` helper, which already returns the name.
- `Parser::parse`'s error bail had moved into the branch's `parse_collecting`; main's `warn::retain_reached` stays in `parse_collecting` so `explain` sees the same filtered warnings.
- Main added four new early exits (`--version`, `supplied_short`) calling `record_cursor`, which the branch had replaced with `record_stop`. They now go through `record_stop`, so those exits close the token trace like every other one.
- `Explain` implements `usage_rs::Run`, since dispatch is generated from the enum now (#1182) rather than hand-written.

## Review feedback

**Consumed tokens with no recorded role** (CodeRabbit, `lib/src/parse.rs`) — real, all three paths. Fixed in `90949fc8`, but not as `Refused`: those words were not refused, they did a job. A `value_terminator` ends a run of values without being one of them, which is the whole reason it was declared, and a `restart_token` resets the positional cursor. So two new `TokenRole` variants, `ValueTerminator { ends }` and `Restart`, named after what the word did. Three tests pin them.

**Exit status before stdout** (CodeRabbit) — right, a failed run has empty stdout and the test reported a string difference instead. Folded into a `stdout_of` helper used by all three sites.

**`--env` without `=`, `synthesized` = true, the both-phases-failed branch** — all pinned now.

**`OutputFormat::from_str`** — delegates to the derived `ValueEnum` instead of matching the same two words again.

Verified: `cargo test --all --all-features`, `cargo clippy --all --all-features -- -D warnings`, `mise run lint`, `mise run render` clean. The `markdown` benchmark's +10.86% was measured against the old base; the rebase re-runs it against current `main`.

jdx:
Four findings from the latest round, all real, in `f63de8f4` and `6e928d2d`.

**Explain loses prior bindings on a hard fail** (Bugbot, High) — correct, and the comment claiming otherwise was the tell. A failure the binding phase cannot continue past leaves through `?`, and the trace the loop owned went with it — so the one case a report exists for produced no tokens at all. The trace now belongs to the caller: `Parser::explain_refused` hands back either the binding phase's own output (the failure came after it) or the tokens it managed (the failure is where it died), the word that caused it carries a role saying so, and the rest of the queue is marked unread.

```
mycli --env=x -l boom later

tokens
  [0]  mycli    program
  [1]  --env=x  refused, no declaration takes this word
  [2]  -l       not read
```

**Help tokens look unbound** (Bugbot, Medium) — `--help`, `-h`, `--version` and `-V` recorded nothing, so the word read as having bound nothing while a whole help page arrived under `errors`. They are `TokenRole::Builtin` now, and the help page is no longer listed as a failure: the invocation worked.

Writing that turned up the reason it looked unbound at all — the CLI's role conversion had a `_ => Self::Unread` catch-all for the `#[non_exhaustive]` role list, so the new role was silently relabelled "not read". `Unread` is a claim about the word, and saying it about a word the parser acted on is worse than admitting the report is behind the parser, which is what it says now.

**Override rows hardcode the long spelling** (Bugbot, Low) — right, a short-only flag was reported as `--q`. Both sides go through `flag_display` now, as the values and shadowed tables already did.

**`explain` can execute mounts** (CodeRabbit, Major) — correct and worth taking seriously: a command that reads two inputs and prints a report should not spawn whatever a spec file names, least of all a spec file that arrived attached to a bug report. It now injects mount answers exactly as `lint` does — empty first, then a spec declaring nothing — so a line inside the command's own vocabulary is explained exactly and one under a mounting command is explained on the declarations that are readable. `empty_mount_answers` moved up beside `OutputFormat`; both callers want it for the same reason. Test: `a_mount_is_never_run_to_answer_a_report`, using `mount run=\"false --usage\"` so a spawn would be visible.

Full suite, clippy `-D warnings` and `mise run render` clean.

jdx:
### The failing gate is measuring the fixture, not the code

`markdown` +11.01% is entirely this PR's spec growth. The benchmark runs `usage g markdown -mf cli/usage.usage.kdl`, and this PR adds a command to that spec — so there is one more page to render.

Isolated on one machine, callgrind instruction counts, same release profile:

| binary | spec | instructions |
| --- | --- | --- |
| `main` | `cli/usage.usage.kdl` (main) | 224,794,149 |
| this PR | `cli/usage.usage.kdl` (main) | 224,877,854 |
| this PR | `cli/usage.usage.kdl` (this PR) | 249,873,081 |

This PR's binary against main's spec is **+0.04%** — noise. Against its own spec it reproduces the gate's number (CI measured 251,811,255 for the head and 226,846,576 for the base). The generator does the same work per page it did before.

That is the same finding as #1171's, and the same underlying property: `cli/usage.usage.kdl` grows whenever `usage` gains a command, so every command-adding PR trips a 1% gate. `tak.toml` already names this situation for the shadow benchmarks — "a gate that fires for that teaches people to ignore it." Pointing `bench.markdown` at `benches/mise.usage.kdl`, which does not move when the CLI does, is the fix; it belongs in its own PR since it resets the series.

The provenance work itself is on `usage-lib`'s interpreter, and the gated `usage-argv` counts are untouched.

PR #1182, opened 2026-08-21
feat(derive): generate command dispatch

jdx:
## What

The `match` from a parsed subcommand enum to the code that carries the command out — one arm per command, ~210 of them at mise's size, none of them checkable because every arm has the same shape. This generates it.

```rust
#[derive(Subcommands)]
#[usage(run)]
enum Commands {
    Install(Install),
    Sponsors(Sponsors),
}

impl Run for Install {
    type Output = miette::Result<()>;
    fn run(self) -> Self::Output { install(&self.tools, self.force) }
}

fn main() -> miette::Result<()> {
    Cli::parse().command.run()
}
```

Two traits in `usage-argv`, behind no feature (two traits, no code): `Run` for commands that need nothing but what they parsed, `RunWith<Ctx>` for a CLI that hands its commands a config, an output handle, or a client. `#[usage(run)]` / `#[usage(run_with)]` on a `Subcommands` enum generate the match; on a struct that holds nothing but its subcommand field they generate the forward, which is the `usage generate` / `mise config` container.

## Nothing reaches the spec

Which Rust function runs a command is not part of what the CLI *is*, and a spec recording it could be read by nothing but the program that wrote it — so this is `#[usage(skip)]`'s rule, not a new spec node. No KDL, usage-lib, Go or bridge changes are in this PR.

**Proved on usage-cli.** Both of its matches (10 root commands, 9 generators) are gone, and `usage --usage-spec` is byte-identical to the checked-in `cli/usage.usage.kdl` — so no manpage, reference page, or completion script changed. `Command::Sponsors` became `Sponsors(sponsors::Sponsors)` with the `effect` and description moving to the struct, which the spec cannot tell apart.

## Decisions, each because the alternative is a wrong program rather than a missing one

- **Two traits, not one with a defaulted context.** A hundred commands needing nothing shared would each carry `fn run(self, _: ())`. `RunWith`'s generated impl is generic over `Ctx`, so `&Config`, `&mut App` and an owned handle all work from one emission; an enum may declare both.
- **The output is the first variant's**, with the others bound to agree — a `match` has one type. A command returning something else is an `E0271` naming that command; a command added and not implemented is an `E0277` naming it.
- **Opt-in.** The generated impl is the only one an enum can have, so a CLI that wants to act between the parse and the dispatch keeps its match. Asking is also what makes an undispatchable variant an error where it is declared.
- **A `run` struct forwards and does nothing else**, so it holds one field: its subcommands, not in an `Option`. A struct with arguments of its own has to decide what becomes of them, and an `Option` has a state nothing generated can decide about — both implement the trait by hand, which is the root's usual case.
- **A variant that holds nothing cannot be dispatched** — bare, inline-fields, or `external_subcommand`. The first two are served by a struct the derive writes under a name nothing else can name; the third holds argv rather than a command. Naming the `Args` struct is the fix, and is where `effect` belongs anyway. A per-variant `#[usage(run = path::to::fn)]` is the shape if an adopter ever wants the bare spelling dispatched; not built, because one mechanism covers the fleet.

## Tests

- `conformance/tests/dispatch.rs` — selection, a command's own failure, boxed variants, no-argument commands, a nested group where the enum's dispatch reaches a struct whose dispatch forwards to the next enum, a root reading its globals before dispatching, `&mut Ctx` threading, one enum dispatching both ways, and the spec-invisibility check.
- 7 new `derive/src/model.rs` unit tests for each refusal and for `run` parsing beside `rename_all`.
- `usage-rs/tests/facade.rs` — the traits reach an adopter through the `usage::` alias.
- `cargo test --all --all-features`, `cargo clippy --all --all-features --all-targets -- -D warnings`, `cargo fmt --all --check`, `prettier -c .` all clean.

## Docs

New [Dispatch](docs/rust/dispatch.md) page, cross-linked from the Rust index, Subcommands, and the clap migration guide; the derive crate docs gained a `# Dispatch` section; `PLAN.md` records the item and the decisions above under "what a CLI framework has to have".

---

> [!NOTE]
> **Medium Risk**
> Adds a new derive-generated dispatch surface and rewires usage-cli’s command routing. Spec, help, and completions are unchanged, but a codegen bug would mis-route commands.
> 
> **Overview**
> Adds **opt-in generated dispatch** so the hand-written `match` from a parsed subcommand enum to handlers is emitted from the same declaration as parse and spec.
> 
> Commands implement `Run`, `RunWith<Ctx>`, `RunAsync`, or `RunAsyncWith<Ctx>` (always available in `usage-argv`, re-exported from `usage`). `#[usage(run)]` / `run_with` / `run_async` / `run_async_with` on a `Subcommands` enum generate the match; on a container struct they only forward to a required subcommand field. Output type is the first variant’s; others must agree. Async traits use `impl Future` with **no `Send` bound**. Dispatch never appears in the spec.
> 
> Refuses shapes that cannot be named for a trait impl (unit/inline variants, `external_subcommand`, structs with their own args or `Option` subcommands). usage-cli’s root and `generate` matches are gone; `Sponsors` is now a unit `Args` struct. Docs and conformance tests cover nested groups, boxed variants, context, non-`Send` futures, and spec invisibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 822853ffd3c66ac2fb49ac93067cf4f4318f781d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added synchronous, contextual, and asynchronous command execution APIs.
  - Added opt-in automatic routing for nested commands, with validation for supported command shapes and consistent outputs.
  - Added source-code links to generated command documentation.
  - Added plain diagnostic rendering when color output is unavailable.

- **Documentation**
  - Added dispatch guidance, migration instructions, examples, testing documentation, and navigation updates.

- **Tests**
  - Added coverage for synchronous, asynchronous, nested, contextual, and error-handling dispatch scenarios.


jdx:
## `RunAsync` / `RunAsyncWith`, plus review feedback

### Async dispatch

Four traits now, differing only in whether a command is handed a context and whether it is awaited:

|           | no context                         | a context                                        |
| --------- | ---------------------------------- | ------------------------------------------------ |
| **sync**  | `Run` — `#[usage(run)]`            | `RunWith<Ctx>` — `#[usage(run_with)]`            |
| **async** | `RunAsync` — `#[usage(run_async)]` | `RunAsyncWith<Ctx>` — `#[usage(run_async_with)]` |

An implementation writes `async fn run_async(self)`; the generated dispatch is an `async fn` that awaits the selected command. The emitters describe one dispatch rather than repeating four: a context makes the impl generic, and being awaited puts `async` on the signature and `.await` on each arm.

The async pair declares `-> impl Future<Output = Self::Output>` rather than `async fn`, which is the same signature to implement against and **imposes no `Send` bound**. A CLI that spawns gets `Send` by inference out of the concrete commands the dispatch reaches; one on a single-threaded runtime keeps a future holding an `Rc` across an await. Both are asserted in `conformance/tests/dispatch_async.rs` (`send_is_inferred_and_never_required`, `a_future_that_is_not_send_still_dispatches`). The alternative, `-> impl Future + Send` in the trait, buys the ability to *demand* `Send` in generic code at the cost of the single-threaded case — there is no way to have both without a fifth trait.

### Addressed

- **Diagnostics quote back the attribute that was written** — an author who writes `#[usage(run_async_with)]` no longer reads an error about `run`; the trait name follows it. New test: `a_dispatch_refusal_names_the_attribute_that_was_written`.
- **`Send` in the boxed-future examples** is now documented as the CLI's choice rather than part of the contract, since `Output` has no bound — and the test carries a non-`Send` case to prove it.
- **The `run_with` example implements every variant**, as a dispatched enum requires.
- **The Rust index example declares the `command` field** it dispatches through.
- **Both (now four) opt-ins are named** wherever dispatch is mentioned in passing — index, subcommands, migration guide, facade docs.
- **`Task` alias** is lifetime-parameterised consistently with its use.
- **Grammar** in the `Sponsors` module doc.

### Declined, with reason

**`#[derive(Cli, Args)]` on one struct with `#[usage(run)]` emits two identical impls.** Real, but the diagnostic is already precise, because it lands on the user's own line rather than in generated code:

```
error[E0119]: conflicting implementations of trait `Run` for type `Ex`
23 | #[derive(Cli, Args)]
   |          ---  ^^^^ conflicting implementation for `Ex`
   |          |
   |          first implementation here
```

The suggested remedy — a compile error when no eligible subcommand field is found — does not address this case (a struct with a subcommand field and both derives still double-emits) and would fire on a path `Cli::check` has already rejected. Emitting from only one of the two derives is not available either: a derive cannot see which others are applied. Left as is.

PR #1183, opened 2026-08-21
feat(spec): answer __usage_spec__ from a binary's own tables

jdx:
## The gap

`Cli::to_kdl()` has always emitted a complete spec. What was missing is that nothing *served* it: getting a spec out of a usage-built CLI was a convention each adopter reimplemented. The docs taught it as a hand-rolled flag (`docs/rust/spec.md`, `docs/rust/index.md`, and the [cobra guide](docs/spec/integrations/cobra.md) for Go), and usage-cli hand-rolled exactly that — a raw argv check in `cli/src/lib.rs` plus a declared `usage_spec: bool` field.

A convention each adopter writes is one an external tool cannot rely on, and the consumers here are not the adopter: `usage g markdown|manpage|json|completion`, `usage lint`, `usage mcp`, and an agent that wants `effect` before it runs something. clap#918 (open since 2017) and clap#6491 are people asking clap for this. usage's answer is that the spec *is* the export — which only holds if every binary hands it over without its author writing code.

## What this does

Every binary answers `__usage_spec__` with its own spec, from the same static tables the parser reads.

```bash
mycli __usage_spec__ | usage g markdown -f - --out-dir docs
```

Four decisions behind it:

- **A hidden word, not a flag**, for the reason `__complete_word__` is one: a request is not something the CLI *does*, so it is answered before the parse, stays out of the tables, cannot collide with an adopter's flags, and does not perturb the document it prints. Only the first word counts; anywhere else it is an ordinary value.
- **On by default** — an opt-in nobody remembers to write is an endpoint nothing can rely on. A declared command of that spelling wins, checked against the tables rather than in the derive, because a `Cli` derive expands one struct and cannot see a separate `Subcommands` enum's variants. `#[usage(spec_endpoint = false)]` opts out.
- **KDL only.** `usage g json -f -` converts. A hand-rolled serializer in a dependency-free crate would be a second emitter to keep in step with the serde-derived shape.
- **`spec_extra`** appends a file's KDL to `to_kdl()` for a node no attribute carries — a deliberate escape hatch, and see the note below on why it currently has no in-tree user.

Three entry points, for the three shapes a program takes: `parse()` answers and exits; `spec_request(&argv)` hands the document back to an embedder that renders its own output; `to_kdl()` is the same document in-process.

## usage-cli, and what happened to `spec_extra`

This was written with usage-cli adopting `spec_extra` for its own `repository` and `source_code_link_template` nodes, which it had been stitching onto the spec after the fact. While the PR was in review, **#1184 landed `source_code_link_template` as a real derive attribute** and deleted that fragment — so usage-cli no longer needs the hook, and `spec_extra` ships with no in-tree user beyond its own test.

That is the canonicality rule working exactly as written: where a declaration cannot lower into the spec, the derive gains the vocabulary rather than the spec losing it. An escape hatch was the weaker answer and the stronger one arrived first. Kept anyway, as a deliberate hatch, and the docs say plainly that nothing here uses it — happy to drop it instead if you would rather not carry an unused attribute.

usage-cli's spec is byte-identical either way, and `mise run render` leaves every generated artifact unchanged.

It reaches the endpoint through `is_spec_request` in `crate::run` rather than through the generated intercept — it hands its command line to `parse_from` and renders its own output, so nothing generated is ever in a position to intercept. That is the embedder path the docs now describe, exercised by the flagship adopter.

## Not in scope, deliberately

- **JSON from the binary** — `usage g json -f -` already converts.
- **View projection.** A view-invoked binary emits the base document, which already declares every `view`; projecting would hand tooling a *lossier* spec depending on which name launched the process.
- **`--usage-spec` as a built-in spelling** — adopter-owned. A CLI in another language has no derive to generate an endpoint, so the flag remains the convention there, and usage-cli answers both.

## Cost

Measured, stripped release, on a four-flag two-subcommand CLI: **773,424 bytes with the endpoint against 708,768 without** — 65KB, which is the KDL writer and cold metadata being reachable from `main`. A CLI that already calls `to_kdl()` for a checked-in artifact links the same code and pays nothing extra. Nothing lands on the parse path: the intercept reads the argv `parse()` had already collected, so it is one comparison and no allocation.

## Verification

- `cargo test --all --all-features` — clean, no failures.
- `cargo fmt --all --check`, `cargo clippy --all --all-features -- -D warnings` — clean.
- `mise run render` then `git diff` — no artifact moved.
- `diff <(usage __usage_spec__) <(usage --usage-spec)` — identical, and both match the checked-in `cli/usage.usage.kdl`.
- `usage __usage_spec__ | usage lint -` — clean (the five pre-existing `info` lines only).
- New tests: `is_spec_request` unit tests in usage-argv (first-word-only, declaration and alias precedence); five facade tests (answer parses as a spec, ordinary argv is not a request, a declared command outranks the endpoint, opting out changes nothing else, `spec_extra` joins the document the endpoint hands over); and a process-level test in `usage-rs/tests/external.rs` that runs the real intercept — with a control asserting the same fixture rejects an ordinary word, so it proves the request is answered *before* the grammar rather than by passing through it.

---

> [!NOTE]
> **Medium Risk**
> Default-on intercept of the first argv word changes meaning for CLIs with default-subcommand routing (the word is no longer a task name) and slightly grows binaries that did not already link `to_kdl()`. Opt-out and declaration-wins rules are explicit and tested.
> 
> **Overview**
> Every derived CLI now answers `__usage_spec__` with its own KDL spec from the same static tables the parser uses, so tools no longer depend on each adopter wiring a hidden `--usage-spec` flag.
> 
> The request is a **hidden first word**, not a flag: it is handled before parse (same shape as `__complete_word__`), stays out of the tables, and cannot collide with adopter flags. A declared command or alias of that spelling wins. `#[usage(spec_endpoint = false)]` opts out (~65KB of KDL writer on a small CLI). `parse()` prints and exits; embedders use `spec_request`; `to_kdl()` is the same document in-process.
> 
> `spec_extra` appends a compile-time KDL file onto `to_kdl()` as an escape hatch. usage-cli answers both `__usage_spec__` and `--usage-spec` via `is_spec_request` because it renders its own output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c29dea1a84582878483aaf5b094234088d232ffc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1450,7 +1450,9 @@ share of clap's top-voted open requests is usage's existing feature set:
   discussion clap#6491 asking for schemas AI agents can read) — the spec _is_
   the export, `effect` is the safety vocabulary an agent wants, and usage-cli
   renders JSON already. clap#6026 was closed with advice to "implement your
-  own argument parser" — which is this project.
+  own argument parser" — which is this project. **And the binary now hands it
+  over itself** — see the endpoint item below, which is what makes this an
+  answer for a tool in front of somebody else's CLI rather than for its author.
 
 When the migration guide is written, a "top clap feature requests that just
 work here" section is cheap and persuasive; the launch-gate documentation items
@@ -1458,6 +1460,36 @@ above are where it lands.
 
 **Worth building, demand attached:**
 
+- [x] **A binary that describes itself** (clap#918, clap#6491) — `__usage_spec__`,
+      answered from the same static tables the parser reads, in every binary that
+      does not opt out. The gap was never `to_kdl`: it was that getting a spec
+      _out_ of a CLI was a convention each adopter reimplemented, which the docs
+      taught as `#[usage(long, hide)] usage_spec: bool` — so no tool in front of
+      somebody else's CLI could rely on it, which is the only position that
+      matters for `usage g …`, `usage lint`, `usage mcp` and an agent reading
+      `effect`. Four decisions, settled 2026-08-21. **A hidden word, not a
+      flag**, for the reason `__complete_word__` is one: a request is not
+      something the CLI _does_, so it is answered before the parse, stays out of
+      the tables, cannot collide with an adopter's flags, and does not perturb
+      the document it prints. **On by default**, which is the whole point — an
+      opt-in nobody remembers to write is an endpoint nothing can rely on — with
+      a declaration of that spelling winning (checked against the tables, since a
+      `Cli` derive cannot see a separate `Subcommands` enum's variants) and
+      `spec_endpoint = false` for a binary counting bytes, worth 65KB on a small
+      CLI. **KDL only**, since `usage g json -f -` converts and a hand-rolled
+      serializer in a dependency-free crate would be a second emitter to keep in
+      step with the serde-derived shape.
+      **`spec_extra`** appends a file's KDL to `to_kdl()` for a node no attribute
+      carries. It was built for usage-cli's `repository` and
+      `source_code_link_template`, and #1184 gave both of those real attributes
+      while this was in review — which is the canonicality rule working as
+      intended, and leaves the hook with no in-tree user. Kept as the escape
+      hatch rather than removed, and documented as one; reconsider if nothing
+      needs it. usage-cli answers both spellings, reaching the endpoint through
+      `is_spec_request` because it renders its own output rather
+      than calling `parse()`. Not projected through a `view`: the document already
+      declares every view, so projecting would hand a tool a lossier spec
+      depending on which name launched the process.
 - [x] **Subcommand help headings** (clap#1553, 38 votes) — `help_heading`
       landed for flags and arguments; this is the same property on `cmd` nodes,
       so a 210-command CLI can group its help into sections. mise is the
--- a/argv/src/lib.rs
+++ b/argv/src/lib.rs
@@ -1064,6 +1064,29 @@ pub fn render_warnings(warnings: &[warn::Warning<'_>]) -> String {
     warn::render_warnings(warnings)
 }
 
+/// The word a tool sends to ask a binary for its own spec.
+///
+/// Not a flag and not a command: a spec request is not something this CLI *does*, so it is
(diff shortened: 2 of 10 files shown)


jdx:
Answering Bugbot's "new process-level contract" note, since it is the right thing to weigh on this PR and worth stating explicitly rather than leaving to the diff.

**What the endpoint shadows.** A first word of `__usage_spec__` no longer reaches the grammar. For most CLIs that word could only ever have been an unknown subcommand or a positional value. The case that actually changes meaning is a root with a `default_subcommand` — mise has `default_subcommand run` (`benches/mise.usage.kdl:8`), so `mise __usage_spec__` previously meant *run the task named `__usage_spec__`* and now prints the spec.

**Why that is acceptable rather than merely unlikely.** `__complete_word__` already has exactly this shape and shipped with the same trade-off: it is intercepted before the parse and shadows the same task-name space in the same CLI. The dunder spelling is chosen to make the collision negligible, and an adopter has two ways out if it ever is not — declare a command or alias of that spelling, which wins, or `#[usage(spec_endpoint = false)]`, which removes the entry point entirely. What I would not defend is a plain word or a flag; both would collide with things people really name.

**What it does not change.** Nothing reaches the endpoint from a non-first position, `parse_from`/`parse_from_argv`/`try_parse_from` never intercept (so a library embedding a usage CLI is unaffected unless it opts in by calling `spec_request`), and the instruction-count gate is flat across two runs — startup measured +0.05% then −0.07%, which brackets zero and says the check is below this benchmark's noise floor.

**On the new `usage-test` crate** (#1181, which this is now rebased onto): the endpoint's tests do not use it because it covers help pages, parse failures and completion candidates, not spec output. If a spec-round-trip helper is wanted there later, `spec_request` plus a `usage_parser::Spec` parse is the shape these tests already use and would move over unchanged.

PR #1290, opened 2026-08-24
feat(spec): render heading prose in the Go help renderer

jdx:
## Summary

#1282 added prose introducing a help section and deferred the Go renderer to a follow-up. This is that follow-up. Until now Go ignored the new `headings` field and printed the heading alone, so the same spec produced two different pages depending on which renderer drew it — the one thing the two are not allowed to do.

- `Heading` joins `Example` on the command's help metadata, lowered from the same `headings` the emitter writes to the spec.
- `groupsSection` takes a prose lookup, supplied on the long page and nil on the short one, matching the Rust renderer's decision to keep a summary a summary.
- `commandsSection` takes the `long` flag its Rust counterpart already has, so a section built by a subcommand's `help_heading` reads the same on both.
- **`usage generate go` now writes the field too.** It did not, so a generated CLI would have printed the heading alone while every other reader of the same spec showed the text. `TestTheTwoProducersAgree` exists to hold the lowering and generated paths together, but it runs over mise's spec, and mise declares no `heading`.
- Only a declared heading takes prose, as in Rust: the default `Flags` and `Arguments` titles name the entries that asked for no section of their own, and those same entries read under a different default title depending on the renderer.

## On verification

Worth being explicit, because the usual guarantee does not reach this feature. The cross-language page harness compares Go against the reference over mise's 211 pages, and mise declares no headings — so it proves this change breaks nothing, not that the new output matches. `corpus/render`, where the Rust side pins these bytes, has no Go reader.

So the prose itself is covered by unit tests asserting the same byte shape the Rust tests assert (`"\nFilters:\n  …\n\n"`), including that the short page omits it and that an undeclared default title takes none, plus a table test for the emitter.

**Wiring Go into `corpus/render` would cover this and every render feature after it by construction.** That is the real fix for the gap and is worth doing next; it is a larger change than this one and did not belong in it.

## Validation

- `go test ./...` — argv, conformance (corpus, producer comparison, the 211-page mise parity tests), spec, shadow
- `cargo test --all --all-features`, `cargo clippy --all --all-features -- -D warnings`
- `go vet ./...`, `gofmt`, `cargo fmt --all`

_This PR was written by Claude Code._

---

> [!NOTE]
> **Medium Risk**
> Changes user-facing help layout and generated table emission; a mismatch would show different pages per renderer, though this is presentation-only and covered by unit tests.
> 
> **Overview**
> The Go help renderer now prints **section prose** from a command’s `headings` metadata, matching the Rust renderer so the same spec no longer drops that text on generated CLIs.
> 
> On `--help`, declared section titles (flags, args, and subcommand `help_heading` groups) get indented intro text between the title and entries. `-h` stays a summary: prose is omitted. Default titles like `Flags` / `Arguments` never take prose, because those names are renderer-specific, not declared headings.
> 
> `Heading` is lowered from the spec JSON and emitted into generated Go tables alongside examples. Unit tests pin the long-page byte shape and that short help and undeclared defaults stay clean.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52bb9f7cc8c83c9b585a806dbe20d03ee073e8fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/go/argv/help.go
+++ b/go/argv/help.go
@@ -102,6 +102,8 @@ type Help struct {
 	SubcommandRequired    bool
 	// Examples are worked invocations, printed last.
 	Examples []Example
+	// Headings is prose for the sections this command's entries build, by title.
+	Headings []Heading
 }
 
 // HelpTable is the cold help table, indexed by key: entry `Key` sits at
--- a/go/argv/page.go
+++ b/go/argv/page.go
@@ -49,6 +49,14 @@ type HelpSpec struct {
 	HelpTemplate string
 }
 
+// Heading is prose introducing one help section, printed between the heading
+// and its entries. Keyed by title, because a section is assembled from
+// everything that names it rather than owned by any one entry.
+type Heading struct {
+	Title string
+	Help  string
+}
+
 // Example is one worked invocation, as a page prints it.
 type Example struct {
 	Header string
@@ -124,7 +132,7 @@ func ShortHelp(spec HelpSpec, path []string, chain []*Command, help HelpTable) s
 	// usage-lib prints the whole path from the root rather than the child's own
 	// name.
 	if meta == nil || !meta.FlattenHelp {
-		commandsSection(&sections.commands, path[min(1, len(path)):], cmd, help)
+		commandsSection(&sections.commands, path[min(1, len(path)):], cmd, help, false)
 	}
 
 	args := visibleArgs(cmd, help, false)
@@ -137,6 +145,7 @@ func ShortHelp(spec HelpSpec, path []string, chain []*Command, help HelpTable) s
 	}
 	groupsSection(&sections.args, &sections.ungroupedArgs, &sections.groupedArgs, "Arguments", len(args),
 		func(i int) string { return headingOf(help, args[i].Key) },
+		nil,
 		func(w *strings.Builder, i int) {
 			a := args[i]
 			h := help.Lookup(a.Key)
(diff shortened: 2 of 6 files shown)


jdx:
Rebased onto `main` and fixed the emitter gap in `52bb9f7c`.

**Emitter omits heading prose** — correct, and the more important half of this PR. I had wired the lowering path (spec JSON → `argv.Help`) but not `usage generate go`, so a CLI built from the generated tables would have printed the heading alone while every other reader of the same spec showed the text. `lib/src/go/mod.rs` now emits `Headings` beside `Examples`, with a table test.

Worth noting why nothing caught it: `TestTheTwoProducersAgree` exists precisely to hold the lowering and generated paths together, but it runs over mise's spec, and mise declares no `heading`. Same blind spot the rest of this feature has, and the same reason the PR argues for giving Go a reader for `corpus/render`.

The rebase also had to absorb #1284, which unified the short and long command lists into one `commandsSection`. Rust gave that function a `long` parameter and kept the prose behind it; Go's now takes the same flag, so a section built by a subcommand's `help_heading` still reads on the long page only.

Re-verified in both languages: `go test ./...` (argv, conformance corpus, producer comparison, the 211-page mise parity tests), `go vet`, `gofmt`, plus `cargo test --all --all-features` and `cargo clippy --all --all-features -- -D warnings`.

PR #1291, opened 2026-08-24
feat(help): align long-help annotations to the description column

jdx:
Follows #1284 and #1287, both merged. Last of the three help-layout changes.

## What

On the wide page each annotation gets a line of its own, and every one of them was written at a fixed four-space indent regardless of where the description it belonged to sat. So a section held together right up to the point an entry had something to annotate:

```
Flags:
  -m, --manager <MANAGER>  Only prune packages for this manager
                           [possible values: brew, brew-cask]
    (default: brew)
```

Both of those lines follow the same description. The `[possible values:]` line is a wrapped continuation, which has always been indented to the column; `(default: brew)` is an annotation, which was not. Two rules for two lines that read as one list.

Now:

```
Flags:
  -m, --manager <MANAGER>  Only prune packages for this manager
                           [possible values: brew, brew-cask]
                           (default: brew)
  -n, --dry-run            Print what would be removed without deleting anything
  -y, --yes                Skip the confirmation prompt
  -h, --help               Print help
```

An annotation is a note about the same entry, so it belongs under the text it qualifies rather than in the gutter beside a column it was ignoring.

## Where the old indent stays

Wherever the description is already a block underneath its usage, there is no column to align to and the annotations join it there — which is where they already were:

- a `next_line_help` page,
- an entry whose description has breaks of its own,
- a column so deep there is no room left to say anything.

**Admonitions** keep the fixed indent regardless: they are multi-line prose rather than a note in a bracket, and a deep column would leave them nothing to wrap into. The **flattened** sections keep it too, since their entries have no column either.

## How the two renderers stay agreed

`entry` now returns the indent it settled on, so the annotations cannot disagree with the description above them about where the column is — the decision is made once, by the code that laid the description out. usage-lib carries the same decision into its templates as a serialized indent rather than making a template work it out.

One subtlety worth knowing if you touch this: `render_help_text` wraps whatever it is handed, but a `next_line_help` page ignores the result and blocks anyway. The indent has to follow what the page *does*, not what the wrapper returned, or the annotations align to a column the description never entered.

## Verification

- `benches/gate/tests/help.rs` — mise's 211 commands, both pages, byte for byte: green.
- `benches/gate/tests/fleet.rs`, `go test ./...` including `go/conformance`: green.
- One corpus vector regenerated (`annotations-on-one-line-and-on-their-own`) — it exists to pin exactly this, and its `doc` now says the rule. `render-oracle` reports no new divergences.
- Three inline snapshots accepted, each showing an annotation moving into the column.

## Perf

`bench.startup` is `usage --help`, so this is on the measured path. The whole cost was one allocation: `long_annotations` built its indent for every entry it was called for, and most entries annotate nothing. Building it only when there is a line to put it on takes this to **0.06% below this branch's base** — measured locally with cachegrind, 868,740 against 869,244.

Being straight about the cumulative number, since the gate only ever sees one step at a time: across all three PRs, `usage --help` went from 859,600 to 868,740, or **+1.06%**. Each step passed its own gate and the largest single item was a quadratic `wrap` that #1284 *fixed*, but the total is above what the gate would allow in one move. Happy to spend another pass on it if you want it back under 1% — the remaining cost is spread across the row assembly rather than sitting in one place.

_This PR description was generated by Claude Code._

---

> [!NOTE]
> **Medium Risk**
> User-facing help layout across three renderers; a mismatch would show up as visible `--help` diffs. Logic is presentation-only, with corpus and snapshot coverage.
> 
> **Overview**
> On the wide `--help` page, choices, env, defaults, and deprecation notes used a fixed four-space indent. Wrapped description continuations already sat in the description column, so those notes looked like a second, misaligned list.
> 
> They now take the indent `entry` actually used: the description column when help sat beside usage, and `BLOCK_INDENT` (4) when the description is already a block (`next_line_help`, author line breaks, or no room in the column). Flattened sections and admonitions stay at the block indent.
> 
> Rust argv, Go, and usage-lib share that rule. `entry` returns the indent; usage-lib serializes it as `ann_indent` so templates do not recompute it. Corpus, snapshots, and Go help docs pin the new grid.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58f6a9c809d71d4e84b77f60d4b3810e60016fd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved help-page formatting so annotations such as defaults, choices, and environment variables align beneath their descriptions.
  * Long and multi-line help entries now preserve consistent indentation across supported layouts.

* **Documentation**
  * Clarified how annotations are displayed and wrapped in short, long, and block-style help pages.

* **Bug Fixes**
  * Corrected annotation alignment for long descriptions and entries displayed on a separate line.

### DIFF
--- a/argv/src/help.rs
+++ b/argv/src/help.rs
@@ -26,6 +26,9 @@ use crate::spec::{
 use crate::Command;
 use crate::DoubleDash;
 
+/// The indent a page uses where it cannot align to its column.
+const BLOCK_INDENT: usize = 4;
+
 /// How many flags or arguments are listed individually before collapsing to a placeholder.
 ///
 /// usage-lib's number. Beyond it the line would be longer than it is useful, so it becomes
@@ -1196,6 +1199,7 @@ fn short_sections(
                     if a.hide_env { &[] } else { a.env_fallback },
                     if a.hide_env { &[] } else { a.deprecated_env },
                     if a.hide_default_value { &[] } else { a.default },
+                    BLOCK_INDENT,
                 );
                 return;
             }
@@ -1247,8 +1251,9 @@ fn short_sections(
                 if f.hide_env { &[] } else { f.env_fallback },
                 if f.hide_env { &[] } else { f.deprecated_env },
                 if f.hide_default_value { &[] } else { f.default },
+                BLOCK_INDENT,
             );
-            flag_notes(out, f, 4);
+            flag_notes(out, f, BLOCK_INDENT);
             return;
         }
         let deprecation =
@@ -1527,6 +1532,7 @@ fn flat_commands_short(out: &mut String, path: &[&str], meta: &CommandMeta<'_>) 
                     } else {
                         arg.default
                     },
+                    BLOCK_INDENT,
                 );
                 continue;
             }
@@ -1584,8 +1590,9 @@ fn flat_commands_short(out: &mut String, path: &[&str], meta: &CommandMeta<'_>) 
                     } else {
                         flag.default
                     },
+                    BLOCK_INDENT,
                 );
-                flag_notes(out, flag, 4);
+                flag_notes(out, flag, BLOCK_INDENT);
                 continue;
             }
             if let Some(help) = flag.help.filter(|help| !help.trim().is_empty()) {
@@ -2102,7 +2109,7 @@ fn long_sections(
         |title| heading_help(meta, title),
         |out, a| {
             let text = a.long_help.or(a.help);
-            entry(
+            let indent = entry(
                 out,
                 &arg_usage(a),
                 text,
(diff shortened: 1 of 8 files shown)


jdx:
The gate agrees with the local measurement: `startup` 885,316 → 885,174, **-0.02%**. (I predicted -0.06% from cachegrind here; the absolute numbers differ between machines, the direction and magnitude don't.)

`markdown` moved +0.36%, under the gate. That path never calls the help renderer, so it is not the alignment itself — it is `SpecFlag` and `SpecArg` each carrying one more field, which markdown's model building copies around. The field is `skip_serializing_if = "String::is_empty"`, so nothing reaches the rendered output.

The cumulative note in the description still stands and is visible in the sparkline: the last two points are the two merged PRs. Say the word if you want a pass at bringing the total back under 1%.

PR #1296, opened 2026-08-24
refactor(lib): internalize kdl and diagnostics

jdx:
## Summary

- internalize the KDL parser and document API used by usage
- replace miette reporting with a lightweight native diagnostic renderer while preserving source spans, labels, and help text
- remove kdl and the default miette dependency while providing opt-in miette reporter interoperability
- include the applicable Apache 2.0 license and third-party notices

The default build remains free of miette. Callers can enable the `miette` feature to make `UsageErr` and KDL parse diagnostics implement `miette::Diagnostic`, preserving source spans, labels, severity, and help text in their existing reporter.

## Tests

- `cargo test --all --all-features --quiet`
- `cargo clippy --all --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

_This pull request was generated by Codex._

---

> [!NOTE]
> **High Risk**
> Vendors a full KDL parser (winnow `unstable-recover`) and changes error types and rendering across the library and CLI. Parse diagnostics and downstream miette interop are security-adjacent and easy to get wrong.
> 
> **Overview**
> **Vendors KDL and drops default miette.** `usage-lib` now ships a trimmed copy of kdl-rs 6.7.1 (`src/kdl/`) and a small local diagnostic layer (`src/miette.rs`) instead of depending on `kdl` and fancy `miette`. Spec parsing still uses the KDL v2 document model, spans, and formatting; serde, queries, and KDL v1 fallback are omitted.
> 
> **Errors render in-process.** `UsageErr` no longer derives `miette::Diagnostic` by default. It renders source labels, help, and codes (`usage::file`, `usage::shell`) itself. An optional `miette` feature reimplements `miette::Diagnostic` so existing reporters still work.
> 
> **Call sites follow the new types.** CLI, conformance, and spec modules import `usage::kdl` and `usage::miette` (`Result`, `bail!`, `IntoDiagnostic`). Apache-2.0 notices for kdl-rs/miette are recorded in `NOTICE.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aaacbe35d9f7f281c872bfafa0813b835c4a2285. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added KDL v2 document parsing, formatting, serialization, and identifier support.
  * Improved configuration and syntax diagnostics with source locations, labels, diagnostic codes, help text, and related error details.
  * Enhanced diagnostic rendering for Unicode text, tabs, varied newline formats, and nested error messages.

* **Documentation**
  * Added third-party notices and Apache-2.0 licensing information for included components.


lib/src/miette.rs
```diff
@@ -0,0 +1,239 @@
...
+    };
+    let mut out = format!(
+        "  × {title}\n   ╭─{location}\n {line_no:>width$} │ {line}\n   · {blank:column$}{mark:─<marked$}┬\n   · {blank:indent$}╰── {label}\n   ╰────",
+        blank = "",
+        mark = "",
+        width = width,
+        column = column,
+        marked = marked,
+        indent = column + marked,
+    );
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221. All diagnostic frame rows now use the dynamic line-number gutter width, with a regression test at line 10. _This reply was generated by Codex._

lib/src/kdl/document.rs
```diff
@@ -0,0 +1,1169 @@
...
+            check_span(ident.value(), ident.span(), source);
+        }
+    }
+
+    #[track_caller]
+    fn check_span(expected: &str, span: SourceSpan, source: &impl miette::SourceCode) {
+        let span = source.read_span(&span, 0, 0).unwrap();
+        let span = std::str::from_utf8(span.data()).unwrap();
+        assert_eq!(span, expected);
+    }
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221 by removing the unreachable vendored upstream test modules and KDL v1 compatibility paths, along with the unexpected-cfg suppression. Only the supported trimmed KDL v2 implementation remains. _This reply was generated by Codex._

lib/src/error.rs
```diff
@@ -62,8 +57,7 @@ pub enum UsageErr {
     Version(String),
 
     #[error("Invalid usage config")]
-    #[diagnostic(transparent)]
-    Miette(#[from] miette::MietteError),
+    Miette(#[from] crate::miette::MietteError),
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221. The wrapped message is now included in `UsageErr::Miette`'s plain Display output, with regression coverage. _This reply was generated by Codex._

lib/src/miette.rs
```diff
@@ -0,0 +1,239 @@
...
+        .find('\n')
+        .map_or(source.len(), |at| offset + at);
+    let line = &source[line_start..line_end];
+    let line_no = source[..line_start].bytes().filter(|b| *b == b'\n').count() + 1;
+    let column = source[line_start..offset].chars().count();
+    let marked = source[offset..line_end]
+        .chars()
+        .take(span.len().max(1))
+        .count()
+        .max(1);
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221. Byte spans are normalized to UTF-8 boundaries and underline width is calculated from the resulting character slice; the test covers a multi-byte character and a caller-supplied mid-codepoint offset. _This reply was generated by Codex._

lib/src/kdl/v2_parser.rs
```diff
@@ -0,0 +1,2531 @@
...
+    let mut doc = nodes.parse_next(input)?;
+    let badend = resume_after_cut(
+        cut_err(eof).context(cx().lbl("EOF").msg("Expected end of document")),
+        any.void(),
+    )
+    .parse_next(input)?
+    .is_none();
+    if badend {
+        document.parse_next(input)?;
+    }
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221. Document recovery and comment-body traversal are iterative now; regression tests exercise 20,000-character inputs for both paths. _This reply was generated by Codex._

lib/src/kdl/mod.rs
```diff
@@ -0,0 +1,24 @@
+//! The KDL document model and parser used by usage.
+//!
+//! This is a trimmed copy of kdl-rs 6.7.1: serde, queries, KDL v1 fallback, and the miette
+//! integration are omitted. The parser and formatting-preserving document types remain intact.
+//! See `NOTICE.md` and `third-party/kdl-LICENSE-APACHE` for attribution and license terms.
```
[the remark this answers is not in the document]
jdx: Fixed in 6dac0221. The module documentation now points to `lib/third-party/LICENSE-APACHE-2.0`. _This reply was generated by Codex._

jdx:
Addressed the remaining review-summary items in `6dac0221` as well: Winnow now requires the tested `>=0.7.15, <0.8` range, and plain `KdlError` formatting includes the first diagnostic message. The full all-feature test suite, Clippy with warnings denied, formatting, diff checks, and the no-default-features library build pass locally.

_This comment was generated by Codex._

lib/src/error.rs
```diff
@@ -7,16 +8,17 @@ use thiserror::Error;
 /// this enum grows every time the spec learns to say something new — `MissingGroup`
 /// arrived with groups, `ArgRequiresDoubleDash` with `double_dash` — and without this
 /// each one is a major release for everyone downstream.
-#[derive(Error, Diagnostic, Debug)]
+///
+/// With the `miette` feature enabled, this implements `miette::Diagnostic` so applications can
+/// pass it directly to their existing miette reporter. The feature is disabled by default.
+#[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum UsageErr {
```
jdx, line 16: This is a semver-breaking change to usage-lib's public API shipped under a plain `refactor(lib)` title: `UsageErr` loses `miette::Diagnostic` by default (now behind the off-by-default `miette` feature), and the public `From<&Spec*> for KdlNode` impls retarget the vendored `usage::kdl` types, so downstream code using the crates.io `kdl` v6 types (or handing `UsageErr` to a miette reporter) stops compiling. release-plz will cut a non-major version from `refactor:`, and consumers like mise pick it up inside `^6` via `cargo update`. Suggest marking the commit/PR `refactor(lib)!:` so release-plz majors it — or intentionally deciding otherwise, but explicitly.

lib/Cargo.toml
```diff
@@ -37,6 +38,7 @@ shell-words = "1"
 strum = { version = "0.28", features = ["derive"] }
 tera = { version = "2", optional = true }
 thiserror = "2"
+winnow = { version = ">=0.7.15, <0.8", features = ["alloc", "unstable-recover"] }
```
jdx: `unstable-recover` is semver-exempt in winnow (documented as free to change in any release), and this is an open range — a future 0.7.x that touches `Recoverable`/`recoverable_parse`/`FromRecoverableError` breaks every fresh downstream build (mise CI, docs.rs, new adopters resolve newest 0.7) until usage repins. Upstream kdl-rs carried the same range, but there the breakage was kdl's to fix; vendoring moves that pager to usage-lib. Consider pinning `=0.7.x`.

lib/src/miette.rs
```diff
@@ -0,0 +1,323 @@
...
+pub(crate) fn render_source(
+    title: &str,
+    source_name: &str,
+    source: &str,
+    span: SourceSpan,
+    label: &str,
+    help: Option<&str>,
+) -> String {
+    let offset = source.floor_char_boundary(span.offset().min(source.len()));
+    let line_start = source[..offset].rfind('\n').map_or(0, |at| at + 1);
```
jdx: `render_source` treats only `\n` as a newline, but the vendored parser's `NEWLINES` accepts `\r`, `\u{0085}`, `\u{000B}`, `\u{000C}`, `\u{2028}`, `\u{2029}`. A CR-only file (`name "t"\rbin "t"\r…`) parses/errors correctly but the diagnostic reports line 1, echoes the whole document as one "line" with raw `\r` bytes (which overwrite the caret row in a terminal), and the `[file:line:col]` header is wrong. Reproduced with the built PR CLI. Splitting on the same newline set the parser uses fixes line numbers, bounds, and the echo.

lib/src/miette.rs
```diff
@@ -0,0 +1,323 @@
...
+    let line_end = source[offset..]
+        .find('\n')
+        .map_or(source.len(), |at| offset + at);
+    let line = &source[line_start..line_end];
+    let line_no = source[..line_start].bytes().filter(|b| *b == b'\n').count() + 1;
+    let column = source[line_start..offset].chars().count();
```
jdx: The caret column is a char count padded with spaces while the source line is echoed raw, so any tab or double-width char before the span misaligns the underline: a tab-indented `.usage.kdl` puts the caret 7 columns left of the offending token in an 8-col-tab terminal (reproduced with the built PR CLI); CJK/emoji shift it similarly. miette expanded tabs (tab_width) before drawing. Cheapest fix: expand tabs in the echoed line and count them as their expansion width when padding.

lib/src/kdl/error.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+impl KdlError {
+    pub(crate) fn render(&self) -> String {
+        self.diagnostics
+            .iter()
+            .map(|diagnostic| {
+                crate::miette::render_source(
+                    diagnostic.message.as_deref().unwrap_or("Unexpected error"),
+                    "",
```
jdx: KDL syntax errors lose their locator: `render_source` only emits the `[name:line:col]` header when `source_name` is non-empty, and this call always passes `""`. Old output rendered `╭─[3:7]` even for unnamed sources; new output is a bare `╭─`, so column info is gone entirely from KDL parse errors and anything parsing `line:col` from stderr breaks. Since spec files are read from a known path, threading the filename into `KdlError` (or at least emitting `[line:col]` for unnamed sources) would restore both.

lib/src/miette.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self.0, f)
+    }
+}
+
+impl Debug for Error {
```
jdx, line 145: This `Debug` (the CLI's top-level reporter) prints only the top error's Display / `UsageErr::render()` and never walks `source()`, so the "Caused by" chains miette printed are dropped. Any `into_diagnostic()`-wrapped error whose Display is a summary (io errors, most `#[source]`-carrying thiserror types) now prints only the headline — the root cause the user acts on is gone. `cli/src/cli/generate/completion.rs` already hand-flattens a source chain for exactly this reason, which is evidence the renderer should own it: walk `source()` here once and the per-call-site flattening becomes unnecessary.

lib/src/error.rs
```diff
@@ -79,21 +76,18 @@ pub enum UsageErr {
...
-    #[diagnostic(transparent)]
     KdlError(#[from] kdl::KdlError),
 
     /// A file the spec model was asked to read could not be read.
     ///
     /// Carries the path as well as the io error: "No such file or directory" on its own
     /// names nothing, and this is reported for spec files given on a command line.
     #[error("{0}\nFile: {1}")]
-    #[diagnostic(code(usage::file))]
     FileError(std::io::Error, std::path::PathBuf),
```
jdx, line 86: The old `#[diagnostic(code(usage::file))]` / `code(usage::shell)` codes are gone from rendered output, and the new `cfg(miette)` `Diagnostic` impl implements no `code()`, so it always returns `None`. Exit code is 181 for every `UsageErr`, so stderr was the only way a wrapper distinguished "spec file unreadable" from other failures — that signal silently disappears, and miette-feature consumers classifying via `code()` get nothing. Worth either re-emitting codes in `render()`/the Diagnostic impl or noting the drop as intentional.

cli/src/lib.rs
```diff
@@ -1,8 +1,9 @@
 #[macro_use]
 extern crate log;
-extern crate miette;
+extern crate self as miette;
```
jdx: `extern crate self as miette;` makes usage-cli impersonate the miette crate so the ~21 files' old `use miette::…` imports keep resolving — and it pushes `Error`/`Report`/`Result`/`IntoDiagnostic`/`bail`/`miette` onto the published crate's public root, where they exist only to feed the alias. Two traps: re-adding any real `miette` dependency later collides with the self-alias, and readers/tooling will assume `use miette::IntoDiagnostic` means the real crate. The mechanical fix is small: rewrite the imports to `use usage::miette::…` (as `lib/src/parse.rs` already does) and delete the alias plus the root re-exports.

lib/src/kdl/node.rs
```diff
@@ -0,0 +1,805 @@
...
+impl<'a> Iterator for ChildrenIterMut<'a> {
+    type Item = &'a mut KdlNode;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.as_mut().and_then(|x| x.next())
+    }
+}
+
+// Vec-style APIs
+impl KdlNode {
```
jdx, line 332: A large slice of the vendored surface has zero callers workspace-wide (verified via a privatize-and-check compile experiment): this whole editing impl (`insert`/`remove`/`retain`/`sort_by`/`sort_by_key` and the iterator structs), the `ensure_v2` v1-migration chain (document→node→entry — the v1 parser it served wasn't vendored), `FormatConfig`/`FormatConfigBuilder` (only ever built with defaults at `spec/mod.rs`), assorted accessors, and ~100 lines of commented-out query code referencing deleted types. Because `pub mod kdl` is only `#[doc(hidden)]`, `dead_code` can never flag any of it, and it's all semver-relevant public surface. The PR already trimmed serde/queries/v1 — the same criterion says this goes too, and `usage` only builds nodes via `new`/`push`/`entries_mut`/`children_mut`.

lib/src/kdl/mod.rs
```diff
@@ -0,0 +1,22 @@
+//! The KDL document model and parser used by usage.
+//!
+//! This is a trimmed copy of kdl-rs 6.7.1: serde, queries, KDL v1 fallback, and the miette
```
jdx, line 3: The vendoring drops upstream kdl-rs's parser test suite (multiline-string dedent, raw strings, escapes, hex/octal/binary numbers, exact failure spans) and keeps only the two recursion regression tests — while also removing the external crate that could serve as an oracle. I differential-fuzzed this parser against kdl 6.7.1 on 1200+ inputs: it's currently faithful, and strictly better in two cases (upstream stack-overflows on `"}"×20000` and on very long `/*…*/` comments that the loop rewrites here fix). But a future in-repo edit that breaks dedent or `0x`/`0o`/`0b` parsing keeps `cargo test --all --all-features` green — nothing left in the workspace exercises those constructs. Suggest porting the relevant upstream parser tests (they're Apache-2.0, already attributed) or adding a KDL conformance corpus.

jdx:
Reviewed at high effort: 8 independent passes (line-by-line including a file-by-file diff of the vendored code against upstream kdl-rs 6.7.1 from the cargo cache, removed-behavior audit against the old miette output, cross-file/API tracing with workspace builds, reuse/simplification/efficiency/altitude, conventions), followed by adversarial verification of every candidate. 10 findings posted inline. The big picture:

**The vendored parser itself is solid.** The trim is faithful to upstream; the only two logic changes (EOF-recovery and `commented_block` recursion rewritten as loops) are semantically equivalent and actually fix two upstream stack overflows. Differential fuzzing against kdl 6.7.1 on 1200+ inputs found no divergence. Licensing/attribution (NOTICE.md, third-party license, module provenance header) is complete.

**The release mechanics are the highest-risk item.** Dropping `Diagnostic` from the default build and retargeting the public `From<&Spec*> for KdlNode` impls is an API break, but the `refactor(lib):` title means release-plz cuts a non-major version — see the inline comment on `error.rs`. Related: the winnow `unstable-recover` open range is semver-exempt exposure usage-lib now owns.

**The native renderer regresses miette fidelity in specific, fixable ways** (each reproduced against the built base and PR binaries): CR/NEL/LS line endings mis-render, tabs/wide chars misalign the caret, KDL syntax errors lose `[line:col]` + filename, `source()` cause chains are dropped, and diagnostic codes are gone. None of these block the direction; they're worth burning down before this ships since spec-error UX is the whole point of the diagnostics work.

Smaller items inline: the `extern crate self as miette` alias in the CLI, the dead vendored API surface kept public via `#[doc(hidden)]`, and the deleted upstream test suite leaving parser constructs uncovered.

Not posted inline for cap reasons, worth a line: the parser still allocates upstream's formatting-preservation metadata (per-node trivia strings, `repr`s) that spec parsing never reads — now that the parser is internal, dropping format capture is free parse-time win on the hot path; and `UsageErr`'s span/label mapping is pattern-matched three separate times (`render()`, `source_code()`, `labels()`), which will drift as variants are added.

_This review was generated by Claude Code._

lib/src/error.rs
jdx: The maintainer has explicitly decided that breaking changes within 6.x do not require a major-release marker here, so the existing non-breaking PR title is intentional. The optional `miette` feature remains available for downstream interoperability.

_This reply was generated by Codex._

lib/Cargo.toml
jdx: Fixed in `aaacbe35`: Winnow is now pinned to exactly `=0.7.15`, the release this vendored parser and its semver-exempt `unstable-recover` API are tested against.

_This reply was generated by Codex._

lib/src/kdl/error.rs
jdx: Fixed in `aaacbe35`: unnamed KDL diagnostics now emit `[line:col]`, while file parsing threads `ParsingContext.file` into `KdlError` and emits `[filename:line:col]`. Both paths have regression coverage.

_This reply was generated by Codex._

lib/src/kdl/node.rs
jdx: Fixed in `aaacbe35`: removed the unused KDL v1 `ensure_v2` chain, editing and iterator APIs, commented query stubs, and public formatting builder and configuration surface. The remaining internal formatting configuration is only what spec rendering uses.

_This reply was generated by Codex._

lib/src/kdl/mod.rs
jdx: Fixed in `aaacbe35`: restored representative Apache-2.0 parser coverage for multiline dedent and newline normalization, raw strings, escapes, hexadecimal, octal, binary values, and exact diagnostic spans, alongside the two iterative-recovery regressions.

_This reply was generated by Codex._

cli/src/lib.rs
jdx: Fixed in `aaacbe35`: all CLI call sites now name `usage::miette` directly. The `extern crate self as miette` alias and the diagnostic re-exports from the CLI crate root are removed.

_This reply was generated by Codex._

lib/src/miette.rs
jdx: Fixed in `aaacbe35`: source lines expand tabs to four-column tab stops, and both prefix and underline widths use Unicode display width. A regression test combines a tab and a double-width CJK character before the span.

_This reply was generated by Codex._

lib/src/miette.rs
jdx: Fixed in `aaacbe35`: the native renderer now recognizes the full KDL newline set, including CRLF as one separator, and tests every accepted newline form for line number, source bounds, and locator output.

_This reply was generated by Codex._

lib/src/error.rs
jdx: Fixed in `aaacbe35`: native rendering again emits `usage::file` and `usage::shell`, and the optional `miette::Diagnostic` implementation returns the same values from `code()`. Both modes are tested.

_This reply was generated by Codex._

lib/src/miette.rs
jdx: Fixed in `aaacbe35`: the top-level native report formatter now walks `std::error::Error::source()` and renders an indexed `Caused by:` chain. A nested-error regression test verifies the root cause remains visible.

_This reply was generated by Codex._

jdx:
The ten inline findings are addressed in `aaacbe35` and their threads are resolved. For the two uncapped follow-up observations: formatting metadata remains intentionally intact for the KDL document round-trip and rendering behavior in this PR; removing parser-side format capture is a separable performance change that should be measured independently. The repeated `UsageErr` source annotation matching is also unchanged for now because the variants have different source-code representations under optional miette interoperability; the new regression tests cover native and miette mappings so additions cannot silently drift.

Local verification passes: `cargo test --all --all-features`, `cargo clippy --all --all-features --all-targets -- -D warnings`, formatting, diff checks, and the no-default-features library build.

_This comment was generated by Codex._

PR #1297, opened 2026-08-24
feat(help): use semantic colors by default

jdx:
## Summary

- render help headings, option literals, and metavariables with semantic colors by default
- add dependency-free, runtime `help_template` style tags with nesting and semantic or physical styles
- support escaped delimiters with doubled dollar signs and keep malformed programmatic templates non-panicking
- validate template styles in Rust derives and KDL specs while stripping tags from plain and generated Go help
- pin Rust, derive, and Go style vocabulary and multiline/empty-section behavior with shared conformance fixtures
- style flattened descendants, required options, lowercase metavariables, and Unicode input consistently

No new crate dependency is added.

## Tests

- `cargo test --all --all-features`
- `cargo clippy --all --all-features --all-targets -- -D warnings`
- `cd go && go test ./...`
- `cd go && go vet ./...`
- `cargo fmt --all -- --check`
- `prettier -c .`

## Summary by CodeRabbit

* **New Features**
  * Help templates now support 23 named styles, including colors, emphasis, headings, options, and metavariables.
  * Added nested and combined styling, escaped delimiters, and styling around help sections.
  * Plain and generated help output removes formatting tags while preserving content.

* **Bug Fixes**
  * Improved styling for required flags, Unicode and lowercase metavariables, and descendant usage details.
  * Invalid or malformed style markup now falls back safely without disrupting help rendering.

* **Documentation**
  * Updated help-template references to document styling, escaping, and plain-output behavior.


argv/src/help.rs
```diff
@@ -480,31 +462,76 @@ fn closing_delimiter(
...
+                }
+            }
+            continue;
+        }
+
+        if rest.starts_with(char::is_uppercase)
+            && previous.is_none_or(|c| c.is_whitespace() || matches!(c, '=' | '[' | '<'))
+        {
+            let end = rest
+                .find(|c: char| !(c.is_ascii_uppercase() || c.is_ascii_digit() || c == '_'))
```
jdx: Infinite loop with unbounded memory growth: the gate `rest.starts_with(char::is_uppercase)` is full Unicode, but this end-scan is ASCII-only — for a leading `Ü`/`É`/`Ф` it returns `Some(0)`, so `end == 0`, `at += 0`, and the loop never advances (in coloured mode each iteration also pushes an empty `\x1b[1;35m\x1b[0m` pair, growing `out` without bound). Reproduced: `styled_flag_usage("ex [ÜBERSICHT]", Style::COLOURED)` hangs until killed. Any synopsis or custom usage string with a non-ASCII uppercase word — a German metavar, a Cyrillic value name — hangs `--help`. Gate and end-scan need to agree; if the token is meant to be ASCII-ALL-CAPS, make the gate `is_ascii_uppercase` too (which also guarantees `end >= 1`).

argv/src/help.rs
```diff
@@ -480,31 +462,76 @@ fn closing_delimiter(
...
+                    .unwrap_or(value.len());
+                if end > 0 {
+                    out.push_str(&style.metavar(&value[..end]));
+                    at += end;
+                }
+            }
+            continue;
+        }
+
+        if rest.starts_with(char::is_uppercase)
```
jdx: Related, milder: the heuristic fires on any capitalized word after whitespace/`=`/`[`/`<`, and the ASCII-uppercase run for a word like `Ex` or `Add` is just the first letter — so a custom usage string `usage = "ex Add <file>"` renders `\x1b[1;35mA\x1b[0mdd`, visibly splitting the word with only its leading capital painted magenta. Requiring the run to be ≥2 chars (or to end at a word boundary) would keep `FILE`/`TOOL@VERSION` coloured without firing on ordinary capitalized words.

argv/src/help/template.rs
```diff
@@ -0,0 +1,429 @@
...
+fn push_escaped(out: &mut String, text: &str) {
+    for ch in text.chars() {
+        out.push(ch);
+        if ch == MARK {
+            out.push(MARK);
+        }
+    }
+}
+
+fn collapse_blank_runs(page: &str) -> String {
```
jdx, line 278: `collapse_blank_runs` judges lines by `visible()` text, so a line that is blank *except for a style marker* gets dropped along with its marker, unbalancing the open/close stack in `render_marked`. Two concrete failure modes, both with templates that `check()` accepts:

1. **Panic on every `--help`**: `help_template "{$heading}\nMY TOOL{/$}\n\n{{usage}}"` (open tag on its own line — a natural multi-line layout) drops the open marker; `render_marked` then pops the root frame and panics at the `expect("validated tags balance")` (coloured) or the final `unwrap()` (plain). Reproduced on this branch. usage-lib's reference renderer and Go both render the same template fine, so this also breaks the compiled/reference parity the PR promises.
2. **ANSI bleed**: the mirror case — `{/$}` alone on a line (e.g. closing a wrapped section that rendered empty) — drops the *close* marker while the opener survives, so the page ends with an active SGR and no `\x1b[0m`: `substitute("{$dim}fine print\n{/$}", true, ..)` returns `\x1b[2mfine print` and the user's prompt stays dim.

`collapse_blank_runs` needs to preserve markers from dropped lines (e.g. re-emit them onto the surviving neighbor) or run before marker insertion.

lib/src/help_template.rs
```diff
@@ -87,23 +88,116 @@ pub fn check(template: &str) -> Result<(), String> {
...
+                match section(after[..end].trim()) {
+                    Some(text) => out.push_str(&text),
+                    None => out.push_str(&rest[..2 + end + 2]),
+                }
+                rest = &after[end + 2..];
+            }
+            1 => {
+                let end = rest
+                    .find('}')
+                    .expect("templates are checked before rendering");
```
jdx: `substitute` was previously total — malformed braces passed through literally — but now panics via this `expect` on a `{$` with no `}` (and argv's `AnsiStyle::apply` ends in `unreachable!`, reachable in plain mode too). The `check` guard only runs on the KDL/derive authoring paths, while `Spec.help_template` is a `pub` field in both crates and nothing on the render path re-validates. A downstream embedder building a Spec programmatically (or converting from clap and copying a template containing `{$`) gets a process abort where v6.2 rendered the text verbatim. Go's `substituteSections` handles the same input by writing it through, so the three renderers diverge exactly on malformed input. Falling back to literal pass-through (like Go) instead of `expect`/`unreachable!` keeps render paths total.

lib/src/help_template.rs
```diff
@@ -87,23 +88,116 @@ pub fn check(template: &str) -> Result<(), String> {
...
+            }
+            _ => rest = &rest[4..],
         }
-        rest = &after[end + 2..];
     }
-    out.push_str(rest);
     collapse_blank_runs(&out)
 }
 
+fn check_styles(template: &str) -> Result<(), String> {
```
jdx, line 168: There's no escape syntax for a literal `{$`: this validation runs on every parsed spec/derive template, so a previously-valid `.usage.kdl` whose `help_template` contains literal `{$` prose (say, documenting this very tag syntax) stops parsing after upgrade — with no way to write the old verbatim output at all (`push_escaped` only escapes the internal marker). Worth adding an escape before the syntax is frozen; retrofitting one later changes the meaning of existing templates.

lib/src/help_template.rs
```diff
@@ -87,23 +88,116 @@ pub fn check(template: &str) -> Result<(), String> {
...
+                        | "underline"
+                )
+            });
+            if let Some(unknown) = unknown {
+                return Err(format!(
+                    "help_template names no style \"{unknown}\"; use heading, option, metavar, \
+                     an ANSI colour, bright-<colour>, bold, dim, italic or underline"
+                ));
+            }
+            if specification.is_empty() {
```
jdx: This `specification.is_empty()` branch is dead code here and in `derive/src/model.rs`'s copy: `"".split('+')` yields one empty fragment, so the unknown-style check above fires first and `{$}` reports the baffling ``help_template names no style ""`` instead of the purpose-built empty-tag message. argv's `check` orders the tests the other way (empty first) — so the three hand-copied validators already emit different diagnostics for identical input. Symptom of the larger issue: the style vocabulary + tag validator now exists in ~5 hand-copies (lib, derive, argv check, argv apply, Go strip) with nothing pinning them to each other.

conformance/tests/help_template.rs
```diff
@@ -87,6 +89,18 @@ fn the_page_a_compiled_parser_prints_is_the_one_the_reference_renders() {
         "{page}"
     );
 
+    let coloured = usage_argv::help::render_styled(
```
jdx, line 92: This single-line trailing tag is the only cross-renderer coverage for the new syntax (plus one `{$orange}` rejection). None of the shapes that currently break are exercised: a tag opening or closing on its own line (panics / leaks unreset ANSI in argv while lib and Go render fine), a tag wrapping a possibly-empty section, or a vocabulary-parity pin across the five copies of the style list. Given the repo's one-engine/parity discipline, a conformance case enumerating the full vocabulary against all three validators and a multi-line-tag page compared across argv/lib/Go would catch both the current bugs and future drift.

argv/src/help.rs
```diff
@@ -180,64 +183,47 @@ impl Sections {
...
-        if line.trim().is_empty() {
-            blank = !out.is_empty();
-            continue;
-        }
-        if !out.is_empty() {
-            out.push('\n');
-            if blank {
-                out.push('\n');
-            }
+    fn styled(
```
jdx: Efficiency, minor: `styled` eagerly styles all 11 section strings, but `concatenated()` (the no-template default) reads 7, and the grouped/ungrouped fields re-slice the same content as `args`/`flags` — so per-line prefix matching (with a fresh String per line) runs roughly twice over identical text, and on sections no template names, per command in `render_all_styled`. Styling lazily where a section is consumed (in `named()`/`concatenated()`) makes only sections that reach the page pay. Same neighborhood: `push_escaped` copies the whole page char-by-char though MARK (U+0002) essentially never occurs — a `contains(MARK)` fast path plus a non-allocating blankness check in `collapse_blank_runs` would keep plain `--help` allocation-light.

jdx:
Reviewed at high effort (line-by-line with runtime repros, removed-behavior/cross-file tracing across the Rust argv, usage-lib, and Go renderers, plus cleanup passes), with adversarial verification of every finding. 9 findings posted inline. The headline items are runtime, not cosmetic:

- **`--help` hangs forever** on any usage line containing a non-ASCII uppercase word (`ÜBERSICHT`, `ФЛАГИ`) — Unicode gate, ASCII end-scan, `at += 0` (reproduced).
- **`--help` panics** for validated templates that put `{$tag}` or `{/$}` on its own line: `collapse_blank_runs` drops marker-only lines, unbalancing the render stack — plain and coloured, while lib and Go render the same template fine (reproduced). The mirror case leaks an unreset SGR into the user's prompt.
- **New panic paths in public API**: lib's `substitute` and argv's style renderer now `expect`/`unreachable!` on malformed tags; `help_template` is a pub field and nothing re-validates on the render path, so programmatically built Specs can abort where v6.2 rendered verbatim (Go passes the same input through).
- **No escape for a literal `{$`** — previously-valid specs are now refused at parse time with no way to write the old output; better to add the escape before the syntax freezes.

Structural: the style vocabulary + validator now exists in ~5 hand-copies (lib, derive, argv check, argv apply, Go), which have already drifted (the `{$}` empty-tag branch is dead in two of them, and the copies emit different diagnostics for identical input), and conformance covers only a single-line trailing tag — none of the shapes above. A vocabulary-parity conformance case plus a multi-line-tag page compared across argv/lib/Go would catch both the current bugs and future drift.

The semantic palette itself and per-section coloring approach look right, and plain output stays byte-identical in the shapes the tests cover.

_This review was generated by Claude Code._

argv/src/help.rs
```diff
@@ -480,31 +484,76 @@ fn closing_delimiter(
...
+
+        if rest.starts_with(char::is_uppercase)
+            && previous.is_none_or(|c| c.is_whitespace() || matches!(c, '=' | '[' | '<'))
+        {
+            let end = rest
+                .find(|c: char| !(c.is_ascii_uppercase() || c.is_ascii_digit() || c == '_'))
+                .unwrap_or(rest.len());
+            out.push_str(&style.metavar(&rest[..end]));
+            at += end;
+            continue;
```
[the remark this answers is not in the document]
jdx: Fixed in 36ad4b60: bracketed positional metavariables are styled regardless of case, with regression coverage for `[file]` and non-ASCII text. _This reply was AI-generated._

argv/src/help.rs
```diff
@@ -556,6 +605,8 @@ fn help_structure(
             .filter_map(|arg| arg.help_heading)
             .map(str::to_string),
     );
+    let mut arg_usages: Vec<String> = args.iter().map(|arg| arg_usage(arg)).collect();
+    arg_usages.sort_unstable_by_key(|usage| core::cmp::Reverse(usage.len()));
```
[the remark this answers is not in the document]
jdx: Fixed in 36ad4b60: flattened descendant argument and flag usages are collected with the same short/long visibility filters as their rendered rows. The new fixture verifies both styles. _This reply was AI-generated._

argv/src/help.rs
```diff
@@ -480,31 +484,76 @@ fn closing_delimiter(
...
-        rest = &rest[end..];
+
+        if rest.starts_with('<') {
+            if let Some(end) = rest.find('>') {
+                let end = end + 1;
+                out.push_str(&style.metavar(&rest[..end]));
+                at += end;
+                continue;
+            }
+        }
```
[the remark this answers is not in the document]
jdx: Fixed in 36ad4b60: a required-option wrapper now leaves `<` literal, styles the enclosed option green, and independently styles any nested value metavar. _This reply was AI-generated._

argv/src/help.rs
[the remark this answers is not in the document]
jdx: Fixed in 36ad4b60: the bare-token gate and scan are consistently ASCII, and bracketed Unicode metavariables follow a separate advancing path. The reproduced input is covered. _This reply was AI-generated._

derive/src/model.rs
```diff
@@ -4314,6 +4315,77 @@ fn check_help_template(template: &str) -> Result<(), String> {
...
+            });
+            if let Some(unknown) = unknown {
+                return Err(format!(
+                    "`help_template` names no style `{unknown}`; use heading, option, metavar, \
+                     an ANSI colour, bright-<colour>, bold, dim, italic or underline"
+                ));
+            }
+            if specification.is_empty() {
+                return Err("`help_template` has an empty style tag `{$}`".to_string());
+            }
```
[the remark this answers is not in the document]
jdx: Fixed in 36ad4b60: empty specifications are checked before vocabulary lookup. The style vocabulary is now one list per runtime, pinned across Rust and Go by `corpus/help-template-styles.txt`. _This reply was AI-generated._

argv/src/help.rs
jdx: Fixed in 36ad4b60: the bare metavariable gate now uses `is_ascii_uppercase`, so it cannot enter with a zero-byte scan. Unicode bracketed metavariables are handled without looping and are covered by regression tests. _This reply was AI-generated._

argv/src/help.rs
jdx: Fixed in 36ad4b60: a bare uppercase run is styled only when its end is a valid token boundary, so `Add` remains one unstyled word while `TOOL@VERSION` is styled as a metavar. _This reply was AI-generated._

argv/src/help/template.rs
jdx: Fixed in 36ad4b60: blank-line collapsing preserves internal open and close markers even when their lines have no visible text. Opening-only, closing-only, ANSI reset, and empty-section cases are covered in unit and renderer-corpus tests. _This reply was AI-generated._

lib/src/help_template.rs
jdx: Fixed in 36ad4b60: both Rust render paths validate defensively and fall back to section-only substitution for malformed style markup. Programmatically constructed specs now render the markup literally instead of reaching `expect` or `unreachable` paths, matching Go. _This reply was AI-generated._

lib/src/help_template.rs
jdx: Fixed in 36ad4b60: doubled dollar signs escape both delimiters. `{$$heading}` renders `{$heading}`, and `{/$$}` renders `{/$}`; derive, KDL, Rust, Go, corpus, and docs cover the syntax. _This reply was AI-generated._

lib/src/help_template.rs
jdx: Fixed in 36ad4b60: the empty-tag diagnostic is reachable and consistent, argv validation and application share one style table, and a shared corpus file pins the public Rust and Go vocabularies while a derive fixture compiles every entry. _This reply was AI-generated._

conformance/tests/help_template.rs
jdx: Fixed in 36ad4b60: the renderer corpus now includes tags on their own lines, a style wrapping an empty section, and escaped delimiters. A full-vocabulary derive fixture plus the shared Rust/Go vocabulary corpus guards parity. _This reply was AI-generated._

argv/src/help.rs
jdx: Fixed in 36ad4b60: default pages are styled once after concatenation, while templates style only named sections as they are consumed. `push_escaped` now has a fast path for ordinary text and blankness is checked without allocating a visible copy. _This reply was AI-generated._

PR #1424, opened 2026-09-13
feat(help): mark the default subcommand and opt in to appending its page

### LINKED ISSUE #1423: default_subcommand: mount CLI's --help doesn't show the default command's flags
`default_subcommand` lets a root fall back to a child (`mise foo` → `mise run foo`), but the root's `--help` never shows that child's flags or args — only the command list and root-only flags. A user has to already know to run `<bin> <default> --help` to discover them.

Example:
```kdl
name "ex"
bin "ex"
default_subcommand "install"
cmd "install" help="Put a package on the system" {
    flag "-u --update" help="Update installed packages"
    arg "[package]" help="What to install"
}
cmd "query" help="Look something up"
```

`ex --help` lists `install` and `query` but never mentions `-u/--update` or `[package]`, even though `ex -u foo` dispatches to `install` today. Hoisting those flags straight into the root grammar isn't right — it'd collide with real root-only flags and make it unclear which command owns what.

Would you take:
- marking the visible default in the command list, e.g. `install  Put a package on the system (default)`
- an opt-in (`default_subcommand_help`) that appends the default command's own help page after the parent's, so `ex --help` also shows `install`'s flags/args without hoisting them into the root spec

Hidden defaults would stay unmarked/unappended.

lu-zero:
The command list now marks a visible `default_subcommand` with `(default)`. Opt in with `default_subcommand_help #true` (Rust: `#[usage(default_subcommand_help)]`) to append that child's own help page after the parent's, so a mount CLI's `--help` shows the default command's flags and args without hoisting them into the root grammar.

```kdl
default_subcommand "install"
default_subcommand_help #true
cmd "install" help="Put a package on the system" {
    flag "-u --update" help="Update installed packages"
    arg "[package]" help="What to install"
}
cmd "query" help="Look something up"
```

`ex --help` now prints:
```
...
Commands:
  install  Put a package on the system (default)
  query    Look something up
  ...

Default command: install
  Unmatched words select this command.

Put a package on the system

Usage: ex install [-u --update] [package]
...
```

Hidden defaults are neither marked nor appended. `flatten_help` already inlines every child, so the append is skipped there. Ported to the Go renderer's `ShortHelp`/`LongHelp` as well; `AllHelp` keeps walking each descendant once rather than append-then-recurse.

New spec field `default_subcommand_help` (bool, requires `default_subcommand`; lint error otherwise).

Closes #1423.

https://claude.ai/code/session_015F1RcxwHJ5HbiVhkfxTyAV

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in setting to append the visible default subcommand’s help page to root help.
  * Default subcommands are marked with `(default)` in command lists and generated documentation.
  * Preserved styling in appended help output.
  * Hidden defaults remain excluded, while flattened or recursive help avoids duplicate output.
* **Bug Fixes**
  * Corrected default-command selection so canonical names take precedence over aliases.
  * Added validation for invalid default-subcommand help configurations.
* **Documentation**
  * Documented the new setting and its help-rendering behavior.


lu-zero:
Addressed CodeRabbit's review findings:

- **`argv/src/help.rs` (appended page not styled)** — already fixed independently in an earlier commit on this branch (threading `Style` through `with_default_command_help`/`rendered_page`), before this review ran against the latest diff.
- **`argv/src/help.rs` (resolve canonical name before alias)** — valid, fixed in 59ec2e6d. `default_visible_child` now resolves by canonical name first, falls back to alias, and applies the hidden filter only to the resolved candidate, matching `go/internal/spec/spec.go`'s `DefaultSubcommand` precedence. Note: `#[derive(Subcommands)]` already rejects a name/alias collision at compile time, so this was unreachable for derive-based CLIs — the fix hardens the shared runtime helper for KDL-authored or hand-built specs, which have no such guard. Locked with a hand-built `CommandMeta` test (a derive-based repro isn't constructible, since the derive itself refuses the fixture).
- **`conformance/tests/default_subcommand_help.rs` (exercise JSON deserialization)** — not applicable: `usage::Spec` only derives `Serialize`, not `Deserialize` (it's `#[non_exhaustive]` and JSON is a one-way export format here; KDL is the actual round-trip format). The suggested `serde_json::from_str::<usage::Spec>(...)` doesn't compile.
- **`lib/src/docs/cli/mod.rs` (avoid trimming before checking whether to append)** — not applicable: the `page.trim().to_string() + "\n"` step predates this PR (unconditional on `origin/main` before this change); the PR only threaded it through the new `append_default_command_help` call, which already returns the same trimmed page unchanged whenever nothing is appended. No behavior change there to fix.

Also fixed the CI render-check failure (rustfmt + regenerated `examples/docs/MISE_INLINE.md`/`MISE_MULTI.md`).

